PR #2, opened 2025-10-25
add picolet to delete sandbox

LiZhenCheng9527:


cmd/picolet/main.go
```diff
@@ -0,0 +1,50 @@
...
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/agent-box/pico-apiserver/pkg/picolet"
+)
+
+func main() {
+	var (
+		port       = flag.String("port", "9090", "Picolet server port")
```
hzxuzhonghu: why do we need a http server?

cmd/picolet/main.go
```diff
@@ -0,0 +1,50 @@
...
+	log.Printf("Picolet started on port %s", *port)
+
+	select {
+	case <-sigCh:
+		log.Println("Received shutdown signal")
+	case err := <-errCh:
+		log.Printf("Picolet error: %v", err)
+	}
+
+	cancel()
```
hzxuzhonghu: ?

pkg/picolet/picolet.go
```diff
@@ -0,0 +1,150 @@
...
+func getClientConfig(kubeconfig string) (*rest.Config, error) {
+	if kubeconfig == "" {
+		return rest.InClusterConfig()
+	}
+
+	return clientcmd.BuildConfigFromFlags("", kubeconfig)
+}
+
+func (p *Picolet) setupRoutes() {
+	p.router.HandleFunc("/delete", p.handleSandboxDelete).Methods("POST")
```
hzxuzhonghu: We have no components to call you to pause

hzxuzhonghu: or delete

pkg/picolet/picolet.go
```diff
@@ -0,0 +1,150 @@
...
+		return false, fmt.Sprintf("Failed to get sandbox: %v", err)
+	}
+
+	if err := p.sandboxClient.Delete(context.TODO(), sandbox); err != nil {
+		return false, fmt.Sprintf("Failed to delete sandbox: %v", err)
+	}
+
+	log.Printf("Pausing sandbox %s/%s", namespace, name)
+
+	return true, "Sandbox paused successfully"
```
hzxuzhonghu: it is in handleSandboxDelete

pkg/picolet/picolet.go
```diff
@@ -0,0 +1,73 @@
...
+		return ctrl.Result{}, err
+	}
+
+	if sandboxIsRunning(sandbox) {
+		// The timer is only set during the first run.
+		if _, exists := r.Cache[req.NamespacedName]; !exists {
+			fmt.Printf("New running sandbox detected %v, setting timer", req.NamespacedName.String())
+			// shutdownTime is an absolute time when the sandbox is deleted
+			shutDownTime := sandbox.Spec.ShutdownTime
+			r.Cache[req.NamespacedName] = sandbox.Spec.ShutdownTime
```
hzxuzhonghu: This is not what we want. `ShutdownTime` is the max ttl a sandbox created with.

But what we are handling here is the session timeout.

pkg/picolet/picolet.go
```diff
@@ -0,0 +1,82 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	sandboxv1alpha1 "sigs.k8s.io/agent-sandbox/api/v1alpha1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	picoapiserver "github.com/agent-box/pico-apiserver/pkg/pico-apiserver"
+)
+
+var (
+	SandboxExpirationInterval = 15 * time.Minute
```
hzxuzhonghu: SessionExpirationTimeout

pkg/picolet/picolet.go
```diff
@@ -0,0 +1,82 @@
...
+				if err := r.Delete(ctx, sandbox); err != nil {
+					if !errors.IsNotFound(err) {
+						return ctrl.Result{}, err
+					}
+				}
+			} else {
+				return ctrl.Result{RequeueAfter: expirationTime.Sub(time.Now())}, nil
+			}
+		} else {
+			sandbox.Annotations[picoapiserver.LastActivityAnnotationKey] = time.Now().Format(time.RFC3339)
```
hzxuzhonghu: i could not understand this branch

LiZhenCheng9527: If there is no Last Activity annotation, then the current time will be set as the Last Activity Time for this sandbox.
I also have questions about how to handle this branch.
Should I add an annotation or just delete it outright?

hzxuzhonghu: why no Last Activity?

LiZhenCheng9527: This branch only executes when `LastActivityAnnotationKey` is not found or its value is empty. 
Where can I obtain the Last Activity Time?

hzxuzhonghu: why canot you get, apiserver always set one first on creation, then also on invoke

pkg/picolet/picolet.go
```diff
@@ -43,7 +43,7 @@ func (r *PicoletReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 				return ctrl.Result{RequeueAfter: 30 * time.Second}, err
 			}
 
-			expirationTime := lastActivity.Add(SandboxExpirationInterval)
+			expirationTime := lastActivity.Add(SessionExpirationTimeout)
```
hzxuzhonghu: one other comment: this delete logic should not be under `if sandboxIsRunning(sandbox) {`

LiZhenCheng9527: Currently, there is no “Paused” state for the sandbox, so it is temporarily placed under “Running.”
 Once the “Paused” state is added for the `agent-sandbox`, the handling here will be updated accordingly.

hzxuzhonghu: NOt only running or paused

pkg/pico-apiserver/k8s_client.go
```diff
@@ -276,6 +285,37 @@ func (c *K8sClient) WaitForSandboxReady(ctx context.Context, sandboxName string,
...
+		},
+	}
+
+	// Convert to JSON
+	patchBytes, err := json.Marshal(patchData)
+	if err != nil {
+		return fmt.Errorf("failed to marshal patch data: %w", err)
+	}
+
+	// Apply the patch using strategic merge patch
```
hzxuzhonghu: Not strategic merge, it is just normal json merge patch

pkg/pico-apiserver/tunnel.go
```diff
@@ -90,6 +90,11 @@ func (s *Server) handleTunnel(w http.ResponseWriter, r *http.Request) {
 	// Client -> Backend
 	go func() {
 		defer wg.Done()
+		// update sandbox last activity timestamp
+		err := s.k8sClient.UpdateSandboxLastActivityWithPatch(r.Context(), session.SandboxName, session.LastActivityAt)
```
hzxuzhonghu, line 95: nit: we need to improve here in case the cmd takes too long

Issue #3, opened 2025-10-28
Apiserver path should respect the spec


hzxuzhonghu:
https://github.com/volcano-sh/agentcube/blob/main/api-spec/sandbox-api-spec.yaml

cc @YaoZengzeng

PR #9, opened 2025-10-29
apiserver path should respect the spec

YaoZengzeng:
fixes: #3

### DIFF
--- a/pkg/pico-apiserver/tunnel.go
+++ b/pkg/pico-apiserver/tunnel.go
@@ -94,7 +94,7 @@ func (s *Server) handleTunnel(w http.ResponseWriter, r *http.Request) {
 		// TODO: improve here to reduce the processing latency of the cmd command
 		err := s.k8sClient.UpdateSandboxLastActivityWithPatch(r.Context(), sandbox.SandboxName, sandbox.LastActivityAt)
 		if err != nil {
-			log.Printf("Failed to updat last activity time for sandbox %s, err %v", sandbox.SandboxName, err)
+			log.Printf("Failed to update last activity time for sandbox %s, err %v", sandbox.SandboxName, err)
 		}
 		s.proxyDataOneWay(backendConn, clientConn, sandboxID, "client->backend")
 	}()


pkg/pico-apiserver/tunnel.go
```diff
@@ -19,12 +19,12 @@ import (
...
-	sessionID := vars["sessionId"]
+	sandboxID := vars["sandboxId"]
 
-	// Check if session exists
-	session := s.sessionStore.Get(sessionID)
-	if session == nil {
-		http.Error(w, "Session not found", http.StatusNotFound)
+	// Check if sandbox exists
+	sandbox := s.sandboxStore.Get(sandboxID)
+	if sandbox == nil {
```
hzxuzhonghu, line 26: some more handling should be checking the sandbox status, if it is paused, should do the resume

pkg/pico-apiserver/apiserver.go
```diff
@@ -53,21 +53,18 @@ func NewServer(config *Config, sandboxController *controller.SandboxReconciler)
 func (s *Server) setupRoutes() {
 	s.router = mux.NewRouter()
 
-	// API v1 routes
-	v1 := s.router.PathPrefix("/v1").Subrouter()
```
hzxuzhonghu, line 57: would prefer we keep /v1 prefix, in case we donot break comapatability we can introduce v2 later

pkg/pico-apiserver/tunnel.go
```diff
@@ -92,17 +92,17 @@ func (s *Server) handleTunnel(w http.ResponseWriter, r *http.Request) {
 		defer wg.Done()
 		// update sandbox last activity timestamp
 		// TODO: improve here to reduce the processing latency of the cmd command
-		err := s.k8sClient.UpdateSandboxLastActivityWithPatch(r.Context(), session.SandboxName, session.LastActivityAt)
+		err := s.k8sClient.UpdateSandboxLastActivityWithPatch(r.Context(), sandbox.SandboxName, sandbox.LastActivityAt)
 		if err != nil {
-			log.Printf("Failed to updat last activity time for sandbox %s, err %v", session.SandboxName, err)
+			log.Printf("Failed to updat last activity time for sandbox %s, err %v", sandbox.SandboxName, err)
```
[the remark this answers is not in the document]
YaoZengzeng: good catch, fixed.

pkg/pico-apiserver/apiserver.go
```diff
@@ -53,23 +53,25 @@ func NewServer(config *Config, sandboxController *controller.SandboxReconciler)
...
 	// Health check (no authentication required)
 	s.router.HandleFunc("/health", s.handleHealth).Methods("GET")
 
-	// Session management endpoints
-	v1.HandleFunc("/sessions", s.authMiddleware(s.handleCreateSession)).Methods("POST")
-	v1.HandleFunc("/sessions", s.authMiddleware(s.handleListSessions)).Methods("GET")
-	v1.HandleFunc("/sessions/{sessionId}", s.authMiddleware(s.handleGetSession)).Methods("GET")
-	v1.HandleFunc("/sessions/{sessionId}", s.authMiddleware(s.handleDeleteSession)).Methods("DELETE")
+	// API v1 routes
+	v1 := s.router.PathPrefix("/v1").Subrouter()
```
YaoZengzeng, line 60: @hzxuzhonghu move here

PR #12, opened 2025-11-03
add run code function

Tweakzx:


sdk-python/agentcube/clients/ssh_client.py
```diff
@@ -90,8 +91,59 @@ def run_code(
...
+            raise ValueError(f"Unsupported language: {language}. Supported: {list(lang_aliases.keys())}")
+
+        if target_lang == "python":
+            quoted_code = shlex.quote(code)
+            command = f"python3 -c {quoted_code}"
+        elif target_lang == "bash":
+            quoted_code = shlex.quote(code)
+            command = f"bash -c {quoted_code}"
+
+        ssh_client = self._ssh_client or self.connect_ssh()
```
hzxuzhonghu: should you set self._ssh_client = ssh_client

sdk-python/agentcube/clients/ssh_client.py
```diff
@@ -90,8 +91,59 @@ def run_code(
...
+        if not ssh_client:
+            raise ConnectionError("No SSH connection established. Please call connect_ssh first.")
+
+        try:
+            stdin, stdout, stderr = ssh_client.exec_command(command, timeout=30)
```
hzxuzhonghu: should add timeout as a param of run function

Issue #13, opened 2025-11-04
apiserver should recover from restart


YaoZengzeng:
Now apiserver store the relation of sandbox id and its corresponding object in memory.

These information will lose once it restart.

PR #15, opened 2025-11-06
apiserver should be able to recover from restart and work in HA mode

YaoZengzeng:
fixes: #13

### DIFF
--- a/pkg/apiserver/store.go
+++ b/pkg/apiserver/store.go
@@ -88,22 +88,6 @@ func (s *SandboxStore) List() []*Sandbox {
 	return sandboxes
 }
 
-// CleanExpired cleans up expired sandboxes
-func (s *SandboxStore) CleanExpired() int {
-	s.mu.Lock()
-	defer s.mu.Unlock()
-
-	now := time.Now()
-	count := 0
-	for sandboxID, sandbox := range s.sandboxes {
-		if now.After(sandbox.ExpiresAt) {
-			delete(s.sandboxes, sandboxID)
-			count++
-		}
-	}
-	return count
-}
-
 // InitializeWithInformer initializes the store with an informer and performs initial sync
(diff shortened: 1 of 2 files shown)


pkg/apiserver/store.go
```diff
@@ -0,0 +1,331 @@
...
+}
+
+// InitializeWithInformer initializes the store with an informer and performs initial sync
+func (s *SandboxStore) InitializeWithInformer(ctx context.Context, informer cache.SharedInformer, k8sClient *K8sClient, namespace string) error {
+	s.mu.Lock()
+	s.informer = informer
+	s.mu.Unlock()
+
+	// Set up event handlers
+	informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
hzxuzhonghu, line 98: there is already a informer in the controller

pkg/apiserver/store.go
```diff
@@ -0,0 +1,331 @@
...
+	for _, sandbox := range s.sandboxes {
+		if now.Before(sandbox.ExpiresAt) {
+			sandboxes = append(sandboxes, sandbox)
+		}
+	}
+	return sandboxes
+}
+
+// CleanExpired cleans up expired sandboxes
+func (s *SandboxStore) CleanExpired() int {
```
hzxuzhonghu: seems not called

YaoZengzeng: good catch, fixed.

pkg/apiserver/store.go
YaoZengzeng: Yes, now this informer is derived from `k8sClient` from apiserver.

But we also have a  k8s client in controller.

I'll raise an other PR to fix this.

YaoZengzeng:
@hzxuzhonghu ptal

YaoZengzeng:
@hzxuzhonghu not get merged automatically.

PR #27, opened 2025-11-17
[feature] add authorization header

Tweakzx:


sdk-python/agentcube/clients/client.py
```diff
@@ -14,14 +14,23 @@ class SandboxClient:
...
+    ):
         """Initialize the client
         
         Args:
             api_url: Pico API server address. Defaults to environment variable API_URL 
                      or DEFAULT_API_URL if not provided
         """
-        self.api_url = api_url or get_env("API_URL", constants.DEFAULT_API_URL)
+        self.api_url = api_url or get_env(constants.API_URL_ENV, constants.DEFAULT_API_URL)
+        self.auth_token = auth_token or get_env("API_TOKEN", "")
```
hzxuzhonghu: I am suggesting we read `/var/run/secrets/kubernetes.io/serviceaccount/token` by default

sdk-python/agentcube/clients/client.py
```diff
@@ -141,8 +150,8 @@ def establish_tunnel(self, sandbox_id: str, auth_token: str = "") -> socket.sock
         req = f"CONNECT /v1/sandboxes/{sandbox_id} HTTP/1.1\r\n"
         req += f"Host: {hostname}:{port}\r\n"
         req += "User-Agent: agentcube-sdk-python/1.0\r\n"
-        if auth_token:
-            req += f"Proxy-Authorization: Bearer {auth_token}\r\n"
+        if self.auth_token:
+            req += f"Authorization: Bearer {self.auth_token}\r\n"
```
hzxuzhonghu, line 154: donot change this, for connect, it use Proxy-Authorization standard

Tweakzx: if use _Proxy-Authorization_, the request will be rejected by agentcube-apiserver.

Tweakzx: @YaoZengzeng should apiserver do some changes according to this suggestion?

PR #35, opened 2025-11-19
basic github workflow

YaoZengzeng:


.github/workflows/main.yml
```diff
@@ -0,0 +1,27 @@
+name: Agentcube CI Workflow
```
YaoZengzeng, line 1: @hzxuzhonghu CI is not triggered, don't know if it is the authority issue.

hzxuzhonghu (approved):
/lgtm

PR #48, opened 2025-12-01
add redis client and interface

LeslieKuo:


pkg/redis/client.go
```diff
@@ -0,0 +1,397 @@
...
+	//   lastActivityIndexKey:  score = LastActivityAt.Unix(),  member = sandboxID
+	expiryIndexKey       string
+	lastActivityIndexKey string
+}
+
+// Option configures client behavior (e.g. key prefixes).
+type Option func(*client)
+
+// WithKeyPrefixes sets custom key prefixes for session / sandbox / lock.
+func WithKeyPrefixes(session, sandbox, lock string) Option {
```
hzxuzhonghu: It is very easy for use to apply this when calling `NewClient`. Then duplicate key would occur. 

`sandbox:expiryexpiry` `sandbox:last_activitylast_activity`

pkg/redis/client.go
```diff
@@ -0,0 +1,397 @@
...
+
+	if !sb.ExpiresAt.IsZero() {
+		pipe.ZAdd(ctx, c.expiryIndexKey, redisv9.Z{
+			Score:  float64(sb.ExpiresAt.Unix()),
+			Member: sb.SandboxID,
+		})
+	}
+
+	if !sb.LastActivityAt.IsZero() {
+		pipe.ZAdd(ctx, c.lastActivityIndexKey, redisv9.Z{
```
hzxuzhonghu: we do need a separate method to only update last activity time

pkg/redis/client.go
```diff
@@ -0,0 +1,397 @@
...
+
+	if _, err := pipe.Exec(ctx); err != nil {
+		return fmt.Errorf("BindSessionWithSandbox: redis TxPipeline EXEC: %w", err)
+	}
+	return nil
+}
+
+// DeleteSessionBySandboxIDTx deletes the bidirectional mapping by sandboxID and
+// removes the related index entries, using WATCH + TxPipelined for atomicity.
+func (c *client) DeleteSessionBySandboxIDTx(ctx context.Context, sandboxID string) error {
```
hzxuzhonghu, line 157: Prefer naming `DeleteSession` or `DeleteSandbox`, which delete all the resources ralated to a session/sandbox.

pkg/redis/client.go
```diff
@@ -0,0 +1,397 @@
...
+	sandboxKey := c.sandboxKey(sandboxID)
+
+	for {
+		err := c.rdb.Watch(ctx, func(tx *redisv9.Tx) error {
```
hzxuzhonghu: I cannot get it why we need `Watch` to monitor if the sandboxKey is changed by others before exec cleanup

pkg/common/types/sandbox.go
```diff
@@ -0,0 +1,59 @@
+package types
+
+import (
+	"fmt"
+	"time"
+)
+
+type SandboxRedis struct {
+	SandboxID      string          `json:"sandboxId"`
+	SandboxName    string          `json:"sandboxName"`
```
hzxuzhonghu: name is already in ID?

pkg/common/types/sandbox.go
```diff
@@ -0,0 +1,59 @@
...
+	Accesses       []SandboxAccess `json:"accesses"`
+	SessionID      string          `json:"sessionId"`
+	CreatedAt      time.Time       `json:"createdAt"`
+	ExpiresAt      time.Time       `json:"expiresAt"`
+	LastActivityAt time.Time       `json:"lastActivityAt,omitempty"`
```
hzxuzhonghu: IIRC, this is to value of sessionID->sandbox. And the time is already set in sorted set. 

These fields seems redundant.

pkg/common/types/sandbox.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"fmt"
+	"time"
+)
+
+type SandboxRedis struct {
+	SandboxID      string          `json:"sandboxId"`
+	SandboxName    string          `json:"sandboxName"`
+	Accesses       []SandboxAccess `json:"accesses"`
```
hzxuzhonghu: EntryPoints?

pkg/redis/client.go
```diff
@@ -0,0 +1,397 @@
...
+		}
+		result = append(result, &sb)
+	}
+
+	return result, nil
+}
+
+// UpdateSandboxLastActivity updates LastActivityAt for the given sandbox and
+// synchronizes the last-activity index. TTL on the session key is preserved.
+func (c *client) UpdateSandboxLastActivity(ctx context.Context, sandboxID string, at time.Time) error {
```
hzxuzhonghu, line 293: I'd suggest only update `lastActivityIndexKey`

pkg/redis/client.go
LeslieKuo: Already added UpdateSandboxLastActivity(ctx context.Context, sandboxID string, at time.Time) method on the Redis client.

pkg/common/types/sandbox.go
LeslieKuo: Done

pkg/redis/client.go
LeslieKuo: Done. Updated UpdateSandboxLastActivity to only mutate the sandbox:last_activity sorted-set index (session value and TTL are no longer touched), and removed LastActivityAt from SandboxRedis so the index is the single source of truth.

pkg/common/types/sandbox.go
LeslieKuo: Done

pkg/redis/client.go
LeslieKuo: I perfer keep it in DeleteSessionBySandboxIDTx. Cause it actually indicate delete session or sandbox, only declare by sandboxId

pkg/redis/client.go
LeslieKuo: Good point. The WATCH here was trying to be overly careful, but for this cleanup operation we don't really need optimistic locking semantics.
I simplified DeleteSessionBySandboxIDTx to:

Do a single GET on sandbox:{sandboxID} to find the current session (if any),

Treat missing mappings as success, and

Use a simple pipeline to DEL the mapping and ZREM from the indexes.

This makes the function best-effort and much easier to reason about, without relying on WATCH.

pkg/redis/client.go
LeslieKuo: Done

pkg/common/types/sandbox.go
LeslieKuo: Yanbin will fix it

PR #52, opened 2025-12-03
[WIP]Add init process for picod in workload manager

VanderChen:


pkg/workloadmanager/controller.go
```diff
@@ -0,0 +1,82 @@
+package workloadmanager
```
tjucoder: this file was renamed to sandbox_controller.go

PR #55, opened 2025-12-03
agent-runtime and code-interpreter delete implement

tjucoder:


k8s/agentcube-apiserver.yaml
```diff
@@ -89,6 +98,11 @@ spec:
             - name: http
               containerPort: 8080
               protocol: TCP
+          env:
+            - name: REDIS_ADDR
+              value: "127.0.0.1:6379"
+            - name: REDIS_PASSWORD
+              value: "********"
```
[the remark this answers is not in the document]
tjucoder: it's just for testing, fix by empty string

pkg/redis/client.go
```diff
@@ -156,13 +160,75 @@ func (c *client) BindSessionWithSandbox(ctx context.Context, sessionID string, s
...
+	if len(cmder) == 0 {
+		return errors.New("StoreSandbox: unexpected empty cmder")
+	}
+
+	if cmder[0].Err() != nil {
+		return fmt.Errorf("StoreSandbox: SET failed: %w", cmder[0].Err())
+	}
+
 	return nil
 }
```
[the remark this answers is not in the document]
tjucoder: already fix it

k8s/agentcube-apiserver.yaml
```diff
@@ -41,6 +41,15 @@ rules:
   - apiGroups: ["runtime.agentcube.volcano.sh"]
     resources: ["codeinterpreters/finalizers"]
     verbs: ["update"]
+  - apiGroups: ["runtime.agentcube.volcano.sh"]
+    resources: ["agentruntimes"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
```
hzxuzhonghu: I donot think we need write permission

k8s/agentcube-apiserver.yaml
```diff
@@ -41,6 +41,15 @@ rules:
...
+  - apiGroups: ["runtime.agentcube.volcano.sh"]
+    resources: ["agentruntimes/status"]
+    verbs: ["get", "update", "patch"]
```
hzxuzhonghu: remove get?

k8s/agentcube-apiserver.yaml
```diff
@@ -41,6 +41,15 @@ rules:
...
+  - apiGroups: ["runtime.agentcube.volcano.sh"]
+    resources: ["agentruntimes/finalizers"]
```
hzxuzhonghu: Do you rely on it later?

pkg/redis/client.go
```diff
@@ -156,13 +162,79 @@ func (c *client) BindSessionWithSandbox(ctx context.Context, sessionID string, s
...
 
+func (c *client) UpdateSandbox(ctx context.Context, sandboxRedis *types.SandboxRedis, ttl time.Duration) error {
```
hzxuzhonghu, line 165: can we remove `BindSessionWithSandbox` now

pkg/redis/client.go
```diff
@@ -156,13 +162,79 @@ func (c *client) BindSessionWithSandbox(ctx context.Context, sessionID string, s
...
+		return fmt.Errorf("UpdateSandbox: marshal sandbox: %w", err)
+	}
+
+	ok, err := c.rdb.SetXX(ctx, sessionKey, b, ttl).Result()
+	if err != nil {
+		return fmt.Errorf("UpdateSandbox: redis SETXX %s: %w", sessionKey, err)
+	}
+
+	if ok == false {
+		return fmt.Errorf("UpdateSandbox: redis SETXX %s, key not exists", sessionKey)
```
hzxuzhonghu, line 183: seems not accurate, no key exist does not cause error

pkg/redis/client.go
```diff
@@ -156,13 +162,79 @@ func (c *client) BindSessionWithSandbox(ctx context.Context, sessionID string, s
...
+	cmder, err := pipe.Exec(ctx)
+	if err != nil {
+		return fmt.Errorf("StoreSandbox: redis TxPipeline EXEC: %w", err)
+	}
+
+	if len(cmder) == 0 {
+		return errors.New("StoreSandbox: unexpected empty cmder")
+	}
+
+	if cmder[0].Err() != nil {
```
hzxuzhonghu: Add ` for _, cmd := range cmds {` check

pkg/redis/client.go
```diff
@@ -195,6 +267,20 @@ func (c *client) DeleteSessionBySandboxIDTx(ctx context.Context, sandboxID strin
 	return nil
 }
 
+func (c *client) DeleteSandboxBySessionIDTx(ctx context.Context, sessionID string) error {
```
hzxuzhonghu, line 272: what does Tx mean? This delete doesnot need to be transaction

pkg/workloadmanager/workload_builder.go
```diff
@@ -136,7 +145,8 @@ func buildSandboxByAgentRuntime(namespace string, name string, ifm *Informers) (
 	}
 	sandbox := buildSandboxObject(buildParams)
 	externalInfo := &sandboxExternalInfo{
-		Ports: agentRuntimeObj.Spec.Ports,
+		Ports:     agentRuntimeObj.Spec.Ports,
+		SessionID: sessionId,
```
hzxuzhonghu, line 149: you can get the sessionId from sandbox label outside

pkg/workloadmanager/handlers.go
```diff
@@ -68,21 +86,34 @@ func (s *Server) handleCreateSandbox(c *gin.Context) {
...
-		respondError(c, http.StatusInternalServerError, "CLIENT_CREATION_FAILED", clientErr.Error())
+	userClient, err := s.k8sClient.GetOrCreateUserK8sClient(userToken, userNamespace, serviceAccountName)
+	if err != nil {
+		log.Printf("create user client failed: %v", err)
+		respondError(c, http.StatusInternalServerError, "CLIENT_CREATION_FAILED", err.Error())
 		return
 	}
+
+	// Store placeholder before creating, make sandbox/sandboxClaim GarbageCollection possible
+	sandboxRedisPlaceHolder := buildSandboxRedisCachePlaceHolder(sandbox, externalInfo)
```
hzxuzhonghu, line 97: I prefer reuse existing logic rather than generating another function, just keep the above comments

BTW, do you miss sandboxClaim?

k8s/agentcube-apiserver.yaml
tjucoder: no, remote it now

k8s/agentcube-apiserver.yaml
tjucoder: remove it

k8s/agentcube-apiserver.yaml
tjucoder: yes, remove  "create", "update", "patch", "delete"

pkg/redis/client.go
tjucoder: yes, I will remove all useless Redis funcion later

pkg/redis/client.go
tjucoder: > seems not accurate, no key exist does not cause error

it should case error, SETXX not exists is unacceptabe in my logic

pkg/redis/client.go
tjucoder: Tx means transaction, delete sandbox must ZRem at same time, so i use Tx

pkg/workloadmanager/workload_builder.go
tjucoder: > you can get the sessionId from sandbox label outside

Yes, but it might be better not to rely on labels?

pkg/workloadmanager/handlers.go
tjucoder: sandboxRedisPlaceHolder has SandboxClaimName field, so not miss claim
existing logic use sandbox.UID, GetCreationTimestamp().. so i use another simple function

pkg/redis/client.go
tjucoder: > Add ` for _, cmd := range cmds {` check

OK, check all cmds
```go
	for i, cmd := range cmder {
		if err = cmd.Err(); err != nil {
			return fmt.Errorf("StoreSandbox: SET failed: %w, cmder index: %v", err, i)
		}
	}
```

k8s/agentcube-apiserver.yaml
```diff
@@ -89,6 +95,11 @@ spec:
             - name: http
               containerPort: 8080
               protocol: TCP
+          env:
+            - name: REDIS_ADDR
+              value: "127.0.0.1:6379"
```
hzxuzhonghu, line 100: Nit: i guess this is a placeholder, please update using a var when we support helm

pkg/redis/client.go
hzxuzhonghu: I mean if not exist, it will return an error, handled in L179

pkg/redis/client.go
hzxuzhonghu: Then you should use c.rdb.`Pipeline()`

hzxuzhonghu (commented):
ONe critical problem: now you set sessionID->sandbox with ttl. The gc will periodically scan those sessions to delete their sandboxes, but there is a big chance you cannot get the sandbox info now.

pkg/redis/client.go
tjucoder: > Then you should use c.rdb.`Pipeline()`

Fix DeleteBySessionIDTx by using TxPipeline
DeleteBySandboxIDTx already deleted in Picod PRs, ignore it

tjucoder:
> ONe critical problem: now you set sessionID->sandbox with ttl. The gc will periodically scan those sessions to delete their sandboxes, but there is a big chance you cannot get the sandbox info now.

Fix it by NoExpired

k8s/agentcube-apiserver.yaml
tjucoder: > Nit: i guess this is a placeholder, please update using a var when we support helm

yes, it's just a place holder, we should update later

pkg/redis/client.go
tjucoder: > I mean if not exist, it will return an error, handled in L179

SetXX not existing not cause any error, just returns false of ok

Issue #64, opened 2025-12-08
Invoke request fails with 'Connection refused' for complex agent


VanderChen:
When an agent is built using frameworks like LangChain, its startup time can be relatively slow due to initialization overhead (e.g., loading models, setting up tools, etc.). Currently, the system assumes the agent is ready to receive requests as soon as its pod enters the Running state. However, the agent process may not have fully started listening on its port by that time.

As a result, the router immediately forwards incoming invoke requests to the pod, but the agent service isn’t ready yet—leading to "Connection refused" errors.

router log
```
2025/12/08 12:05:03 AgentRuntime invoke request: namespace=default, name=math-agent, path=/
2025/12/08 12:05:05 The selected entrypoint for session-id ce069388-7be2-44a0-9fb8-e651f400700b to sandbox is http://192.168.25.129:8080
2025/12/08 12:05:05 Start to proxy for session-id db2ee3c3-ff12-44f4-9077-9a1b4961d567 for AgentRuntime
2025/12/08 12:05:05 Forwarding request to: http://192.168.25.129:8080/
2025/12/08 12:05:05 Proxy error: dial tcp 192.168.25.129:8080: connect: connection refused
```
agent log
```log
Starting Math Agent on port 8080 at 2025-12-08 12:05:06
Health check: http://localhost:8080/health
Agent endpoint: http://localhost:8080/
Math Agent is running on port 8080
```
Timing analysis shows that the router starts forwarding requests approximately 1 second before the agent begins listening

**Proposed Fix**:
Add a proper readiness probe (e.g., HTTP or TCP) that verifies the agent endpoint is responsive before marking the pod as ready. Alternatively, implement a startup delay or health check mechanism aligned with the agent's actual readiness.

VanderChen:
Proposed Fix Sequence
```mermaid
sequenceDiagram
    participant Client
    participant Router
    participant WorkloadMgr as Workload Manager
    participant K8s as Kubernetes API
    participant App as AgentRuntime Pod

    Client->>Router: Create Sandbox Request
    Router->>WorkloadMgr: Forward Create Request
    WorkloadMgr->>K8s: Create Sandbox CR

    
    loop Readiness Probe
        K8s->>App: HTTP GET /health/ready
        App-->>K8s: 200 OK (when ready)
    end
    
    K8s->>WorkloadMgr: Pod Ready Event
    WorkloadMgr->>WorkloadMgr: Verify Ready=true
[...]

PR #71, opened 2025-12-11
Fix issue 64

LeslieKuo:
add ReadinessProbe to fix issue https://github.com/volcano-sh/agentcube/issues/64

### DIFF
--- a/pkg/workloadmanager/workload_builder.go
+++ b/pkg/workloadmanager/workload_builder.go
@@ -7,10 +7,12 @@ import (
 
 	"github.com/google/uuid"
 	runtimev1alpha1 "github.com/volcano-sh/agentcube/pkg/apis/runtime/v1alpha1"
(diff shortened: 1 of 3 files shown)


pkg/workloadmanager/workload_builder.go
```diff
@@ -79,6 +82,68 @@ func buildSandboxObject(params *buildSandboxParams) *sandboxv1alpha1.Sandbox {
 	}
 	sandbox.Spec.PodTemplate.ObjectMeta.Labels[SessionIdLabelKey] = params.sessionID
 	sandbox.Spec.PodTemplate.ObjectMeta.Labels["sandbox-name"] = params.sandboxName
+
+	// Handle Readiness Probe
```
MahaoAlex, line 86: Is the readiness probe only being added to container[0]? If this follows the design logic and is a deliberate design limit, please add code comments to mark this point clearly.

PR #68, opened 2025-12-09
Workloadmanager Improvements

acsoto:
This pull request primarily renames the project and all related components from `agentcube-apiserver` to `workloadmanager`, ensuring consistent naming across the codebase, documentation, Docker, Kubernetes manifests, and build scripts. Additionally, it introduces new configuration options for sandbox image pulling and updates dependencies.

**Project-wide renaming and configuration updates:**

*Major rename of binaries, scripts, and manifests:*
[...]

PR #75, opened 2025-12-15
Add Cache adapter layer impl

tjucoder:
1) add cache layer impl, support redis and valkey
2) router pkg using cache layer instead of redis function directly
3) we do not remove redis pkg in case of conflect, after merge this PR, I will remove redis pkg and using cahce layer for workload-manager in PR https://github.com/volcano-sh/agentcube/pull/68

So, plz merge this PR before https://github.com/volcano-sh/agentcube/pull/68

@hzxuzhonghu

### DIFF
--- a/pkg/common/types/sandbox.go
+++ b/pkg/common/types/sandbox.go
@@ -10,6 +10,7 @@ type SandboxRedis struct {
 	SandboxNamespace string               `json:"sandboxNamespace"`
 	SandboxName      string               `json:"sandboxName"`
 	SandboxClaimName string               `json:"sandboxClaimName"`
+	Name             string               `json:"name"`
 	EntryPoints      []SandboxEntryPoints `json:"entryPoints"`
 	SessionID        string               `json:"sessionId"`
 	CreatedAt        time.Time            `json:"createdAt"`
--- a/pkg/store/singleton.go
+++ b/pkg/store/singleton.go
@@ -14,8 +14,8 @@ const (
 )
 
 var (
-	initStoreOnce       = &sync.Once{}
-	provider      Store = nil
+	initStoreOnce = &sync.Once{}
+	provider      Store
 )
 
 // Storage get store singleton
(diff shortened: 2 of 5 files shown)


pkg/common/types/sandbox.go
```diff
@@ -10,6 +10,7 @@ type SandboxRedis struct {
 	SandboxNamespace string               `json:"sandboxNamespace"`
 	SandboxName      string               `json:"sandboxName"`
 	SandboxClaimName string               `json:"sandboxClaimName"`
+	Name             string               `json:"name"`
```
hzxuzhonghu, line 13: no need to mutate this, it will conflict with #68

pkg/store/singleton.go
```diff
@@ -0,0 +1,69 @@
...
+// support Redis, Valkey, Redis as default, can be setting by env STORE_TYPE
+// --- redis STORE_TYPE environments ---
+// REDIS_ADDR:     redis address, required
+// REDIS_PASSWORD: redis password, required
+// --- valkey STORE_TYPE environments ---
+// VALKEY_ADDR:          valkey address, required
+// VALKEY_PASSWORD:      valkey password, required
+// VALKEY_DISABLE_CACHE: disable valkey client cache, optional
+// VALKEY_FORCE_SINGLE:  force setting valkey single mode, optional
+func Storage() Store {
```
hzxuzhonghu, line 31:
```suggestion
func New() Store {
```

pkg/common/types/sandbox.go
tjucoder: > no need to mutate this, it will conflict with #68

I have to mutate this to make complile and testing OK
After this PR merged, i will  resolve https://github.com/volcano-sh/agentcube/pull/68

pkg/store/singleton.go
tjucoder: > 

It's not new, is singleton instance

hzxuzhonghu:
please make the ci happy

tjucoder:
> please make the ci happy

Done

pkg/store/singleton.go
```diff
@@ -0,0 +1,69 @@
...
+	initStoreOnce.Do(func() {
+		err := initStore()
+		if err != nil {
+			log.Fatalf("init redis store failed: %v", err)
+		}
+	})
+	return provider
+}
+
+func initStore() error {
```
MahaoAlex, line 41: Is it necessary to send a Ping request after initialization to verify that the store address is available and reachable?

Issue #74, opened 2025-12-15
supersed log with klog


hzxuzhonghu:
raw log lib does not provide much capabilities like logging level control, klog can be used instead.

hzxuzhonghu:
@FAUST-BENCHOU Thank you for take over this, another suggestion is to lower the log level in many places. There are too many spammy logs now. 


```
Default (V=0): This level logs only essential user-facing information and critical errors. It is minimal by default but can be silenced with a "quiet" flag in some CLIs.
Suggested (V=2): This is generally recommended as a default for most systems (like the kubelet) because it provides enough detail to help when reading logs and debugging common issues without being excessively verbose. It can decrease the time needed for support and troubleshooting.
Troubleshooting (Higher V-levels): For highly detailed, granular debugging of specific, complex issues, you may temporarily increase the level further (e.g., -v=4 or higher) to capture extremely detailed execution flow information. These higher levels are generally not suitable for continuous logging in a production environment due to the volume of logs generated. 
```

FAUST-BENCHOU:
@hzxuzhonghu I've changed all instances of klog.V(x) to the default V(0). If any other adjustments are needed, feel free to let me know.

hzxuzhonghu:
The code is changing quickly, i think first step it is ok to move forward

PR #77, opened 2025-12-17
supersed log with klog

FAUST-BENCHOU:
Closes https://github.com/volcano-sh/agentcube/issues/74


hzxuzhonghu:
We can do another round of update after this merged

PR #89, opened 2025-12-21
add readinessProbe

LeslieKuo:
fix issue https://github.com/volcano-sh/agentcube/issues/64

### DIFF
--- a/pkg/workloadmanager/handlers.go
+++ b/pkg/workloadmanager/handlers.go
@@ -112,7 +112,7 @@ func (s *Server) handleCreateSandbox(c *gin.Context) {
 	case strings.HasSuffix(reqPath, "/code-interpreter"):
 		sandboxReq.Kind = types.CodeInterpreterKind
 	default:
-	}    
+	}
 
 	if err := sandboxReq.Validate(); err != nil {
 		klog.Errorf("request body validation failed: %v", err)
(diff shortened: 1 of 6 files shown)


pkg/common/types/sandbox.go
```diff
@@ -33,6 +33,35 @@ type CreateSandboxRequest struct {
 	Metadata           map[string]string `json:"metadata"`
 	PublicKey          string            `json:"publicKey,omitempty"`
 	InitTimeoutSeconds int               `json:"initTimeoutSeconds,omitempty"`
+	ReadinessProbe     *ReadinessProbe   `json:"readinessProbe,omitempty"`
```
hzxuzhonghu, line 36: we donot need to pass thie probe config in the request body, the probe config is already in the pod spec.

For code interpreter, i think we may need to add a new api field in the spec to indicate the probe

pkg/workloadmanager/handlers.go
```diff
@@ -199,6 +201,12 @@ func (s *Server) handleCreateSandbox(c *gin.Context) {
 		return
 	}
 
+	if err := s.k8sClient.WaitForSandboxDependenciesReady(c.Request.Context(), namespace, sandboxName, sandboxReadinessTimeout); err != nil {
```
hzxuzhonghu, line 204: We can just check `IsSandboxReady`, and return after ready

Issue #86, opened 2025-12-18
Memory Leak in `SandboxReconciler` map on request timeout or deletion


acsoto:
### Current Behavior
In `pkg/workloadmanager/sandbox_controller.go`, `WatchSandboxOnce` registers a channel in `pendingRequests` to wait for a Sandbox to become `Running`.

If the `handleCreateSandbox` request times out (default 3m), the function returns, but the entry in `pendingRequests` is **not removed**.

Although the code comment says `// it will be deleted in GarbageCollection`, this refers to the Kubernetes **CRD resource**, not the internal **memory map entry**.

When the Sandbox CR is eventually deleted (by GC or manually), the `Reconcile` loop returns early because `r.Get` returns `NotFound`:

```go
	if err := r.Get(ctx, req.NamespacedName, sandbox); err != nil {
		return ctrl.Result{}, client.IgnoreNotFound(err)
	}
    // The cleanup logic below is NEVER reached if resource is deleted
	if status == "running" {
		// ...
		delete(r.pendingRequests, req.NamespacedName)
	}
```

### The Problem
If a Sandbox fails to start (timeout) and is subsequently deleted, the `Reconcile` loop never cleans up the `pendingRequests` entry. This causes the `pendingRequests` map to grow indefinitely with stale entries, leading to a **memory leak**.

### Proposed Solution
1.  **Modify `WatchSandboxOnce`**: It should return a cleanup function (closure).
2.  **Update `handleCreateSandbox`**: Call this cleanup function using `defer` or explicitly in the timeout case to ensure the map entry is removed regardless of the outcome.

Alternatively, handle the `NotFound` error in `Reconcile` to checks if there are pending requests for the deleted resource and clean them up.

PR #90, opened 2025-12-22
fix:memory leak in SandboxReconciler map

FAUST-BENCHOU:
Closes https://github.com/volcano-sh/agentcube/issues/86

### DIFF
--- a/pkg/workloadmanager/sandbox_controller.go
+++ b/pkg/workloadmanager/sandbox_controller.go
@@ -2,10 +2,8 @@ package workloadmanager
 
 import (
 	"context"
-	"fmt"
 	"sync"
 
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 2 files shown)


pkg/workloadmanager/sandbox_controller.go
```diff
@@ -75,5 +85,15 @@ func (r *SandboxReconciler) WatchSandboxOnce(_ context.Context, namespace, name
...
 
-	return resultChan
+	// Return cleanup function that removes the entry from pendingRequests
+	cleanup := func() {
+		r.mu.Lock()
+		if _, exists := r.pendingRequests[key]; exists {
+			klog.Infof("Cleaning up pending request for sandbox %s/%s", key.Namespace, key.Name)
+			delete(r.pendingRequests, key)
+		}
+		r.mu.Unlock()
```
hzxuzhonghu: prefer using defer

pkg/workloadmanager/sandbox_controller.go
```diff
@@ -75,5 +85,15 @@ func (r *SandboxReconciler) WatchSandboxOnce(_ context.Context, namespace, name
 	fmt.Printf("Registered for future notification for sandbox %s/%s\n", key.Namespace, key.Name)
 	r.mu.Unlock()
 
-	return resultChan
+	// Return cleanup function that removes the entry from pendingRequests
+	cleanup := func() {
```
hzxuzhonghu: why donot you provide a `UnWatchSandbox` method instead of returning the cleanup

pkg/workloadmanager/sandbox_controller.go
```diff
@@ -75,5 +85,15 @@ func (r *SandboxReconciler) WatchSandboxOnce(_ context.Context, namespace, name
 	fmt.Printf("Registered for future notification for sandbox %s/%s\n", key.Namespace, key.Name)
```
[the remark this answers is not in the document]
hzxuzhonghu: use klog

pkg/workloadmanager/sandbox_controller.go
hzxuzhonghu, line 50: THis can be done as a todo: move the belo logic into above branch to prevent double check

pkg/workloadmanager/sandbox_controller.go
```diff
@@ -29,6 +30,15 @@ type SandboxStatusUpdate struct {
...
 	sandbox := &sandboxv1alpha1.Sandbox{}
 	if err := r.Get(ctx, req.NamespacedName, sandbox); err != nil {
+		// If resource is deleted (NotFound), clean up any pending requests to prevent memory leak
+		if apierrors.IsNotFound(err) {
+			r.mu.Lock()
+			if _, exists := r.pendingRequests[req.NamespacedName]; exists {
+				klog.Infof("Cleaning up pending request for deleted sandbox %s/%s", req.Namespace, req.Name)
+				delete(r.pendingRequests, req.NamespacedName)
+			}
+			r.mu.Unlock()
```
hzxuzhonghu: THis is not needed, whoever watch is responsible to unwatch

Issue #88, opened 2025-12-19
Umbrella: strengthen unit test coverage across AgentCube components


hzxuzhonghu:
## Summary
- Expand unit test coverage across AgentCube control plane, data plane, and tooling.
- Establish baseline coverage targets (>=70% for Go packages, smoke coverage for Python CLI/SDK) with CI reporting.

## Motivation
- Protect the rapidly evolving proposal codebase from regressions as we bring components online.
- Build confidence for future contributors and reviewers by providing fast feedback loops.

## Goals
- Capture current coverage numbers and document the biggest gaps.
- Add focused unit tests for critical behaviors across the components listed below.
- Wire coverage reporting into CI (Go and Python) so we can track improvements over time.

### Go control plane & services
- [ ] `pkg/agentd`: cover runtime registration, lifecycle orchestration, and error paths.
- [ ] `pkg/picod`: exercise execution pipeline, auth helpers, and filesystem utilities.
- [ ] `pkg/router`: validate session manager, HTTP/WebSocket handlers, and config parsing.
- [ ] `pkg/workloadmanager`: focus on sandbox/code interpreter controllers, auth/JWT helpers, and workload builder.
- [ ] `pkg/store`: ensure interface contract tests plus Redis/Valkey adapters (consider fake clients or testcontainers).
- [ ] `pkg/common/types` & shared utilities: add serialization/validation tests where missing.

### Binaries & entrypoints
- [ ] `cmd/agentd`, `cmd/router`, `cmd/workload-manager`, `cmd/picod`: ensure flag parsing and wiring logic have basic coverage (e.g., using in-memory config).

### CLI & Python SDK
- [ ] `cli/agentcube`: add pytest coverage for command behaviors and error handling.
- [ ] `sdk-python`: cover agentcube clients/utilities with mocked HTTP interactions.

### Tooling & CI
- [ ] Update `Makefile`/CI to run `go test ./... -coverprofile` and surface coverage summary.
- [ ] Add Python test job (pytest) with coverage report for CLI/SDK.
- [ ] Document how to run the test suites locally (README or docs/devguide) and set initial coverage targets.

## Exit Criteria
- Checklist above is tackled via focused follow-up issues/PRs linked back here.
- Coverage targets and CI jobs are in place, with documentation updated accordingly.

Please link follow-up issues/PRs to this umbrella to track progress.

hzxuzhonghu:
Anyone interested can take over a part of this

[...]

PR #96, opened 2025-12-23
test:pkg/agentd and ci test coverage

FAUST-BENCHOU:
Link to https://github.com/volcano-sh/agentcube/issues/88
See https://github.com/volcano-sh/agentcube/actions/runs/20459304547 as an example

### DIFF
--- a/.github/workflows/test-coverage.yml
+++ b/.github/workflows/test-coverage.yml
@@ -1,0 +1,69 @@
+name: Test Coverage
+
+on:
+  pull_request:
+    paths-ignore:
+      - '**.md'
+      - '**.svg'
+      - '**.png'
+  merge_group: # enable merge queue
+  workflow_call:
+    secrets:
+      CODECOV_TOKEN:
+        required: false
+
+jobs:
+  coverage:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Free Disk Space
+        uses: jlumbroso/free-disk-space@v1.3.1
+        with:
+          tool-cache: false
+          android: true
+          dotnet: true
+          haskell: true
+          large-packages: true
+          docker-images: true
+          swap-storage: true
+
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.24'
+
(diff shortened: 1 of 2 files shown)


.github/workflows/test-coverage.yml
```diff
@@ -0,0 +1,69 @@
...
+          echo "### Coverage Summary" >> $GITHUB_STEP_SUMMARY
+          go tool cover -func=coverage.out | tail -1 >> $GITHUB_STEP_SUMMARY
+
+      - name: Upload coverage reports to Codecov
+        uses: codecov/codecov-action@v4
+        with:
+          # Even though token upload token is not required for public repos,
+          # but adding a token might increase successful uploads as per:
+          # https://community.codecov.com/t/upload-issues-unable-to-locate-build-via-github-actions-api/3954
+          token: ${{ secrets.CODECOV_TOKEN }}
```
hzxuzhonghu, line 58: what permission does it require

FAUST-BENCHOU: Sir, Codecov does not explicitly mention the permission requirements for the secret, only that the repository’s corresponding token needs to be set. However, logically, public repositories should not require a secret to function (my forked repository, however, does require setting a secret).
image

hzxuzhonghu: I understand that it is to prevent ratelimited by github. I guess we can generate a token with only read permission for this repo

pkg/agentd/agentd_test.go
```diff
@@ -118,28 +203,197 @@ func TestReconciler_Reconcile_WithLastActivity(t *testing.T) {
...
+			// Verify sandbox still exists (should not be deleted for these cases)
+			sandbox := &sandboxv1alpha1.Sandbox{}
+			err = fakeClient.Get(context.Background(), types.NamespacedName{
+				Name:      tt.sandbox.Name,
+				Namespace: tt.sandbox.Namespace,
+			}, sandbox)
+			assert.NoError(t, err, "Sandbox should still exist")
+
+			// Future time should result in requeue
+			if tt.name == "Sandbox with future last-activity-time should not be deleted" {
```
hzxuzhonghu: This is too hack

PR #114, opened 2025-12-30
doc: Add proposal for PicoD Plain Authentication Design

LaynePeng:
**What type of PR is this?**
/kind documentation
/kind feature

**What this PR does / why we need it:**
Currently, AgentCube’s `picod` daemon enforces authentication using a client self-signed key-pair mechanism, which binds a specific client directly to a `picod` instance. While secure, this 1:1 model is incompatible with centralized gateway architectures where an upstream Router is responsible for authentication and session management.

This PR introduces a design proposal for **PicoD Plain Authentication**.

**Key highlights of the proposal:**
1.  **Shared Identity Model**: Introduces a Router-managed identity where Router replicas share a Private Key (Secret) to sign tokens, and `picod` instances share a Public Key (ConfigMap) to verify them.
2.  **Decoupled Provisioning**: The WorkloadManager injects the Public Key into `picod` via Environment Variables (`PICOD_AUTH_PUBLIC_KEY`), removing the need for complex volume mounts or key payloads in the API.
3.  **Simplified Auth Flow**: Supports standard JWT-based authentication, allowing the Python SDK to connect seamlessly to sandbox instances allocated by the Router.
4.  **High Availability**: Includes a leader-elected bootstrap mechanism for Routers to automatically generate and synchronize keys across replicas.

This design enables "Gateway-Managed Sandbox Access" scenarios, simplifying client workflows while maintaining robust security boundaries.

### DIFF
--- a/docs/design/PicoD-Plain-Authentication-Design.md
+++ b/docs/design/PicoD-Plain-Authentication-Design.md
@@ -11,6 +11,7 @@ However, emerging use cases require a more flexible architecture where the clien
 *   Managing the security of connections between clients (agents) and the service.
 
 The existing self-signed key-pair model is incompatible with this centralized management flow, as it bypasses the Router's ability to mediate access. To address this, we propose a new **Plain Authentication** mechanism for `picod`. This design enables the Router/Gateway to manage credentials and connection security, simplifying the client-side workflow while maintaining robust access control.
+
 ## Use Cases
 
 ### Gateway-Managed Sandbox Access
@@ -25,13 +26,12 @@ To ensure **High Availability (HA)** across multiple Router replicas and enfor
 
 1. **Shared Authority (Router)**:
     
-    - All Router replicas share a single cryptographic identity to function as a unified Certificate Authority (CA).
-    - **Private Key Storage**: Stored in a Kubernetes Secret (picod-router-identity). This is restricted to the Router namespace/service account.
+    - All Router replicas share a single cryptographic identity to function as a unified Token Issuer.
+    - **Private Key Storage**: Stored in a Kubernetes Secret (picod-router-identity). The Private Key is accessible only by the Router component.
     - **Public Key Distribution**: Published to a Kubernetes ConfigMap (picod-router-public-key). This is accessible by the WorkloadManager and PicoD instances.
         
 2. **Decoupled Provisioning (WorkloadManager)**:
     
-    - The WorkloadManager is relieved of cryptographic payload management.
     - It provisions sandboxes by injecting the Public Key into the picod container as an Environment Variable (PICOD_AUTH_PUBLIC_KEY).
         
 3. **Local Verification (PicoD)**:


docs/design/PicoD-Plain-Authentication-Design.md
```diff
@@ -0,0 +1,220 @@
...
+}
+```
+
+## Scope & Constraints
+
+### In Scope
+
+1. **Router Identity Management**: Implementation of the "Bootstrap" logic to atomically create/load keys from Kubernetes.
+2. **JWT Implementation**: Signing logic in Router and verification logic in PicoD.
+3. **WorkloadManager Updates**: Modifying the Pod Spec generation to include the picod-router-public-key volume mount.
```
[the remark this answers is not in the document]
LaynePeng: Right, the inconsistent causes by my updating.

docs/design/PicoD-Plain-Authentication-Design.md
```diff
@@ -0,0 +1,220 @@
...
+  # Base64 encoded PKCS#8 Private Key (RSA or ECDSA)
+  private.pem: <base64-encoded-key>
+```
+
+**B. Identity ConfigMap (Public)**
+
+- **Name**: picod-router-public-key
+- **Purpose**: Stores the public key mounted into PicoD instances.
+
+codeYaml
```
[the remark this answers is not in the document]
LaynePeng: Markdown format problem, fixed

docs/design/PicoD-Plain-Authentication-Design.md
```diff
@@ -0,0 +1,220 @@
...
+        Note right of Router: Load existing Identity
+    end
+
+    Router->>Router: Load Private Key into Memory
+    Router->>K8s: APPLY ConfigMap "picod-router-public-key"
+    Note right of Router: Publishes Public Key for consumption
+
+    Note over Router, PicoD: 2. Provisioning
+    Router->>WM: Request Sandbox (No Key Payload)
+    WM->>K8s: Create Pod (EnvFrom: picod-router-public-key)
```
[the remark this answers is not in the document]
LaynePeng: Okay

docs/design/PicoD-Plain-Authentication-Design.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Design Details
+
+### Architecture Overview
+
+To ensure **High Availability (HA)** across multiple Router replicas and enforce the principle of **Least Privilege**, this design implements a **Shared Identity Model** backed by Kubernetes primitives.
+
+1. **Shared Authority (Router)**:
+    
+    - All Router replicas share a single cryptographic identity to function as a unified Certificate Authority (CA).
```
hzxuzhonghu: It not works as CA, we donot generate x509 cert here. It is jwt token issuer

docs/design/PicoD-Plain-Authentication-Design.md
```diff
@@ -0,0 +1,218 @@
...
+    - **Private Key Storage**: Stored in a Kubernetes Secret (picod-router-identity). This is restricted to the Router namespace/service account.
```
hzxuzhonghu: >This is restricted to the Router namespace/service account. 

This is confusing

docs/design/PicoD-Plain-Authentication-Design.md
```diff
@@ -0,0 +1,218 @@
...
+    - **Public Key Distribution**: Published to a Kubernetes ConfigMap (picod-router-public-key). This is accessible by the WorkloadManager and PicoD instances.
+        
+2. **Decoupled Provisioning (WorkloadManager)**:
+    
+    - The WorkloadManager is relieved of cryptographic payload management.
```
hzxuzhonghu: we never do cryptographic payload management

docs/design/PicoD-Plain-Authentication-Design.md
```diff
@@ -0,0 +1,218 @@
...
+
+### Architecture Overview
+
+To ensure **High Availability (HA)** across multiple Router replicas and enforce the principle of **Least Privilege**, this design implements a **Shared Identity Model** backed by Kubernetes primitives.
+
+1. **Shared Authority (Router)**:
+    
+    - All Router replicas share a single cryptographic identity to function as a unified Certificate Authority (CA).
+    - **Private Key Storage**: Stored in a Kubernetes Secret (picod-router-identity). This is restricted to the Router namespace/service account.
+    - **Public Key Distribution**: Published to a Kubernetes ConfigMap (picod-router-public-key). This is accessible by the WorkloadManager and PicoD instances.
```
hzxuzhonghu: I donot think this is a good practice to make workload manager startup depend on `router`.  In workload manager we can use downward api to project the public key to the code interpreter without reading the key

docs/design/PicoD-Plain-Authentication-Design.md
```diff
@@ -0,0 +1,218 @@
...
+#### 1. Bootstrap Phase (Concurrency Control)
+
+Upon startup, every Router replica executes an **Atomic Initialization Routine** to resolve the shared identity. This logic handles race conditions using Kubernetes' optimistic locking capabilities:
+
+1. **Identity Acquisition**: The Router attempts to retrieve the picod-router-identity Secret.
+    
+    - If Missing: The Router generates a new RSA/ECDSA key pair in memory and attempts to **CREATE** the Secret.
+    - Concurrency Handling: If the creation fails with 409 Conflict (implying another replica initialized it simultaneously), the Router discards its generated key and fetches the existing Secret created by the peer.
+        
+2. **Public Key Publication**: Once the Private Key is successfully loaded, the Router reconciles the picod-router-public-key ConfigMap. It ensures the Public Key in the ConfigMap matches the Private Key in memory.
```
hzxuzhonghu, line 55: we can store both private key and public key into a secret.

docs/design/PicoD-Plain-Authentication-Design.md
hzxuzhonghu, line 31: Be careful not to make startup depend on the existence of ConfigMap

docs/design/PicoD-Plain-Authentication-Design.md
```diff
@@ -0,0 +1,218 @@
...
+}
+```
+
+## Scope & Constraints
+
+### In Scope
+
+1. **Router Identity Management**: Implementation of the "Bootstrap" logic to atomically create/load keys from Kubernetes.
+2. **JWT Implementation**: Signing logic in Router and verification logic in PicoD.
+3. **WorkloadManager Updates**: Modifying the Pod Spec generation to include the `PICOD_AUTH_PUBLIC_KEY` environment variable.
```
hzxuzhonghu, line 212: Can you supplement how we reuse the token, how we rotate it

hzxuzhonghu: Do we share token  across different router instances, how to do it

docs/design/PicoD-Plain-Authentication-Design.md
LaynePeng: Okay，I will correct this to 'unified Token Issuer'.

docs/design/PicoD-Plain-Authentication-Design.md
LaynePeng: No? ...Okay, I will simplily remove this line.

docs/design/PicoD-Plain-Authentication-Design.md
LaynePeng: Yes, but don't you think we seperating them has two main advatages:
1. . The Private Key (Secret) is restricted to the Router, while the Public Key (ConfigMap) can be consumed by other components (e.g. picod) without granting them GET Secret permissions.
2. Separating them simplifies concurrency handling. The Secret follows a 'Create-Only' (Immutable) pattern to handle the initial race condition. The ConfigMap, being derived data, allows for idempotent reconciliation. This decoupling makes the logic cleaner and less error-prone compared to managing a single mutable object.

docs/design/PicoD-Plain-Authentication-Design.md
LaynePeng: Ack.

docs/design/PicoD-Plain-Authentication-Design.md
LaynePeng: My intention was to state that RBAC rules to restriction. I will simplify this to:
'The Private Key is accessible only by the Router component.'"

docs/design/PicoD-Plain-Authentication-Design.md
hzxuzhonghu: 1. IIUC, this won't require picod to have secret read permission
2. Separating two objects can be non atomic, say we have created priviate secret successfully, but public key failed. Or vice verse.  Do we have a good way to recover.?

LaynePeng: Agreed. I will merge them into a single Secret to simplify the implementation. How do you think? @acsoto

docs/design/PicoD-Plain-Authentication-Design.md
LaynePeng: @VanderChen any thought?

hzxuzhonghu: This can be treated as an improvement, currently i can see from the workflow it signs a token per request, and picod also do authenticate per request. 

If we do support token cache, we can sign one token multi requests, and also in picod, it can cache the authentication result.

docs/design/PicoD-Plain-Authentication-Design.md
acsoto: Either one is ok

acsoto: Set both into one secret #115

docs/design/PicoD-Plain-Authentication-Design.md
acsoto: This can be another issue?

hzxuzhonghu: Either is ok to me, we need to file an track issue

Issue #104, opened 2025-12-26
Router redis auth info handling


t2wang:
In the file [agentcube-router.yaml](https://github.com/volcano-sh/agentcube/blob/main/k8s/agentcube-router.yaml) , user needs to specify redis address and password for the router to be functional.

Currently, there is no clear guidance on how users should be passing in these sentitive user data.

env:
  - name: REDIS_ADDR
    value: "127.0.0.1:6379"
  - name: REDIS_PASSWORD
    value: ""

hzxuzhonghu:
For controbutors who are familiar with k8s secret and volume mount.

PR #99, opened 2025-12-24
Add agentcube helm chart

t2wang:

PR #117, opened 2026-01-04
feat: Allow Redis password to be loaded from a file, enabling Kubernetes secret integration

yashisrani:
Fix: #104 

### Key Changes:

Store: 
- Updated pkg/store to support the `REDIS_PASSWORD_FILE` environment variable. 
- It now reads the password from the specified file path if `REDIS_PASSWORD` is not set.

Deployment: 
- Updated k8s/agentcube-router.yaml to mount the redis-auth secret to `/etc/agentcube/secrets` and configured the application to read the password from this location.

### DIFF
--- a/k8s/agentcube-router.yaml
+++ b/k8s/agentcube-router.yaml
@@ -76,7 +76,6 @@ spec:
             initialDelaySeconds: 5
             periodSeconds: 5
 
-
 ---
 apiVersion: v1
 kind: Service


k8s/agentcube-router.yaml
```diff
@@ -36,8 +46,8 @@ spec:
           env:
             - name: REDIS_ADDR
               value: "127.0.0.1:6379"
-            - name: REDIS_PASSWORD
-              value: ""
+            - name: REDIS_PASSWORD_FILE
+              value: "/etc/agentcube/secrets/password"
```
hzxuzhonghu: @yashisrani  

Our intention is to use downward api FYI https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#define-a-container-environment-variable-with-data-from-a-single-secret

Donot store the passwd in the filesystem, which is not allowed from security view

k8s/agentcube-router.yaml
```diff
@@ -7,6 +7,16 @@ kind: Namespace
...
 
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: redis-auth
+  namespace: agentcube
+type: Opaque
+stringData:
+  password: "SuperSecretRedisPassword"
```
hzxuzhonghu, line 18: Please wait until https://github.com/volcano-sh/agentcube/pull/99, adapt to helm chart too

PR #122, opened 2026-01-07
Adjust http timeouts to make system more reliable

### LINKED ISSUE #106: Revist timeout settings across all the components and sdk
There are many http client, server settings across the repo. We should take a look at them to see if they are correctly set to satisfy different api calls, short task execution, long tasks.

- SDK as a client, we should take a look at the code to check the client http settings
- router: works both as server for ingress clients and client for egress, including both workload manager and backend agent/code interpreter
- workload manager: works as the server for operationg sandbox, of which the create may takes longer time.
- picod: the code inerpreter works as a server to run code / cmd

We should define agentcube's upper boundary: maximum delay an api call can take.  How should we support longer task running, like an async task, this can be discussed separately.

Also should consider the connection resuse to prevent client and server side race if the idle connection timeout is set conflicting.

hzxuzhonghu:
Fix #106

### DIFF
--- a/cmd/router/main.go
+++ b/cmd/router/main.go
@@ -35,15 +35,12 @@ func main() {
 		tlsCert               = flag.String("tls-cert", "", "Path to TLS certificate file")
 		tlsKey                = flag.String("tls-key", "", "Path to TLS key file")
 		debug                 = flag.Bool("debug", false, "Enable debug mode")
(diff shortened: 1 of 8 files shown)


hzxuzhonghu:
@YaoZengzeng @VanderChen @tjucoder

PR #138, opened 2026-01-13
Fixes #133

### LINKED ISSUE #133: Why is the session ID used when deleting a sandbox?
Why is sandboxid used when sandboxes are added or updated and sessionid used when sandboxes are deleted?

func (s *SandboxStore) onSandboxDelete(obj interface{}) {
	var unstructuredObj *unstructured.Unstructured

	switch obj := obj.(type) {
	case *unstructured.Unstructured:
		unstructuredObj = obj
	case cache.DeletedFinalStateUnknown:
		unstructuredObj, _ = obj.Obj.(*unstructured.Unstructured)
	default:
		return
	}

	if unstructuredObj == nil {
		return
	}

	// Get sandbox ID from labels
	labels := unstructuredObj.GetLabels()
	sessionID := labels[SessionIdLabelKey]
	if sessionID == "" {
		return
	}

	s.mu.Lock()
	delete(s.sandboxes, sessionID)
	s.mu.Unlock()
}

acsoto:
Fixes #133

### DIFF
--- a/pkg/workloadmanager/store.go
+++ b/pkg/workloadmanager/store.go
@@ -202,16 +202,15 @@ func (s *SandboxStore) onSandboxDelete(obj interface{}) {
 		return
 	}
 
-	// Get sandbox ID from labels
-	labels := unstructuredObj.GetLabels()
-	sessionID := labels[SessionIdLabelKey]
-	if sessionID == "" {
+	sandbox, err := convertK8sSandboxToSandbox(unstructuredObj)
+	if err != nil {
 		return
 	}
-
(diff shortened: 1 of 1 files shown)


hzxuzhonghu:
>I noticed that the SandboxStore is running an Informer to watch Sandbox CRDs, but the cached data doesn't appear to be consumed anywhere. I am wondering if there are plans for this cache, or if it might be considered for cleanup? If so I will do that.


Good catch, i just removed it in #139

acsoto:
> I think this can be closed

ok

Issue #147, opened 2026-01-15
Rollback SandboxClaim when sandbox creation fails


hzxuzhonghu:
## Summary
During sandbox creation in `pkg/workloadmanager/handlers.go`, the rollback only deletes the Sandbox resource. For Code Interpreter/other flows that use `SandboxClaim`, failures after the claim is created leave the claim behind, despite the TODO indicating it should be rolled back too.

## Where
- `handleCreateSandbox` rollback logic (around the TODO at `sandboxRollbackFunc`).
- The rollback path calls `deleteSandbox(...)` unconditionally, never `deleteSandboxClaim(...)` when a claim was created.

## Impact
- When creation fails after a `SandboxClaim` is created (e.g., waiting for the sandbox pod IP or store update errors), the claim remains in the cluster.
- Users may see leaked `SandboxClaim` objects and potential name conflicts on retry.

## Steps to reproduce
1. Create a Code Interpreter sandbox (kind `SandboxClaim`).
2. Induce a failure after the claim is created but before the handler finishes (e.g., make `GetSandboxPodIP` fail or cause a store update error).
3. Observe that the `SandboxClaim` is not deleted during rollback.

## Expected
- Rollback should delete the created `SandboxClaim` when the request path used a claim.
- Sandbox rollback should respect which resource type was created (Sandbox vs. SandboxClaim).

## Actual
- Rollback always calls `deleteSandbox`, so `SandboxClaim` remains.

## Proposed fix
- Track whether a claim was created (e.g., `sandboxClaim != nil`) and call `deleteSandboxClaim` in rollback for that case; otherwise delete the sandbox as today.
- Consider consolidating the rollback helper to handle both resource types and keep the timeout and logging consistent.

## Additional context
- Comment in code: `// TODO(hzxuzhonghu): in some case we need to rollback sandboxClaim` (current branch `router-cleanup`).

PR #153, opened 2026-01-19
Rollback SandboxClaim when sandbox creation fails

FAUST-BENCHOU:
Closes https://github.com/volcano-sh/agentcube/issues/147


hzxuzhonghu (approved):
/lgtm
/approve

It would be bettet to add ut coverage for this corner case later

PR #171, opened 2026-01-23
workloadmanager: prefer ENV for Router public key with Secret fallback

### LINKED ISSUE #60: Using ENV instead of secret for codeinterpreter sandbox public key injection
In the current version, the public key from workloadmgr is injected into codeinterpreter by mounting a secret, which is used for authentication on the init interface.

However, this approach brings two issues:
1. Secrets cannot be mounted across namespaces.
2. Multiple workloadmgr instances will have multiple public-private key pairs.

The solution is to modify the implementation to use ENV to mount the public key into the codeinterpreter sandbox.

aaradhychinche-alt:
This PR completes the fix proposed in Issue #60 by making WorkloadManager prefer loading the Router public key from an environment variable, while keeping the existing Secret-based mechanism as a fallback for backward compatibility.

Previously, although PicoD already supported ENV-based public key loading and CodeInterpreter sandboxes received the key via environment variables, WorkloadManager still depended on polling the picod-router-identity Secret. This created namespace coupling and operational complexity in multi-controller setups.

Changes included
Core logic

Added AGENTCUBE_ROUTER_PUBLIC_KEY as the preferred source for Router’s public key.

Introduced ENV-first loading logic with fallback to Secret polling.

Added clear logging to indicate whether ENV or Secret was used.

Added workloadmanager.routerPublicKey to Helm values.

Updated WorkloadManager template to inject the ENV variable when configured.

Documentation :-

Updated PicoD authentication design docs to reflect the new ENV-first flow.

Added usage examples and migration guidance.

fixes #60

### DIFF
--- a/docs/design/PicoD-Plain-Authentication-Design.md
+++ b/docs/design/PicoD-Plain-Authentication-Design.md
@@ -65,8 +65,10 @@ Upon startup, every Router replica executes an **Atomic Initialization Routine*
 #### 2. Provisioning Phase
 
 - The **Router** sends a sandbox allocation request to the **WorkloadManager**. Crucially, this request **does not** contain key data.
-- The **WorkloadManager** constructs the Pod specification. It defines an environment variable `PICOD_AUTH_PUBLIC_KEY` that sources its value from the `picod-router-public-key` ConfigMap (using `valueFrom: configMapKeyRef`).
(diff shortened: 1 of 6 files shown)


hzxuzhonghu:
Sorry , this is not align with the usage now. unless we are going to support other deploy way

aaradhychinche-alt:
@hzxuzhonghu got it thank You for the response !! Should i close it then ?? !!

PR #172, opened 2026-01-25
Feat: Optimize pod cache to reduce memory usage

### LINKED ISSUE #46: Replace controller-runtime with raw k8s informer
Controller runtime wraps all the cache for all the resources we access in the cache. This will cost much more memory than we need. 

Say in reading the pod, actually we only care about pods created for sandboxes, but controller runtime would cache all pods locally

mahil-2040:
Closes #46

## Overview

This document describes how to test the pod cache optimization that reduces memory usage in the workload-manager by implementing selective caching for pods.

## Problem Statement

The controller-runtime wraps all cache for all resources we access. This costs much more memory than needed. For example, when reading pods, we only care about pods created for sandboxes (those with the `sandbox-name` label), but controller-runtime would cache ALL pods in the cluster locally.

## Solution Implemented

The solution implements a **filtered cache** using controller-runtime's `cache.Options` with `ByObject` configuration. This ensures that only pods with the `sandbox-name` label are cached, significantly reducing memory footprint.

### Changes Made

**File: `cmd/workload-manager/main.go`**

1. Added filtered cache configuration to the controller manager:
   ```go
   Cache: cache.Options{
       ByObject: map[client.Object]cache.ByObject{
           &corev1.Pod{}: {
               Label: labels.SelectorFromSet(labels.Set{
                   "sandbox-name": "", // Only cache pods with this label
               }),
           },
       },
   }
   ```

2. Added required imports:
   - `corev1 "k8s.io/api/core/v1"`
   - `"k8s.io/apimachinery/pkg/labels"`
   - `"sigs.k8s.io/controller-runtime/pkg/cache"`
   - `"sigs.k8s.io/controller-runtime/pkg/client"`

### Label-Based Filtering

All sandbox pods are labeled with `sandbox-name` in the workload builder:

```go
// From pkg/workloadmanager/workload_builder.go
sandbox.Spec.PodTemplate.ObjectMeta.Labels["sandbox-name"] = params.sandboxName
```

The filtered cache uses this label as a selector, ensuring only relevant pods are cached.

### Pod Access Points

The workload-manager accesses pods in the following locations:

1. **`pkg/workloadmanager/k8s_client.go`** - `GetSandboxPodIP()`:
   - Uses `podLister.Pods(namespace).Get(podName)` for direct lookup
   - Uses `podLister.Pods(namespace).List(labels.SelectorFromSet(map[string]string{"sandbox-name": sandboxName}))` for label-based lookup
   - Both methods will work correctly with the filtered cache

2. **`pkg/workloadmanager/garbage_collection.go`**:
   - Does NOT directly access pods
   - Works with Sandbox/SandboxClaim resources only
## Memory Usage Validation

To validate the filtered pod cache optimization, I tested memory usage of
`workloadmanager` in a Kind cluster with many non-sandbox pods.

### Test Setup
- ~60 total pods across multiple namespaces (non-sandbox)
- 5–6 sandbox pods created via CodeInterpreter sessions
- WorkloadManager deployed with filtered controller-runtime cache
- Memory measured directly from the Kind node process (metrics-server is unstable in Kind)

### Baseline 

Screenshot 2026-01-25 191525

---
### After optimization
image

--- 

| Scenario | Sandbox Pods | RSS Memory |
|--------|--------------|------------|
| Baseline | 0 | ~42 MB |
| After optimization | 5–6 | ~44 MB |

### Conclusion
Memory usage increases only slightly with sandbox pods and does not scale
with the total number of non-sandbox pods in the cluster, confirming that
the filtered cache behaves as intended.

### DIFF
--- a/pkg/workloadmanager/k8s_client.go
+++ b/pkg/workloadmanager/k8s_client.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/dynamic/dynamicinformer"
 	"k8s.io/client-go/informers"
@@ -121,13 +122,17 @@ func NewK8sClient() (*K8sClient, error) {
 
 	// Create informer factory for core resources (Pods, etc.)
 	// Use filtered factory to only cache pods with sandbox-name label
+	requirement, err := labels.NewRequirement(SandboxNameLabelKey, selection.Exists, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create label requirement: %w", err)
+	}
 	informerFactory := informers.NewFilteredSharedInformerFactory(
 		clientset,
 		0, // resync period
 		metav1.NamespaceAll,
 		func(opts *metav1.ListOptions) {
 			// Filter to only watch pods with sandbox-name label
-			opts.LabelSelector = SandboxNameLabelKey
+			opts.LabelSelector = requirement.String()
 		},
 	)
 


hack/update-codegen.sh
acsoto, line 1: permissions changed from 755 (executable) to 644 (non-executable)?

pkg/workloadmanager/k8s_client.go
```diff
@@ -118,7 +118,16 @@ func NewK8sClient() (*K8sClient, error) {
...
+	// Use filtered factory to only cache pods with sandbox-name label
+	informerFactory := informers.NewFilteredSharedInformerFactory(
+		clientset,
+		0, // resync period
+		metav1.NamespaceAll,
+		func(opts *metav1.ListOptions) {
+			// Filter to only watch pods with sandbox-name label
+			opts.LabelSelector = "sandbox-name"
+		},
+	)
```
acsoto, line 137: This is redundant and creates inconsistency. The workload-manager uses both caches:
     - Controller-runtime cache
     - Client-go informers

cmd/workload-manager/main.go
```diff
@@ -68,13 +72,26 @@ func main() {
...
 	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
 
-	// Setup controller manager
+	// Setup controller manager with filtered cache for Pods
+	// Only cache pods that have the "sandbox-name" label to reduce memory usage
+	labelSelector, err := labels.Parse("sandbox-name")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "unable to parse label selector: %v\n", err)
+		os.Exit(1)
+	}
```
acsoto: This parses "sandbox-name" as a label key existence check. But the original code used:
```
labels.SelectorFromSet(labels.Set{"sandbox-name": ""})
```

This will not fail

pkg/workloadmanager/k8s_client.go
hzxuzhonghu: The goal is to unify the usage

cmd/workload-manager/main.go
hzxuzhonghu: prefer selector, err := labels.NewRequirement(workload.GroupNameLabelKey, selection.Exists, nil)

hack/update-codegen.sh
mahil-2040: Sorry, this was unintentional. The permission changes happened due to Windows/WSL file system differences when I was working on the PR. I'll revert these permission-only changes and keep only the actual code modifications (main.go and k8s_client.go).

cmd/workload-manager/main.go
mahil-2040: @hzxuzhonghu  Great suggestion! I've updated the code to use `labels.NewRequirement` with `selection.Exists`. It's much clearer and more explicit. Thanks!

pkg/workloadmanager/k8s_client.go
mahil-2040: @hzxuzhonghu Thanks for the clarification! Yes, the goal is to ensure both caching mechanisms are consistently filtered:

1. **Controller-runtime cache** (main.go) - Used by Sandbox/CodeInterpreter reconcilers
2. **Client-go informer cache** (k8s_client.go) - Used by the API server for pod IP lookups

Without filtering both, the client-go informer would still cache all cluster pods, which would defeat the purpose of the memory optimization. By filtering both, we ensure:
- Consistent behavior across all pod access points
- Maximum memory reduction (both caches only store sandbox pods)
- No confusion about which cache contains what data

The memory savings come from reducing BOTH caches, not just one.

pkg/workloadmanager/k8s_client.go
```diff
@@ -118,7 +120,16 @@ func NewK8sClient() (*K8sClient, error) {
...
 	// Create informer factory for core resources (Pods, etc.)
-	informerFactory := informers.NewSharedInformerFactory(clientset, 0)
+	// Use filtered factory to only cache pods with sandbox-name label
+	informerFactory := informers.NewFilteredSharedInformerFactory(
+		clientset,
+		0, // resync period
+		metav1.NamespaceAll,
+		func(opts *metav1.ListOptions) {
+			// Filter to only watch pods with sandbox-name label
+			opts.LabelSelector = SandboxNameLabelKey
```
hzxuzhonghu: requirement, err := labels.NewRequirement(workloadmanager.SandboxNameLabelKey, selection.Exists, nil)
opts.LabelSelector = requirement.String()

cmd/workload-manager/main.go
```diff
@@ -68,13 +73,28 @@ func main() {
...
 		Scheme: schemeBuilder,
 		Metrics: metricsserver.Options{
 			BindAddress: "0", // Disable metrics server
 		},
 		HealthProbeBindAddress: "0", // Disable health probe server
+		Cache: cache.Options{
+			ByObject: map[client.Object]cache.ByObject{
+				&corev1.Pod{}: {
+					Label: labelSelector,
+				},
```
hzxuzhonghu: can you explain why add pod cache

pkg/workloadmanager/k8s_client.go
```diff
@@ -51,6 +51,8 @@ var (
 	SessionIdLabelKey = "runtime.agentcube.io/session-id" // revive:disable-line:var-naming - keep label backward compatible
 	// WorkloadNameLabelKey labels key for workload name
 	WorkloadNameLabelKey = "runtime.agentcube.io/workload-name"
+	// SandboxNameLabelKey labels key for sandbox name
+	SandboxNameLabelKey = "sandbox-name" // revive:disable-line:var-naming
```
hzxuzhonghu:
```suggestion
	SandboxNameLabelKey = "runtime.agentcube.io/sandbox-name" 
```

cmd/workload-manager/main.go
mahil-2040: Sure! The pod cache is needed because the workload-manager needs to look up pod IPs when routing requests to sandbox sessions. 

When a session is created, we need to find the corresponding pod's IP address to forward traffic to it. The `GetSandboxPodIP` function (in k8s_client.go) uses the pod lister to find pods by their sandbox-name label.

Without caching, we'd have to query the Kubernetes API server every time we need a pod IP, which would be slow and put unnecessary load on the API server. The cache keeps a local copy of pod information so lookups are fast.

The optimization here is that instead of caching ALL pods in the cluster (which wastes memory), we only cache the pods we actually care about - the ones with the sandbox-name label.

pkg/workloadmanager/k8s_client.go
mahil-2040: I've updated the client-go informer filter to use the same explicit approach.

pkg/workloadmanager/k8s_client.go
mahil-2040: @hzxuzhonghu I’m concerned this change could break backward compatibility. Copilot is suggesting that while "sandbox-name" is already used by existing sandbox pods, changing it to "runtime.agentcube.io/sandbox-name" would cause the filtered cache to miss existing pods, and GetSandboxPodIP may fail during rolling upgrades. Should we keep "sandbox-name" to maintain compatibility, or do you have a migration strategy in mind?

hzxuzhonghu: sooner or later we should update, not worry about compatibility now, we have not released yet

cmd/workload-manager/main.go
hzxuzhonghu: I think GetSandboxPodIP already using podLister.

pkg/workloadmanager/k8s_client.go
mahil-2040: So i should implement this change now, right, ignoring the already existing pods?

hzxuzhonghu: I think we can do now

cmd/workload-manager/main.go
hzxuzhonghu: donot miss this, i think we can revert this change @mahil-2040

pkg/workloadmanager/k8s_client.go
mahil-2040: ok i will implement it and push the changes

cmd/workload-manager/main.go
hzxuzhonghu: @mahil-2040

mahil-2040: Yes I am working on this. Give me some time I will revert the changes.

hzxuzhonghu: Sorry, i thought you missed it

hzxuzhonghu (commented):
/lgtm

PR #176, opened 2026-01-28
avoid panic on client disconnect

### LINKED ISSUE #78: Router Panic during prolonged agent response
Using an condeact agent that triggers multiple api calls to the code interpreters and LLM, a single invoke call could take dozens of seconds to the agent to finalize its response.

In this scenario, there is a panicing issue from the router.

Image

From the logs, we can see that the agent itself makes multiple successful calls to the code interpreters through the router; however, an issue likely arises when returning the result from the agent to the user.

Image

Related logs from the agent:

Image

ask-elad:
**What type of PR is this?**
/kind bug

This PR fixes a router panic that occurs when a client disconnects during a long running agent invocation.
The router uses httputil.ReverseProxy to forward requests, and when the client times out or disconnects, the standard library panics with http.ErrAbortHandler. This panic was previously unhandled, causing the router to crash and Gin to emit recovery warnings.

The change treats client disconnects as a normal early exit by catching http.ErrAbortHandler at the proxy boundary, preventing the router from panicking while preserving existing behavior for all other error cases.

Fixes #78 

Additionally tests can be written for this in a new pr if this is fine

NONE
```release-note
NONE
```

### DIFF
--- a/pkg/router/handlers.go
+++ b/pkg/router/handlers.go
@@ -142,6 +142,24 @@ func (s *Server) handleCodeInterpreterInvoke(c *gin.Context) {
 	s.handleInvoke(c, namespace, name, path, types.CodeInterpreterKind)
 }
 
+// recoverProxyAbort treats http.ErrAbortHandler from ReverseProxy as a normal client disconnect.
+func recoverProxyAbort(sandbox *types.SandboxInfo) {
+	if r := recover(); r != nil {
+		if r == http.ErrAbortHandler {
+			if sandbox != nil {
+				klog.V(2).Infof(
+					"Client disconnected, stopping proxy (session: %s)",
+					sandbox.SessionID,
+				)
+			} else {
+				klog.V(2).Info("Client disconnected, stopping proxy (session: unknown)")
+			}
+			return
+		}
+		panic(r)
+	}
+}
+
 // forwardToSandbox forwards the request to the specified sandbox endpoint
 func (s *Server) forwardToSandbox(c *gin.Context, sandbox *types.SandboxInfo, path string) {
 	// Extract url from sandbox - find matching entry point by path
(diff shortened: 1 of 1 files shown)


pkg/router/handlers.go
```diff
@@ -142,6 +142,24 @@ func (s *Server) handleCodeInterpreterInvoke(c *gin.Context) {
...
+				)
+			} else {
+				klog.V(2).Info("Client disconnected, stopping proxy (session: unknown)")
+			}
+			return
+		}
+		panic(r)
+	}
+}
+
```
acsoto, line 162: we could flatten the nested if statements in recoverProxyAbort to make it easier to read.

pkg/router/handlers.go
```diff
@@ -142,6 +142,24 @@ func (s *Server) handleCodeInterpreterInvoke(c *gin.Context) {
...
+				klog.V(2).Infof(
+					"Client disconnected, stopping proxy (session: %s)",
+					sandbox.SessionID,
+				)
+			} else {
+				klog.V(2).Info("Client disconnected, stopping proxy (session: unknown)")
+			}
+			return
+		}
+		panic(r)
```
hzxuzhonghu, line 159: why do you call panic, this is not recommended in production

hzxuzhonghu (commented):
BTW, donot we have  recovery middleware

```
	v1.Use(gin.Recovery())

```

pkg/router/handlers.go
hzxuzhonghu: how do you improve?

acsoto: something like this would be better
```go
func recoverProxyAbort(sandbox *types.SandboxInfo) {
    r := recover()
    if r == nil {
        return
    }
    if r != http.ErrAbortHandler {
        panic(r)
    }
    if sandbox != nil {
        klog.V(2).Infof("Client disconnected, stopping proxy (session: %s)", sandbox.SessionID)
    } else {
        klog.V(2).Info("Client disconnected, stopping proxy (session: unknown)")
    }
}
```

PR #179, opened 2026-01-29
Optimize environment variable slice allocation in ExecuteHandler

acsoto:
Pre-allocate the slice capacity when merging environment variables in `ExecuteHandler`.

Previously, `append` was used on the result of `os.Environ()`, which could trigger multiple allocations as the slice grew. Since we know the final size (`len(os.Environ()) + len(req.Env)`), pre-allocating the capacity avoids these unnecessary allocations and copy operations.

### DIFF
--- a/pkg/picod/execute.go
+++ b/pkg/picod/execute.go
@@ -107,7 +107,9 @@ func (s *Server) ExecuteHandler(c *gin.Context) {
 
 	// Set environment variables
 	if len(req.Env) > 0 {
-		currentEnv := os.Environ()
+		environ := os.Environ()
+		currentEnv := make([]string, len(environ), len(environ)+len(req.Env))
+		copy(currentEnv, environ)
 		for k, v := range req.Env {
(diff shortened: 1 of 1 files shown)


pkg/picod/execute.go
```diff
@@ -107,7 +107,9 @@ func (s *Server) ExecuteHandler(c *gin.Context) {
 
 	// Set environment variables
 	if len(req.Env) > 0 {
-		currentEnv := os.Environ()
+		environ := os.Environ()
```
hzxuzhonghu, line 110: I want to know why do we pass os env to the exec cmd. This may expose os settings to malicious code

acsoto: Thanks for the valid security concern.

Currently, picod inherits the host environment (os.Environ()) by design to ensure the availability of the underlying toolchain like Python (e.g., PATH, HOME, LANG), as picod acts as an arbitrary command executor.

We should sanitize the environment (e.g., using a whitelist for PATH and other essentials) to prevent secret leakage.

PR #180, opened 2026-01-29
Optimize slice allocation in ListFilesHandler

acsoto:
Pre-allocate the `files` slice capacity based on the number of directory entries in `ListFilesHandler`.

Previously, `append` was used on a nil slice, causing repeated allocations as the slice grew. Pre-allocating the known size (`len(entries)`) prevents this overhead.


hzxuzhonghu (approved):
/lgtm

PR #186, opened 2026-01-30
fix(router): add auth header for workload manager

acsoto:
**What type of PR is this?**

/kind bug

Adds Authorization header for Router->Workload Manager sandbox creation using service account token, so WM auth works when enabled.

Fixes #

Router now reads  /var/run/secrets/kubernetes.io/serviceaccount/token for WM calls.

```release-note
NONE
```

### DIFF
--- a/pkg/router/session_manager.go
+++ b/pkg/router/session_manager.go
@@ -195,10 +195,6 @@ func (m *manager) createSandbox(ctx context.Context, namespace string, name stri
 }
 
 func loadWorkloadManagerAuthToken() string {
-	if token := strings.TrimSpace(os.Getenv("API_TOKEN")); token != "" {
-		return token
-	}
-
 	b, err := os.ReadFile(serviceAccountTokenPath)
 	if err != nil {
 		if !os.IsNotExist(err) {
--- a/docs/devguide/code-interpreter-python-sdk.md
+++ b/docs/devguide/code-interpreter-python-sdk.md
@@ -29,7 +29,7 @@ The `CodeInterpreterClient` is the main entry point. You can initialize it direc
 | `ttl` | `int` | `3600` | Session time-to-live (seconds) |
 | `workload_manager_url` | `str` | `None` | Control Plane URL (falls back to env `WORKLOAD_MANAGER_URL`) |
 | `router_url` | `str` | `None` | Data Plane Router URL (falls back to env `ROUTER_URL`) |
-| `auth_token` | `str` | `None` | Auth token (falls back to env `API_TOKEN`, then K8s SA token) |
+| `auth_token` | `str` | `None` | Auth token (falls back to K8s SA token) |
 | `session_id` | `str` | `None` | Reuse existing session instead of creating new |
 | `verbose` | `bool` | `False` | Enable debug logging |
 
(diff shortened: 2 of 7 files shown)


hzxuzhonghu (dismissed):
Thanks for the fix! Adding the Authorization header from API_TOKEN / service account token makes sense, and the tests cover env/file/missing/error cases well.

Non-blocking suggestions:
- Consider caching the token read if sandbox creation is frequent.
- If token file is missing in some deployments, a lower-verbosity log might avoid noise.

LGTM otherwise.

pkg/router/session_manager.go
```diff
@@ -139,6 +143,9 @@ func (m *manager) createSandbox(ctx context.Context, namespace string, name stri
 		return nil, fmt.Errorf("failed to create HTTP request: %w", err)
 	}
 	req.Header.Set("Content-Type", "application/json")
+	if token := loadWorkloadManagerAuthToken(); token != "" {
```
YaoZengzeng, line 146: So with this PR, the router will created session for us by default and the 401 issues I mentioned before will not happen again?

pkg/router/session_manager.go
```diff
@@ -186,3 +193,18 @@ func (m *manager) createSandbox(ctx context.Context, namespace string, name stri
 
 	return sandbox, nil
 }
+
+func loadWorkloadManagerAuthToken() string {
+	if token := strings.TrimSpace(os.Getenv("API_TOKEN")); token != "" {
```
hzxuzhonghu: This is not complete, how does this work with workloadmanager

acsoto: SDK uses the same logic `sdk-python/agentcube/clients/control_plane.py`
```python
token_path = "/var/run/secrets/kubernetes.io/serviceaccount/token"
token = auth_token or os.getenv("API_TOKEN") or read_token_from_file(token_path)

...
if token:
    self.session.headers["Authorization"] = f"Bearer {token}"

```

acsoto: WM `pkg/workloadmanager/auth.go`

```go
authHeader := c.GetHeader("Authorization")
...
parts := strings.SplitN(authHeader, " ", 2)
if len(parts) != 2 || parts[0] != "Bearer" { ... }
token := parts[1]

authenticated, serviceAccount, err := s.validateServiceAccountToken(...)

tokenReview := &authv1.TokenReview{ Spec: authv1.TokenReviewSpec{ Token: token } }
result, err := s.k8sClient.clientset.AuthenticationV1().TokenReviews().Create(...)
```

pkg/router/session_manager.go
acsoto: I think so because SDK uses this logic `sdk-python/agentcube/clients/control_plane.py` to create

```python
if token:
    self.session.headers["Authorization"] = f"Bearer {token}"
```

pkg/router/session_manager.go
hzxuzhonghu: I mean if the `API_TOKEN` is injected from a thirdparty identity provider, how should workload manager validate it

acsoto: That's a valid point. After reviewing the implementation, I’ve confirmed that WM now exclusively accepts K8s TokenReview with system:serviceaccount:*. Both the Router and SDK currently utilize ServiceAccount tokens. The API_TOKEN mentioned in previous versions of the SDK and documentation was a legacy feature; I have removed it to prevent further confusion.

PR #194, opened 2026-02-01
fix(auth): use strings.Fields for better Authorization header parsing

RushabhMehta2005:
**What type of PR is this?**

/kind enhancement

Currently, the code uses `strings.SplitN(authHeader, " ", 2)` to separate the Bearer prefix from the token. This is fragile because it strictly expects exactly 1 space separator. If a client sends a header with multiple spaces (example: using curl or in testing), the parsed token will incorrectly capture the leading whitespace causing failure.

NONE

NONE

```release-note
NONE
```

### DIFF
--- a/pkg/workloadmanager/auth.go
+++ b/pkg/workloadmanager/auth.go
@@ -91,7 +91,7 @@ func (s *Server) authMiddleware(c *gin.Context) {
 	}
 
 	// Check if it's a Bearer token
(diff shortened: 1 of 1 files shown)


hzxuzhonghu:
can you please rebase and push again, so the pending tests can be run

RushabhMehta2005:
@hzxuzhonghu Seems like the bot already merged this PR before I could respond to your comment and rebase. Is this of concern?

PR #207, opened 2026-02-06
fix missed lifecycle shutdown time update in sandbox helper test

acsoto:
**What type of PR is this?**

/kind bug

This is a small follow-up fix after #201. In the `agent-sandbox` v0.1.1 migration, `ShutdownTime` moved under `Lifecycle`. One test case in `pkg/workloadmanager/sandbox_helper_test.go` still used the old struct literal shape (`SandboxSpec{ShutdownTime: ...}`), which was missed in the previous PR.

This PR updates that test fixture to the v0.1.1 API shape:
`SandboxSpec{Lifecycle: {ShutdownTime: ...}}`.

ref #201

Fixes # (N/A)

Verification run:
- `go test ./pkg/workloadmanager -run TestBuildSandboxInfo_TableDriven -count=1`

```release-note
NONE
```

### DIFF
--- a/pkg/workloadmanager/auth_test.go
+++ b/pkg/workloadmanager/auth_test.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/workloadmanager/sandbox_helper_test.go
```diff
@@ -102,7 +102,9 @@ func TestBuildSandboxInfo_TableDriven(t *testing.T) {
 						CreationTimestamp: metav1.NewTime(now),
 					},
 					Spec: sandboxv1alpha1.SandboxSpec{
-						ShutdownTime: &metav1.Time{Time: shutdownTime},
```
hzxuzhonghu, line 105: Both are right, 

```
	// Lifecycle defines when and how the sandbox should be shut down.
	// +optional
	Lifecycle `json:",inline"`
```

hzxuzhonghu (approved):
/lgtm

Issue #223, opened 2026-02-24
can not reuse session for code-interpreter


warjiang:
**What happened**:
When I try the `sdk-python/examples/basic_usage.py` with `session_reuse_example` function, some problem occured: 

<img width="4774" height="1608" alt="Image" src="https://github.com/user-attachments/assets/0b9ac27c-8177-4057-9e1e-0479af1f146e" />

it seems there is no mechanism to reuse python variable across different turn in the same session

**What you expected to happen**:
reuse python var stored in the previous turn

**How to reproduce it (as minimally and precisely as possible)**:
just run `python sdk-python/examples/basic_usage.py`

**Anything else we need to know?**:

**Environment**:

- agentcube version: latest
  -  workload image url: volcano-sh/workloadmanager@sha256:2ce822a37482a0d22760276a5964f06e6e89d8e56b87caa95e8143f5bc8872cf
  - router image url:  volcano-sh/agentcube-router@sha256:ea06f30883b88b9220e1e4416046f68a1e489987ef5903ca1ef00a296331f662
  - picod: volcano-sh/picod@sha256:3d95dffd991a060ac620af8e39d2803709fa4c236df81533e7912b6fcc84a049
- Kubernetes version: v1.31.4
- Others:

hzxuzhonghu:
This is a known issue, because picod run each command separately now, there is a plan to implement stateful runner with jupyter 

cc @acsoto

acsoto:
Thanks for reporting this, and sorry for the confusion.

You are right about the current behavior: each `/api/execute` call runs in a fresh process, so Python in-memory variables are not preserved across separate `run_code` invocations. Reusing `session_id` currently preserves filesystem state, not Python variable state.

For context, I did clarify this in docs/code comments in #146, but I missed `sdk-python/examples/basic_usage.py`, which was still demonstrating variable reuse and was misleading. That miss is on me, sorry about that.

I have opened a PR to fix the example and align it with actual runtime behavior: #226

In the updated example, session reuse persists state via files (write/read `/tmp/value.txt`) instead of in-memory Python variables.

acsoto:
Thanks for raising this — we’ve thought about this too, and it’s indeed more complex with the current execution model.

warjiang:
> Thanks for raising this — we’ve thought about this too, and it’s indeed more complex with the current execution model.

Maybe we have more discussion here if any process about this feature, btw looking forward for this feature 😄  ~

PR #146, opened 2026-01-15
Fix SDK docs

acsoto:
Fix #143

This pull request updates documentation and example usage for the Python SDK to clarify session reuse behavior, improve configuration parameter explanations, and add missing file management features. The main focus is to make it clear that only file system state persists across sessions—not Python variables—and to provide better guidance for users.

**Documentation improvements:**

* Added a configuration parameters table and environment variables section to `docs/devguide/code-interpreter-python-sdk.md` for clearer setup instructions.
* Clarified session reuse: emphasized that only file system state persists, not Python variables, in both documentation and code examples. [[1]](diffhunk://#diff-5598e8ab7f94cc95e20bf7d16ead26dd00940e6405545c67c34dd26b3a0dcc88R186-R205) [[2]](diffhunk://#diff-bff4746ba0ba4adf5e43d50938988fe4e53416bf24f69b4818eb8c05fcd73e36L99-R112) [[3]](diffhunk://#diff-87c222f1a19c4e72e29ceeeff4d5aae9a779dbdb522772b312230ba62821a5eeL40-R47)

**Feature additions:**

* Documented support for running Bash scripts in addition to Python, with examples for both.
* Added documentation for the `list_files` method to show how to list files and directories in a session.

**Example updates:**

* Updated code samples in `sdk-python/README.md` and inline docstrings to demonstrate file-based state persistence instead of variable persistence. [[1]](diffhunk://#diff-bff4746ba0ba4adf5e43d50938988fe4e53416bf24f69b4818eb8c05fcd73e36L47-R49) [[2]](diffhunk://#diff-87c222f1a19c4e72e29ceeeff4d5aae9a779dbdb522772b312230ba62821a5eeL40-R47)

PR #225, opened 2026-02-24
fix sdk session reuse example

acsoto:
## Summary
This PR fixes a misleading Python SDK example for session reuse.

The runtime behavior is that each `/api/execute` invocation runs in a fresh process. Reusing `session_id` preserves filesystem state in the same sandbox, but does not preserve in-memory Python variables across separate `run_code` calls.

## User Impact
Users running `sdk-python/examples/basic_usage.py` could see a `NameError` in the session reuse section (`x` not defined) and assume session reuse was broken, even though the actual limitation is process-level execution semantics.

## Root Cause
The SDK docs and code comments were already clarified in prior work, but `sdk-python/examples/basic_usage.py` still demonstrated variable reuse (`x = 42` across calls), which contradicted real behavior and the docs.

## Fix
Updated `sdk-python/examples/basic_usage.py` to demonstrate file-based persistence across reused sessions:
- Changed section wording from generic state persistence to file-system state persistence.
- Replaced variable-based reuse example with writing `/tmp/value.txt` in the first client and reading it in the reused client.
- Updated printed step labels/comments to match actual behavior.

## Validation
- `python3 -m py_compile sdk-python/examples/basic_usage.py`

## Related
- Refs #223
- Context from prior docs clarification: #146

### DIFF
--- a/sdk-python/examples/basic_usage.py
+++ b/sdk-python/examples/basic_usage.py
@@ -18,7 +18,7 @@ AgentCube SDK Basic Usage Example
 Demonstrates:
 1. Basic command execution and code running
 2. File operations
-3. Session reuse for state persistence (useful for AI workflows)
+3. Session reuse for file-system state persistence (useful for AI workflows)
 """
 
 from agentcube import CodeInterpreterClient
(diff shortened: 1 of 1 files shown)


acsoto:
Superseded by #226 (same change with branch/title/template aligned). Closing this one to avoid confusion.

PR #241, opened 2026-03-11
Authentication and Authorization Design Proposal

mahil-2040:
**What type of PR is this?**

/kind documentation

This PR introduces a design proposal for implementing authentication and authorization in Agentcube.

Fixes #

```release-note
NONE
```

### DIFF
--- a/docs/design/auth-proposal.md
+++ b/docs/design/auth-proposal.md
@@ -88,7 +88,7 @@ When `--tls-cert-source=file` is used, the cert/key/CA files can be populated by
 
 [SPIFFE](https://spiffe.io/) (Secure Production Identity Framework for Everyone) is a CNCF graduated project that provides a standard for service identity. It defines:
 
-- **SPIFFE ID:** A URI-formatted identity, e.g., `spiffe://agentcube.local/router`
+- **SPIFFE ID:** A URI-formatted identity, e.g., `spiffe://agentcube.local/ns/agentcube-system/sa/agentcube-router`
 - **SVID (SPIFFE Verifiable Identity Document):** An X.509 certificate or JWT that proves a workload holds a given SPIFFE ID.
 
 [SPIRE](https://spiffe.io/docs/latest/spire-about/spire-concepts/) is the production implementation of SPIFFE. It has two components:
@@ -97,11 +97,6 @@ When `--tls-cert-source=file` is used, the cert/key/CA files can be populated by
 - **SPIRE Agent:** Runs on every node (DaemonSet). Performs workload attestation, verifying process identity by querying the kernel and kubelet, and delivers SVIDs to local workloads via a Unix domain socket (the Workload API).
 
 SPIRE handles the entire certificate lifecycle (issuance, rotation, revocation) automatically. Workloads never see raw keys on disk; certificates arrive through the Workload API and are rotated before they expire.
-
-<!-- PLACEHOLDER: SPIRE Architecture Diagram -->
-<!-- Diagram from https://spiffe.io/docs/latest/spire-about/spire-concepts/ showing
-     SPIRE Server at top, SPIRE Agents on each node, workloads connecting to
-     agents via the Workload API Unix domain socket. -->
 
 ### 1.2 Why Single-Cluster SPIRE
 
@@ -119,13 +114,15 @@ All AgentCube components operate under a single trust domain:
 spiffe://agentcube.local
 ```
 
-Each component receives a unique SPIFFE ID based on its Kubernetes metadata:
+Each component receives a unique SPIFFE ID following the Istio-convention format `spiffe://<trust-domain>/ns/<namespace>/sa/<service-account>`, which encodes the workload's Kubernetes namespace and service account directly in the URI path.
+
+SPIRE uses **selectors** to verify workload identity during attestation. When a workload requests its certificate, the SPIRE Agent inspects the workload's Kubernetes metadata (namespace, service account, labels) and matches it against registered selectors. A workload only receives its SVID if all registered selectors match.
 
 | Component | SPIFFE ID | Selectors |
 |---|---|---|
-| Router | `spiffe://agentcube.local/router` | `k8s:ns:agentcube-system`, `k8s:sa:agentcube-router` |
-| WorkloadManager | `spiffe://agentcube.local/workload-manager` | `k8s:ns:agentcube-system`, `k8s:sa:agentcube-workload-manager` |
-| PicoD (Sandboxes) | `spiffe://agentcube.local/picod` | `k8s:pod-label:app:picod`, `k8s:sa:agentcube-sandbox` |
+| Router | `spiffe://agentcube.local/ns/agentcube-system/sa/agentcube-router` | `k8s:ns:agentcube-system`, `k8s:sa:agentcube-router` |
+| WorkloadManager | `spiffe://agentcube.local/ns/agentcube-system/sa/agentcube-workload-manager` | `k8s:ns:agentcube-system`, `k8s:sa:agentcube-workload-manager` |
+| PicoD (Sandboxes) | `spiffe://agentcube.local/sa/agentcube-sandbox` | `k8s:pod-label:app:picod`, `k8s:sa:agentcube-sandbox` |
 
 > **Note:** PicoD cannot use a namespace selector because sandbox pods are created in the namespace specified by the user's request, not in a fixed namespace. To prevent identity spoofing in multi-tenant clusters, PicoD registration requires both the `app:picod` pod label and a dedicated `agentcube-sandbox` ServiceAccount. WorkloadManager creates this ServiceAccount in the target namespace as part of sandbox provisioning. Using a specific ServiceAccount requires RBAC permission, preventing arbitrary workloads from obtaining the PicoD SPIFFE ID.
 >
@@ -223,126 +220,154 @@ spec:
             name: spire-agent-config
 ```
 
+#### SPIRE Controller Manager
+
+The [SPIRE Controller Manager](https://github.com/spiffe/spire-controller-manager) is deployed as a `Deployment` alongside the SPIRE Server. It watches for `ClusterSPIFFEID` custom resources and automatically syncs registration entries to the SPIRE Server. This makes workload registration fully declarative and is the recommended approach for production deployments.
+
 ### 1.5 Workload Registration
 
-Registration entries are created via the SPIRE Server CLI:
+Registration entries tell the SPIRE Server which workloads should receive which SPIFFE IDs. There are two approaches depending on the environment:
+
+#### Production: ClusterSPIFFEID CRDs (Recommended)
+
+With the SPIRE Controller Manager deployed, registration entries are defined as Kubernetes CRDs. These are managed declaratively alongside other AgentCube manifests and are automatically synced to the SPIRE Server:
+
+```yaml
+# Router registration
+apiVersion: spire.spiffe.io/v1alpha1
+kind: ClusterSPIFFEID
+metadata:
+  name: agentcube-router
+spec:
+  spiffeIDTemplate: "spiffe://agentcube.local/ns/{{ .PodMeta.Namespace }}/sa/{{ .PodSpec.ServiceAccountName }}"
+  podSelector:
+    matchLabels:
+      app: agentcube-router
+  namespaceSelector:
+    matchNames:
+      - agentcube-system
+---
+# WorkloadManager registration
+apiVersion: spire.spiffe.io/v1alpha1
+kind: ClusterSPIFFEID
+metadata:
+  name: agentcube-workload-manager
+spec:
+  spiffeIDTemplate: "spiffe://agentcube.local/ns/{{ .PodMeta.Namespace }}/sa/{{ .PodSpec.ServiceAccountName }}"
+  podSelector:
+    matchLabels:
+      app: agentcube-workload-manager
+  namespaceSelector:
+    matchNames:
+      - agentcube-system
+---
+# PicoD registration (namespace-agnostic)
+apiVersion: spire.spiffe.io/v1alpha1
+kind: ClusterSPIFFEID
+metadata:
+  name: agentcube-sandbox
+spec:
+  spiffeIDTemplate: "spiffe://agentcube.local/sa/{{ .PodSpec.ServiceAccountName }}"
+  podSelector:
+    matchLabels:
+      app: picod
+```
+
+#### Development: Manual CLI
+
+For local development (e.g., `kind` clusters), registration entries can be created manually by exec-ing into the SPIRE Server pod:
 
 ```bash
-# Register the Router
-spire-server entry create \
-    -spiffeID spiffe://agentcube.local/router \
+kubectl exec -n agentcube-system <spire-server-pod> -- \
+  spire-server entry create \
+    -spiffeID spiffe://agentcube.local/ns/agentcube-system/sa/agentcube-router \
     -parentID spiffe://agentcube.local/spire-agent \
     -selector k8s:ns:agentcube-system \
     -selector k8s:sa:agentcube-router
 
-# Register the WorkloadManager
-spire-server entry create \
-    -spiffeID spiffe://agentcube.local/workload-manager \
+kubectl exec -n agentcube-system <spire-server-pod> -- \
+  spire-server entry create \
+    -spiffeID spiffe://agentcube.local/ns/agentcube-system/sa/agentcube-workload-manager \
     -parentID spiffe://agentcube.local/spire-agent \
     -selector k8s:ns:agentcube-system \
     -selector k8s:sa:agentcube-workload-manager
 
-# Register PicoD instances (namespace-agnostic, sandboxes can run in any namespace)
-# Requires both the app:picod label and the dedicated agentcube-sandbox ServiceAccount
-spire-server entry create \
-    -spiffeID spiffe://agentcube.local/picod \
+kubectl exec -n agentcube-system <spire-server-pod> -- \
+  spire-server entry create \
+    -spiffeID spiffe://agentcube.local/sa/agentcube-sandbox \
     -parentID spiffe://agentcube.local/spire-agent \
     -selector k8s:pod-label:app:picod \
     -selector k8s:sa:agentcube-sandbox
 ```
+
+> **Note:** Manual CLI entries do not survive SPIRE Server restarts (unless backed by persistent storage) and are not recommended for production use.
 
 #### Attestation Flow (Example: Router)
 
 1. Router process connects to the local SPIRE Agent via `/run/spire/sockets/agent.sock`
 2. Agent identifies the calling process via PID and queries the kubelet for pod metadata (namespace, service account, labels)
 3. Agent matches the discovered selectors against registration entries fetched from the Server
-4. Match found → Agent issues an X.509 SVID with `spiffe://agentcube.local/router` as URI SAN
+4. Match found → Agent issues an X.509 SVID with `spiffe://agentcube.local/ns/agentcube-system/sa/agentcube-router` as URI SAN
 5. Router receives a TLS certificate, private key, and trust bundle - ready to serve and initiate mTLS
-
-<!-- PLACEHOLDER: Workload Attestation Flow Diagram -->
-<!-- Diagram from https://spiffe.io/docs/latest/spire-about/spire-concepts/#workload-attestation
-     showing: workload calls Workload API → agent queries kernel/kubelet →
-     agent matches selectors → agent returns SVID -->
 
 ### 1.6 mTLS Integration
 
-Each component needs two changes: configure its server to require client certificates, and configure its clients to present its own certificate while verifying the server's. The TLS configuration is sourced differently depending on the certificate mode.
+Each component needs two changes: configure its server to require client certificates, and configure its clients to present its own certificate while verifying the server's. The mTLS implementation uses Go's standard `crypto/tls` package — no external libraries are required.
 
-#### SPIRE Mode
-
-The `go-spiffe` library (`github.com/spiffe/go-spiffe/v2`) provides helpers that make SPIRE-based mTLS integration straightforward:
-
-**Router** (outbound client only — serves external clients over HTTPS via Keycloak, does not need an mTLS server):
+Regardless of the certificate source (SPIRE or file-based), the cert/key/CA files end up on disk and the Go code is identical. When using SPIRE, the [SPIFFE Helper](https://github.com/spiffe/spiffe-helper) sidecar writes SVIDs to disk as PEM files, making them consumable by standard TLS configuration.
 
 ```go
-source, err := workloadapi.NewX509Source(ctx,
-    workloadapi.WithClientOptions(
-        workloadapi.WithAddr("unix:///run/spire/sockets/agent.sock"),
-    ),
-)
-
-// Client config for calling WorkloadManager
-wmTLSConfig := tlsconfig.MTLSClientConfig(source, source,
-    tlsconfig.AuthorizeID(spiffeid.RequireIDFromString("spiffe://agentcube.local/workload-manager")),
-)
-
-// Client config for proxying to PicoD
-picodTLSConfig := tlsconfig.MTLSClientConfig(source, source,
-    tlsconfig.AuthorizeID(spiffeid.RequireIDFromString("spiffe://agentcube.local/picod")),
-)
-```
-
-**WorkloadManager** (server-side, accepts calls only from Router):
-
-```go
-tlsConfig := tlsconfig.MTLSServerConfig(source, source,
-    tlsconfig.AuthorizeID(spiffeid.RequireIDFromString("spiffe://agentcube.local/router")),
-)
-```
-
-**PicoD** (server-side, accepts invocations proxied from Router):
-
-```go
-tlsConfig := tlsconfig.MTLSServerConfig(source, source,
-    tlsconfig.AuthorizeID(spiffeid.RequireIDFromString("spiffe://agentcube.local/router")),
-)
-```
-
-#### File-Based Mode
-
-When `--tls-cert-source=file` is configured, each component loads its certificate, private key, and CA bundle from disk:
-
-```go
-// Load certificate and key
+// Load certificate and key (from any source: SPIRE via spiffe-helper, cert-manager, self-signed, etc.)
 cert, err := tls.LoadX509KeyPair(certFile, keyFile)
 
 // Load CA bundle for peer verification
 caCert, err := os.ReadFile(caFile)
 caPool := x509.NewCertPool()
 caPool.AppendCertsFromPEM(caCert)
+```
 
-// Server TLS config (WorkloadManager, PicoD)
+#### Server-Side (WorkloadManager, PicoD)
+
+WorkloadManager and PicoD require incoming connections to present a valid client certificate signed by the trusted CA:
+
+```go
 serverTLSConfig := &tls.Config{
     Certificates: []tls.Certificate{cert},
     ClientCAs:    caPool,
     ClientAuth:   tls.RequireAndVerifyClientCert,
 }
 
-// Client TLS config (Router)
+server := &http.Server{
+    TLSConfig: serverTLSConfig,
+}
+```
+
+#### Client-Side (Router)
+
+The Router presents its own certificate when connecting to WorkloadManager and PicoD, and verifies the server's certificate against the trusted CA:
+
+```go
 clientTLSConfig := &tls.Config{
     Certificates: []tls.Certificate{cert},
     RootCAs:      caPool,
 }
+
+httpClient := &http.Client{
+    Transport: &http.Transport{
+        TLSClientConfig: clientTLSConfig,
+    },
+}
 ```
 
 The cert/key/CA files can be provisioned by any mechanism:
+- **SPIRE:** The [SPIFFE Helper](https://github.com/spiffe/spiffe-helper) sidecar fetches SVIDs from the Workload API and writes them to disk as PEM files. Certificates are rotated automatically.
 - **cert-manager:** Automatically issues and rotates certificates, stored as Kubernetes Secrets and mounted into pods.
 - **Self-signed CA:** Generated manually or via a script for development and testing environments.
 - **Let's Encrypt / corporate PKI:** Externally issued certificates placed on disk or in Secrets.
 
-*(Note: WorkloadManager manages PicoD pods via the Kubernetes API, so it does not make direct HTTP requests to PicoD and does not need a PicoD client mTLS configuration in either mode.)*
+*(Note: WorkloadManager manages PicoD pods via the Kubernetes API, so it does not make direct HTTP requests to PicoD and does not need a PicoD client mTLS configuration.)*
 
-In both modes, the mTLS enforcement is identical — this replaces the existing `PICOD_AUTH_PUBLIC_KEY` environment variable mechanism entirely.
+This replaces the existing `PICOD_AUTH_PUBLIC_KEY` environment variable mechanism entirely.
 
 ### 1.7 Communication Channel Summary
 
@@ -380,10 +405,10 @@ graph LR
         end
 
         subgraph "AgentCube Components"
-            R["Router<br/>spiffe://agentcube.local/router"]
-            WM["WorkloadManager<br/>spiffe://agentcube.local/workload-manager"]
-            P1["PicoD Sandbox 1<br/>spiffe://agentcube.local/picod"]
-            P2["PicoD Sandbox 2<br/>spiffe://agentcube.local/picod"]
+            R["Router<br/>spiffe://agentcube.local/ns/agentcube-system/sa/agentcube-router"]
+            WM["WorkloadManager<br/>spiffe://agentcube.local/ns/agentcube-system/sa/agentcube-workload-manager"]
+            P1["PicoD Sandbox 1<br/>spiffe://agentcube.local/sa/agentcube-sandbox"]
+            P2["PicoD Sandbox 2<br/>spiffe://agentcube.local/sa/agentcube-sandbox"]
         end
     end
 
--- a/docs/design/auth-proposal.md
+++ b/docs/design/auth-proposal.md
@@ -22,7 +22,7 @@ These gaps amount to three distinct problems:
 
 - **Internal (machine-to-machine):** Components trust each other implicitly based on network reachability. A compromised or rogue pod on the same network can impersonate any component.
 - **External (user-to-platform):** Anyone who can reach the Router endpoint can invoke sandboxes, with no identity verification and no audit trail.
-- **Authorization:** Even where authentication exists, there is no mechanism to control what an authenticated identity is allowed to do. There are no roles, no permission checks, and no namespace-scoped access control.
+- **Authorization & Isolation:** Even where authentication exists, there is no mechanism to control what an authenticated identity is allowed to do. There are no roles, no permission checks, no namespace-scoped access control, and no resource-level tenant isolation (preventing User A from invoking User B's sandbox).
 
 This proposal addresses all three problems using CNCF industry-standard tooling.
 
@@ -31,6 +31,7 @@ This proposal addresses all three problems using CNCF industry-standard tooling.
 - Establish zero-trust, mutually authenticated communication between all AgentCube internal components (Router, WorkloadManager, PicoD) using X.509 mTLS.
 - Provide external client/SDK authentication at the Router level via an industry-standard identity provider.
 - Implement role-based access control (RBAC) for external users using Keycloak's built-in authorization capabilities.
+- Enforce resource-level tenant isolation to guarantee users can only access sandboxes they created or that are shared with their group.
 - Keep all new auth features opt-in behind configuration flags so existing deployments are unaffected.
 - Minimize per-request latency overhead from authentication and authorization.
 - Supersede the existing PicoD-Plain-Authentication key distribution mechanism with automated certificate lifecycle management.
@@ -76,19 +77,20 @@ Internal communication between AgentCube components is secured using mutual TLS 
 Configuration flags for each component:
 
 ```
---tls-cert-source=spire|file    (default: spire)
---tls-cert-file=<path>          (for file mode)
---tls-key-file=<path>           (for file mode)
---tls-ca-file=<path>            (for file mode)
+--mtls-cert-file=<path>          
+--mtls-key-file=<path>           
+--mtls-ca-file=<path>            
 ```
 
-When `--tls-cert-source=file` is used, the cert/key/CA files can be populated by any mechanism - for example, a Kubernetes Secret mounted as a volume (managed by cert-manager), or static files for development. The mTLS enforcement (requiring client certs, verifying peer identity) is identical in both modes.
+The application simply loads the certificates directly from the paths provided via the `mtls-*-file` CLI flags.
+
+These cert/key/CA files can be populated on disk by any mechanism - for example, the SPIFFE Helper sidecar syncing SVIDs to disk, a Kubernetes Secret mounted as a volume (managed by cert-manager), or static files for development. The mTLS enforcement (requiring client certs, verifying peer identity) is identical in all cases.
 
 ### 1.1 SPIRE Background
 
 [SPIFFE](https://spiffe.io/) (Secure Production Identity Framework for Everyone) is a CNCF graduated project that provides a standard for service identity. It defines:
 
-- **SPIFFE ID:** A URI-formatted identity, e.g., `spiffe://agentcube.local/ns/agentcube-system/sa/agentcube-router`
+- **SPIFFE ID:** A URI-formatted identity, e.g., `spiffe://cluster.local/ns/agentcube-system/sa/agentcube-router`
 - **SVID (SPIFFE Verifiable Identity Document):** An X.509 certificate or JWT that proves a workload holds a given SPIFFE ID.
 
 [SPIRE](https://spiffe.io/docs/latest/spire-about/spire-concepts/) is the production implementation of SPIFFE. It has two components:
@@ -111,7 +113,7 @@ This design uses a single-cluster SPIRE deployment with one SPIRE Server and a s
 All AgentCube components operate under a single trust domain:
 
 ```
-spiffe://agentcube.local
+spiffe://cluster.local
 ```
 
 Each component receives a unique SPIFFE ID following the Istio-convention format `spiffe://<trust-domain>/ns/<namespace>/sa/<service-account>`, which encodes the workload's Kubernetes namespace and service account directly in the URI path.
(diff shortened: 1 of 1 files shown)


docs/design/auth-proposal.md
```diff
@@ -0,0 +1,608 @@
...
+- **Disabled:** Router behaves exactly as today - no authentication required.
+- **Enabled:** All invocation endpoints require `Authorization: Bearer <token>`. Health checks (`/health/live`, `/health/ready`) remain unauthenticated.
+
+---
+
+## 3. Authorization (Keycloak RBAC)
+
+### 3.1 Overview
+
+Authentication verifies *who* the user is. Authorization determines *what* that user is allowed to do. This proposal uses Keycloak's built-in Authorization Services for role-based access control rather than building a custom policy engine.
```
hzxuzhonghu: The proposal says P3 uses Keycloak Authorization Services, but the implementation below only checks `realm_access.roles` locally from the JWT. Those are different architectures: simple role-claim checks do not provide Keycloak AuthZ resource decisions or namespace-scoped permissions unless extra claims projection/token exchange is defined. It would help to pick one approach explicitly here so the design does not overstate what the Router can enforce from a validated access token alone.

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,608 @@
...
+spiffe://agentcube.local
+```
+
+Each component receives a unique SPIFFE ID based on its Kubernetes metadata:
+
+| Component | SPIFFE ID | Selectors |
+|---|---|---|
+| Router | `spiffe://agentcube.local/router` | `k8s:ns:agentcube-system`, `k8s:sa:agentcube-router` |
+| WorkloadManager | `spiffe://agentcube.local/workload-manager` | `k8s:ns:agentcube-system`, `k8s:sa:agentcube-workload-manager` |
+| PicoD (Sandboxes) | `spiffe://agentcube.local/picod` | `k8s:ns:agentcube-system`, `k8s:pod-label:app:picod` |
```
hzxuzhonghu: This selector looks too strict for the current runtime model. PicoD sandboxes are created in the CodeInterpreter/runtime namespace, not a fixed `agentcube-system` namespace, so a SPIRE registration keyed to `k8s:ns:agentcube-system` would fail to attest sandboxes in user namespaces. The design probably needs either namespace-agnostic selectors or one registration pattern per workload namespace.

hzxuzhonghu (commented):
Posting the main design consistency issues I found in the auth proposal.

docs/design/auth-proposal.md
mahil-2040: you're right, the proposal was conflating two different things. Keycloak Authorization Services implies server-side policy evaluation with RPTs, but actual implementation just reads realm_access.roles from the JWT locally. that was my mistake.

I've updated the proposal to describe it as realm role checking from the JWT, and removed references to Authorization Services, and clarified that namespace scoping would use protocol mappers for custom claims rather than Keycloak's resource-based permissions.

docs/design/auth-proposal.md
mahil-2040: you're right, sandbox pods are created in the namespace from the user's request, not in agentcube-system. I've updated the PicoD SPIRE registration to use only k8s:pod-label:app:picod without a namespace constraint, so it can attest PicoD pods regardless of which namespace they're in.

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,607 @@
...
+    KC-->>SDK: Access Token (JWT) + Refresh Token
+
+    Note over SDK, Sandbox: Invocation
+    SDK->>Router: POST /v1/namespaces/.../invocations<br/>Authorization: Bearer jwt
+    Router->>Router: Validate JWT signature via cached JWKS
+    Router->>Router: Extract claims (sub, roles)
+    Router->>Router: Check role authorization
+
+    alt Valid token + authorized role
+        Router->>WM: Forward (mTLS)
```
hzxuzhonghu: This changes the Router -> WorkloadManager hop to workload-level mTLS, but the proposal does not say how the authenticated end-user identity survives that hop. Today WorkloadManager authorization and user-scoped Kubernetes client construction are keyed off the caller's bearer token / service account. If Router becomes the only authenticated caller here, all sandbox creates become attributable to Router identity unless some user context is propagated separately. Please make the intended model explicit: either WorkloadManager remains user-aware (and describe how identity/audit context is forwarded), or Router becomes the sole control-plane principal and ownership/audit semantics move up to Router.

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,607 @@
...
+# Pre-obtained token
+client = CodeInterpreterClient(
+    auth=TokenAuth(access_token="<keycloak-jwt>")
+)
+
+# Usage unchanged - auth is transparent
+result = client.run_code("python", "print('hello')")
+```
+
+The SDK handles token refresh automatically using the refresh token.
```
hzxuzhonghu: The SDK section mixes two different token lifecycles. `authorization_code` flows can use refresh tokens, but `client_credentials` typically does not issue a refresh token. For the service-account / automation case above, the SDK would normally obtain a new access token by repeating the client-credentials exchange rather than refreshing one. It would help to split those behaviors explicitly so the doc does not overstate refresh-token support for the confidential-client flow.

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,607 @@
...
+
+## Future Enhancements
+
+### OPA for Authorization
+
+The standard across CNCF projects is to strictly separate authentication and authorization. Projects like Volcano, Istio, and ArgoCD follow the pattern of using tools like Keycloak for authentication and Open Policy Agent (OPA) for authorization. If AgentCube's authorization needs grow beyond what Keycloak RBAC offers, OPA could replace Keycloak's authorization role, keeping Keycloak focused purely on identity and token issuance.
+
+The key advantages of OPA over Keycloak-based RBAC:
+
+- **Local evaluation:** OPA runs as a sidecar alongside the Router, so policy checks are local and avoid a network call to Keycloak on every request.
```
hzxuzhonghu: This benefit is overstated as written. The RBAC design in Section 3 already performs local JWT validation plus local role-claim checks, so it also avoids a per-request round trip to Keycloak. OPA's advantage here is richer, decoupled local policy evaluation rather than lower hot-path network overhead.

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,607 @@
...
+
+Each component receives a unique SPIFFE ID based on its Kubernetes metadata:
+
+| Component | SPIFFE ID | Selectors |
+|---|---|---|
+| Router | `spiffe://agentcube.local/router` | `k8s:ns:agentcube-system`, `k8s:sa:agentcube-router` |
+| WorkloadManager | `spiffe://agentcube.local/workload-manager` | `k8s:ns:agentcube-system`, `k8s:sa:agentcube-workload-manager` |
+| PicoD (Sandboxes) | `spiffe://agentcube.local/picod` | `k8s:pod-label:app:picod` |
+
+> **Note:** PicoD uses only a pod-label selector without a namespace constraint because sandbox pods are created in the namespace specified by the user's request, not in a fixed namespace.
```
hzxuzhonghu: Using only `k8s:pod-label:app:picod` to mint `spiffe://agentcube.local/picod` is risky in a multi-tenant cluster: any workload that can create a pod with that label can potentially receive the PicoD SPIFFE ID and impersonate a sandbox runtime. The design should document compensating controls here, such as privileged admission policy for that label, dedicated sandbox service accounts, or tighter selectors where deployers can constrain sandbox namespaces.

hzxuzhonghu (commented):
Posted a few design-consistency comments on the current PR head. The main issues are the missing explanation of how end-user identity/audit context survives the Router -> WorkloadManager hop once that channel becomes pure workload mTLS, and the security implications of using a label-only SPIRE selector for PicoD. I also left two smaller documentation comments around OAuth token refresh behavior and the OPA comparison wording.

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,607 @@
...
+
+- **SPIFFE ID:** A URI-formatted identity, e.g., `spiffe://agentcube.local/router`
+- **SVID (SPIFFE Verifiable Identity Document):** An X.509 certificate or JWT that proves a workload holds a given SPIFFE ID.
+
+[SPIRE](https://spiffe.io/docs/latest/spire-about/spire-concepts/) is the production implementation of SPIFFE. It has two components:
+
+- **SPIRE Server:** Central signing authority. Manages registration entries (which selectors map to which SPIFFE IDs) and issues SVIDs to agents.
+- **SPIRE Agent:** Runs on every node (DaemonSet). Performs workload attestation, verifying process identity by querying the kernel and kubelet, and delivers SVIDs to local workloads via a Unix domain socket (the Workload API).
+
+SPIRE handles the entire certificate lifecycle (issuance, rotation, revocation) automatically. Workloads never see raw keys on disk; certificates arrive through the Workload API and are rotated before they expire.
```
hzxuzhonghu: we also need another cert provision way without relying on spire, because spire can be not used by users.
In other word, we should support any x509 certs, whether self signed, or signed by other mechnism, certmanager or by let's encrypt

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,607 @@
...
+
+```
+spiffe://agentcube.local
+```
+
+Each component receives a unique SPIFFE ID based on its Kubernetes metadata:
+
+| Component | SPIFFE ID | Selectors |
+|---|---|---|
+| Router | `spiffe://agentcube.local/router` | `k8s:ns:agentcube-system`, `k8s:sa:agentcube-router` |
```
hzxuzhonghu: Suggest this spiffe id fomat: spiffe://<cluster domain>/ns/<namespace>/sa/<service account>

hzxuzhonghu: what does selector mean

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,607 @@
...
+            name: spire-agent-config
+```
+
+### 1.5 Workload Registration
+
+Registration entries are created via the SPIRE Server CLI:
+
+```bash
+# Register the Router
+spire-server entry create \
```
hzxuzhonghu: does these cmd run in the pod of spire server manually?

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,607 @@
...
+5. Router receives a TLS certificate, private key, and trust bundle - ready to serve and initiate mTLS
+
+<!-- PLACEHOLDER: Workload Attestation Flow Diagram -->
+<!-- Diagram from https://spiffe.io/docs/latest/spire-about/spire-concepts/#workload-attestation
+     showing: workload calls Workload API → agent queries kernel/kubelet →
+     agent matches selectors → agent returns SVID -->
+
+### 1.6 mTLS Integration
+
+The `go-spiffe` library (`github.com/spiffe/go-spiffe/v2`) provides helpers that make mTLS integration straightforward. Each component needs two changes: configure its server to require client certificates, and configure its clients to present its own certificate while verifying the server's.
```
hzxuzhonghu: THis is not necessary, standard go tls package can also implement this

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,607 @@
...
+Router → PicoD:            HTTP + custom JWT (PicoD-Plain-Auth)
+WorkloadManager → PicoD:   (No direct HTTP calls, managed via K8s API)
+```
+
+**After:**
+
+```
+SDK → Router:              HTTPS + Keycloak JWT (see Section 2)
+Router → WorkloadManager:  mTLS (SPIRE SVIDs)
+Router → PicoD:            mTLS (SPIRE SVIDs)
```
hzxuzhonghu: From above section, does all the picod share same spiffe id? 

Previsouly we donot use tls between router and picod, one consideration is latency.
The other concern here is we also allow user build their customized code interperet.

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,607 @@
...
+```mermaid
+sequenceDiagram
+    participant SDK as AgentCube SDK
+    participant KC as Keycloak
+    participant Router as Router
+    participant WM as WorkloadManager
+    participant Sandbox as PicoD
+
+    Note over SDK, KC: Authentication
+    SDK->>KC: POST /realms/agentcube/protocol/openid-connect/token
```
hzxuzhonghu: How does the keycloack auth the caller?

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,607 @@
...
+    KC-->>SDK: Access Token (JWT) + Refresh Token
+
+    Note over SDK, Sandbox: Invocation
+    SDK->>Router: POST /v1/namespaces/.../invocations<br/>Authorization: Bearer jwt
+    Router->>Router: Validate JWT signature via cached JWKS
```
hzxuzhonghu: where is cached jwks get from? Suppose it is the ketcloak, should document it

docs/design/auth-proposal.md
hzxuzhonghu: do we need to bring user identity or just the router's identity here? We can discuss

mahil-2040: @hzxuzhonghu  I agree this is a gap in the current proposal. 

I think we should go with this option: WorkloadManager remains user-aware via forwarded identity context. Here's how it works:

With Keycloak, external users are Keycloak identities (sub: user-123), not Kubernetes ServiceAccounts - so the current pattern of forwarding a K8s SA token and constructing a user-scoped K8s client no longer applies. Instead the workflow should be like this:

```
SDK → Router:
  1. Authenticates via Keycloak JWT (validated locally using cached JWKS)
  2. Extracts user claims (sub, realm_access.roles)
  3. Checks role authorization (e.g., sandbox:invoke)

Router → WorkloadManager (over mTLS):
  4. Forwards user context as trusted headers:
     X-AgentCube-User-ID: user-123
     X-AgentCube-User-Roles: sandbox:invoke

WorkloadManager:
  5. Verifies caller is Router via mTLS (spiffe://agentcube.local/router)
  6. Trusts X-AgentCube-User-* headers (guaranteed authentic by mTLS — only Router can send them)
  7. Uses X-AgentCube-User-ID for sandbox ownership tagging and audit logging
  8. Uses its OWN ServiceAccount for all Kubernetes API calls (not the end user's)
  9. Does NOT re-evaluate authorization (Router already enforced it)
```

So tell me if this is ok . i have updated the proposal by adding a dedicated "User Identity Propagation" section to make the model explicit or if you have any other suggestion then we can discuss that as well. just let me know your thoughts.

docs/design/auth-proposal.md
mahil-2040: Regarding custom code interpreters, could you clarify the concern here?

Is the concern about the added complexity for users building custom interpreters? because with mTLS enforced on Router→PicoD, custom interpreters would need to handle TLS certificates. 

And can you suggest a approach for supporting custom interpreters alongside mTLS?

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,718 @@
...
+### 1.4 SPIRE Deployment
+
+#### SPIRE Server
+
+Deployed as a `StatefulSet` in `agentcube-system`. Key configuration:
+
+```hcl
+server {
+    bind_address = "0.0.0.0"
+    bind_port    = "8081"
```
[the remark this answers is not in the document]
mahil-2040: In the official SPIRE documentation it is mentioned that "bind_port" should be string (https://spiffe.io/docs/latest/deploying/configuring/)

mahil-2040:
@hzxuzhonghu PTAL, i have made the changes according to your feedback.

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,724 @@
...
+1. **Workload Manager Auth** (`pkg/workloadmanager/auth.go`): Optional Kubernetes TokenReview-based ServiceAccount token validation, gated behind `config.EnableAuth`, plus per-sandbox ownership checks using the extracted user identity (effectively relying on Kubernetes RBAC when using the user-scoped client).
+2. **Router → PicoD Auth** (`PicoD-Plain-Authentication-Design`): A custom RSA key-pair scheme where the Router signs JWTs and PicoD verifies them using a public key exposed via the `PICOD_AUTH_PUBLIC_KEY` environment variable. The key pair (`private.pem`, `public.pem`) is stored in the `picod-router-identity` Secret, and the WorkloadManager reads this Secret to inject the public key into PicoD pods. This works for the Router→PicoD channel but leaves other internal channels unauthenticated.
+3. **Router → WorkloadManager**: Optional, one-sided authentication. `pkg/router/session_manager.go` can attach a `Authorization: Bearer <serviceaccount token>` header, and WorkloadManager can validate it when `--enable-auth` is enabled. This is not mutual workload identity or a zero-trust model, and when auth is disabled any pod on the cluster network can call the WorkloadManager API.
```
hzxuzhonghu, line 18: We should consider also bring the user's identity in the token

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,724 @@
...
+4. **External Clients → Router**: No authentication. The `handleInvoke` handler in `pkg/router/handlers.go` processes incoming requests without verifying the caller's identity.
+
+These gaps amount to three distinct problems:
+
+- **Internal (machine-to-machine):** Components trust each other implicitly based on network reachability. A compromised or rogue pod on the same network can impersonate any component.
+- **External (user-to-platform):** Anyone who can reach the Router endpoint can invoke sandboxes, with no identity verification and no audit trail.
```
hzxuzhonghu, line 24: Can limit only the creator or the group can access a sandbox

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,724 @@
...
+---
+
+## 1. Internal Workload Authentication (X.509 mTLS)
+
+Internal communication between AgentCube components is secured using mutual TLS (mTLS) with X.509 certificates. The mTLS enforcement layer is **certificate-source agnostic** - it works with any valid X.509 cert/key/CA bundle, regardless of how the certificates are provisioned. Two certificate source modes are supported:
+
+| Mode | Certificate Source | Rotation | Best For |
+|---|---|---|---|
+| **SPIRE** (recommended) | SPIRE Workload API issues short-lived SVIDs automatically | Automatic (default: 1 hour TTL) | Production deployments needing zero-touch certificate management |
+| **File-based** | Certs loaded from disk (provisioned by cert-manager, self-signed CA, Let's Encrypt, corporate PKI, etc.) | Manual or delegated to the provisioning tool | Environments where SPIRE is not available, or operators prefer existing PKI infrastructure |
```
hzxuzhonghu, line 75: We can use spiffee as cert san like istio `spiffe://cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account`

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,724 @@
...
+
+Configuration flags for each component:
+
+```
+--tls-cert-source=spire|file    (default: spire)
```
hzxuzhonghu: how about removing this flag, it should load the cert below if provided and exist.

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,724 @@
...
+- **SVID (SPIFFE Verifiable Identity Document):** An X.509 certificate or JWT that proves a workload holds a given SPIFFE ID.
+
+[SPIRE](https://spiffe.io/docs/latest/spire-about/spire-concepts/) is the production implementation of SPIFFE. It has two components:
+
+- **SPIRE Server:** Central signing authority. Manages registration entries (which selectors map to which SPIFFE IDs) and issues SVIDs to agents.
+- **SPIRE Agent:** Runs on every node (DaemonSet). Performs workload attestation, verifying process identity by querying the kernel and kubelet, and delivers SVIDs to local workloads via a Unix domain socket (the Workload API).
+
+SPIRE handles the entire certificate lifecycle (issuance, rotation, revocation) automatically. Workloads receive certificates through the Workload API and they are rotated before they expire.
+
+### 1.2 Why Single-Cluster SPIRE
```
hzxuzhonghu, line 103: make sense to implement single custer first

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,724 @@
...
+| Priority | Layer | Problem | Solution |
+|---|---|---|---|
+| P1 (Urgent) | Internal workload identity | Machine-to-machine trust between Router, WorkloadManager, PicoD | X.509 mTLS (SPIRE recommended, file-based certs also supported) |
+| P2 | External user authentication | Client/SDK identity verification at the Router | Keycloak (OIDC/OAuth2) |
+| P3 | Authorization | Role-based access control for external users | Keycloak realm roles (JWT claim checking) |
+| P4 (Stretch) | Cloud provider federation | Enterprise SSO via cloud IAM | Keycloak identity brokering |
+
+---
+
+## 1. Internal Workload Authentication (X.509 mTLS)
```
hzxuzhonghu, line 68: Reading through the whole section 1, i find it is only talking about x509 certs with spire. Seems not align with the desc below

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,724 @@
...
+
+All AgentCube components operate under a single trust domain:
+
+```
+spiffe://agentcube.local
+```
+
+Each component receives a unique SPIFFE ID following the Istio-convention format `spiffe://<trust-domain>/ns/<namespace>/sa/<service-account>`, which encodes the workload's Kubernetes namespace and service account directly in the URI path.
+
+SPIRE uses **selectors** to verify workload identity during attestation. When a workload requests its certificate, the SPIRE Agent inspects the workload's Kubernetes metadata (namespace, service account, labels) and matches it against registered selectors. A workload only receives its SVID if all registered selectors match.
```
hzxuzhonghu, line 121: Is this the default using guide of spire?

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,724 @@
...
+
+| Component | SPIFFE ID | Selectors |
+|---|---|---|
+| Router | `spiffe://agentcube.local/ns/agentcube-system/sa/agentcube-router` | `k8s:ns:agentcube-system`, `k8s:sa:agentcube-router` |
```
hzxuzhonghu:
```suggestion
| Router | `spiffe://cluster.local/ns/agentcube-system/sa/agentcube-router` | `k8s:ns:agentcube-system`, `k8s:sa:agentcube-router` |
```

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,724 @@
...
+| WorkloadManager | `spiffe://agentcube.local/ns/agentcube-system/sa/workloadmanager` | `k8s:ns:agentcube-system`, `k8s:sa:workloadmanager` |
+| PicoD (Sandboxes) | `spiffe://agentcube.local/sa/agentcube-sandbox` | `k8s:pod-label:app:picod`, `k8s:sa:agentcube-sandbox` |
```
hzxuzhonghu: same

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,724 @@
...
+
+> **Note:** PicoD cannot use a namespace selector because sandbox pods are created in the namespace specified by the user's request, not in a fixed namespace. To prevent identity spoofing in multi-tenant clusters, PicoD registration requires both the `app:picod` pod label and a dedicated `agentcube-sandbox` ServiceAccount. WorkloadManager creates this ServiceAccount in the target namespace as part of sandbox provisioning. Using a specific ServiceAccount requires RBAC permission, preventing arbitrary workloads from obtaining the PicoD SPIFFE ID.
```
hzxuzhonghu, line 129: >WorkloadManager creates this ServiceAccount in the target namespace as part of sandbox provisioning.

Can you elaborate on this?

docs/design/auth-proposal.md
mahil-2040: So instead of having the Router extract the user identity and pass it blindly via X-AgentCube-User-ID headers to the WorkloadManager, the Router should perform an OAuth 2.0 Token Exchange (RFC 8693) with Keycloak.

The Router would exchange the incoming user token for a new downstream token (where the subject is the user, and the actor claim is the Router) and pass that nested token to the WorkloadManager.

is my understanding correct?

If yes, then Should I update the proposal to replace the custom HTTP header approach with this Token Exchange flow as the core requirement, or should we keep the header approach for now and mark Token Exchange as a  Stretch goal in the proposal?

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,724 @@
...
+>
+> **Security Hardening:** For multi-tenant clusters, operators are encouraged to deploy a `ValidatingAdmissionPolicy` (or equivalent policy engine rule) that restricts the `app: picod` label to pods created by the WorkloadManager ServiceAccount. This provides defense-in-depth beyond the ServiceAccount selector.
+
+### 1.4 SPIRE Deployment
+
+#### SPIRE Server
+
+Deployed as a `StatefulSet` in `agentcube-system`. The StatefulSet runs two critical containers: the SPIRE Server itself, and the SPIRE Controller Manager sidecar (which shares the Server's Unix Domain Socket via an `emptyDir` volume). Key SPIRE Server configuration:
```
hzxuzhonghu: no need to mention much of how to install spire server

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,724 @@
...
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: spire-agent
+  namespace: agentcube-system
+spec:
+  selector:
+    matchLabels:
+      app: spire-agent
+  template:
```
hzxuzhonghu: ditto, hide the details, has very little to do with you design

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,724 @@
...
+            path: /run/spire/sockets
+            type: DirectoryOrCreate
+        - name: spire-config
+          configMap:
+            name: spire-agent-config
+```
+
+#### SPIRE Controller Manager
+
+The [SPIRE Controller Manager](https://github.com/spiffe/spire-controller-manager) watches for `ClusterSPIFFEID` custom resources and automatically syncs registration entries to the SPIRE Server. This makes workload registration fully declarative and is the recommended approach for production deployments.
```
hzxuzhonghu, line 149: donot you mention self register by agent it self? So with controller, what is the expected way to register

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,724 @@
...
+    matchLabels:
+      kubernetes.io/metadata.name: agentcube-system
+---
+# PicoD registration (namespace-agnostic)
+apiVersion: spire.spiffe.io/v1alpha1
+kind: ClusterSPIFFEID
+metadata:
+  name: agentcube-sandbox
+spec:
+  spiffeIDTemplate: "spiffe://agentcube.local/sa/{{ .PodSpec.ServiceAccountName }}"
```
hzxuzhonghu: so by ignoring namespace here, does it mean all picod across namespaces would receive svid? 

And if yes, are they same?

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,724 @@
...
+  spire-server entry create \
+    -spiffeID spiffe://agentcube.local/sa/agentcube-sandbox \
+    -parentID spiffe://agentcube.local/spire-agent \
+    -selector k8s:pod-label:app:picod \
+    -selector k8s:sa:agentcube-sandbox
+```
+
+> **Note:** Manual CLI entries are stored in the SPIRE Server's datastore (SQLite by default). They survive restarts if the datastore is backed by persistent storage, but are harder to manage and audit compared to the declarative CRD approach.
+
+#### Attestation Flow (Example: Router)
```
hzxuzhonghu, line 240: I am concerned the delay here, in agencube its low latency on sandbox provision should be highly prioritized. Especially for agentic rl scenario.

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,724 @@
...
+        TLSClientConfig: clientTLSConfig,
+    },
+}
+```
+
+The cert/key/CA files can be provisioned by any mechanism:
+- **SPIRE:** The [SPIFFE Helper](https://github.com/spiffe/spiffe-helper) sidecar fetches SVIDs from the Workload API and writes them to disk as PEM files. Certificates are rotated automatically.
+- **cert-manager:** Automatically issues and rotates certificates, stored as Kubernetes Secrets and mounted into pods.
+- **Self-signed CA:** Generated manually or via a script for development and testing environments.
+- **Let's Encrypt / corporate PKI:** Externally issued certificates placed on disk or in Secrets.
```
hzxuzhonghu, line 407: would you implement this way?

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,724 @@
...
+
+> **Note:** The architecture diagram above shows the SPIRE-based deployment. When using file-based certificates (`--tls-cert-source=file`), the SPIRE Infrastructure components are not deployed; certificates are provisioned externally and mounted into pods. The mTLS connections between AgentCube components remain the same.
+
+### 1.9 Impact on Existing PicoD-Plain-Authentication
+
+The X.509 mTLS approach supersedes the PicoD-Plain-Authentication design:
+
+| Current (PicoD-Plain-Auth) | New (X.509 mTLS) |
+|---|---|
+| Router generates RSA key pair, stores both keys in `picod-router-identity` Secret | SPIRE issues short-lived X.509 SVIDs automatically, or certs are loaded from disk (cert-manager, self-signed, etc.) |
```
hzxuzhonghu: so you should dynamiclly load certs in both sides go code

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,724 @@
...
+    else Invalid/expired token
+        Router-->>SDK: 401 Unauthorized
+    else Insufficient permissions
+        Router-->>SDK: 403 Forbidden
+    end
+```
+
+### 2.3 Keycloak Deployment
+
+Keycloak is deployed as a `Deployment` in `agentcube-system`. A dedicated realm called `agentcube` is created during installation containing:
```
hzxuzhonghu: ` A dedicated realm called `agentcube``, I would suggest letting users register to support multi tenants

hzxuzhonghu (commented):
This proposal not cover router -> agent runtime authentication. how could we delegate user's identity to the agent?

docs/design/auth-proposal.md
mahil-2040: I’ve updated the proposal to make the tenant isolation explicit using Resource-Level Access Control (RLAC):

Creation: When a sandbox is created, the WorkloadManager tags all related resources (CRDs/Pods) with immutable labels like agentcube.io/owner: user-123, based on JWT claims (sub, groups).
Invocation: On each request, the Router checks the caller’s JWT against the sandbox’s labels. If they don’t match (not owner or allowed group), it returns 403 Forbidden.

This ensures strict tenant isolation , no user can access another user’s sandbox, even within the same namespace.

docs/design/auth-proposal.md
mahil-2040: We have defined Istio-like SPIFFE IDs in Section 1.3, but I realized the Go code snippets in Section 1.6 weren't explicitly checking them during the connection. I've just updated the proposal's tls.Config examples to include a VerifyPeerCertificate check that extracts and verifies the URI SAN for the peer's SPIFFE ID.

docs/design/auth-proposal.md
mahil-2040: I've updated the proposal to remove --tls-cert-source. Now, the system will use simple auto-detection: if the --tls-*-file flags are provided and the cert/key files exist on disk at startup, the component operates in file-based mode. Otherwise, it falls back to authenticating with the SPIRE Workload API directly.

docs/design/auth-proposal.md
mahil-2040: Changed agentcube.local to cluster.local

docs/design/auth-proposal.md
mahil-2040: I added a new section (1.6 File-Based Provisioning with cert-manager) that shows how to achieve the same zero-trust mTLS setup with File-based provisioning. It includes a sample Certificate CRD (mapping uris to SPIFFE IDs) and how the generated Secret plugs into our --mtls-*-file args.

docs/design/auth-proposal.md
mahil-2040: Yes, we are utilizing the native k8s:ns and k8s:sa selectors.

For context, this default integration is documented directly in the official repository:(https://github.com/spiffe/spire/blob/main/doc/plugin_agent_workloadattestor_k8s.md)

docs/design/auth-proposal.md
mahil-2040: To simplify sandbox provisioning across namespaces, the WorkloadManager ensures the required ServiceAccount exists at runtime:

Here is the specific flow during provisioning:

Check: When a request arrives to run a sandbox in namespace X, the WorkloadManager uses its Kubernetes client to GET the agentcube-sandbox ServiceAccount in namespace X.

Provision: If it does not exist, the WorkloadManager (using its own RBAC ClusterRole) issues a CREATE serviceaccount request to provision it on the fly within that target namespace.

Mount: The WorkloadManager then creates the PicoD Sandbox Pod and explicitly bounds it to serviceAccountName: agentcube-sandbox.

This avoids requiring users to pre-provision ServiceAccounts.

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,758 @@
...
+
+### 3.1 Overview
+
+Authentication verifies *who* the user is. Authorization determines *what* that user is allowed to do. This proposal uses Keycloak's realm roles for role-based access control. Keycloak embeds the user's assigned roles directly into the JWT access token, and the Router checks these roles locally from the validated token - no additional call to Keycloak is needed at request time.
+
+### 3.2 Role Hierarchy
+
+Keycloak realm roles are organized in a simple hierarchy:
+
+| Role | Permissions | Inherits |
```
YaoZengzeng, line 669: what does `Inherits` means? Like `sandbox:manage` include permissions of `sandbox:invoke` plus its own permissions?

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,758 @@
...
+|---|---|---|
+| `sandbox:invoke` | Invoke agent runtimes, invoke code interpreters, list sessions | - |
+| `sandbox:manage` | Create/delete AgentRuntime and CodeInterpreter CRDs | `sandbox:invoke` |
```
YaoZengzeng, line 672: Seems we create `AgentRuntime` and `CodeInterpreter` by running kubectl manually, not from SDK.

docs/design/auth-proposal.md
hzxuzhonghu: got， but our api AgenRuntime also allows user specifying service account. Can you support this case?

docs/design/auth-proposal.md
hzxuzhonghu: This is a critical issue, maybe we need also jwt based SVID

docs/design/auth-proposal.md
hzxuzhonghu: Right, i think both `X-AgentCube-User-ID` and token exchange make sense

docs/design/auth-proposal.md
hzxuzhonghu: add some comment about the work flow

docs/design/auth-proposal.md
```diff
@@ -0,0 +1,758 @@
...
+
+### 1.5 Workload Registration
+
+Registration entries tell the SPIRE Server which workloads should receive which SPIFFE IDs. There are two approaches depending on the environment:
+
+#### Production: ClusterSPIFFEID CRDs (Recommended)
+
+With the SPIRE Controller Manager deployed, registration entries are defined as Kubernetes CRDs. These are managed declaratively alongside other AgentCube manifests and are automatically synced to the SPIRE Server:
+
+```yaml
```
hzxuzhonghu, line 170: Assume this is all installed beforehand, right?

docs/design/auth-proposal.md
mahil-2040: Yes, exactly! In Keycloak this uses "Composite Roles. Assigning a user the higher-level sandbox:manage role automatically grants them the sandbox:invoke permissions as well, preventing us from having to assign both roles to administrators.

docs/design/auth-proposal.md
mahil-2040: Not quite beforehand. In the spire-controller-manager we can bundle these ClusterSPIFFEID manifests directly into the main AgentCube Helm chart/release bundle.
When the admin runs `helm install agentcube`, the WorkloadManager, the Router, and these SPIFFE ID registrations are all applied to the cluster at the same time.

docs/design/auth-proposal.md
mahil-2040: You're right. currently `AgentRuntime` and `CodeInterpreter` CRDs are created via kubectl or the AgentCube CLI (agentcube publish), both of which go directly to the Kubernetes API and are governed by Kubernetes RBAC, not Keycloak.

The `sandbox:manage` role is defined as a forward-looking placeholder for when CRD lifecycle operations are eventually exposed through the Router. For now, only `sandbox:invoke` is actively enforced by the Keycloak middleware. I've added a clarifying note to the proposal to make this distinction explicit.

docs/design/auth-proposal.md
mahil-2040: Yes, the AgentRuntime API exposes a full corev1.PodSpec via podTemplate, which includes the ServiceAccountName field.

But When mTLS is enabled (`--enable-mtls`), the WorkloadManager will enforce serviceAccountName: `agentcube-sandbox` on all sandbox pods regardless of what the user specifies in their AgentRuntime spec. this is necessary because the SPIRE registration entry for PicoD is bound to the `agentcube-sandbox` ServiceAccount selector allowing arbitrary SAs would break the identity attestation chain.

When mTLS is disabled (the current default), the user-specified ServiceAccount is respected as-is, preserving full backward compatibility.

docs/design/auth-proposal.md
mahil-2040: Yes, to both questions.

1. Because sandboxes are provisioned dynamically into arbitrary, user-specified namespaces, we must omit the `namespaceSelector`. This allows the SPIRE Agent to attest PicoD pods globally across the cluster, provided they possess the required ServiceAccount and label.
2. Yes, every sandbox receives the exact same SPIFFE ID (`spiffe://cluster.local/sa/agentcube-sandbox`).

The shared SPIFFE ID purely provides machine-to-machine authentication (proving to the Router that the target is a legitimate, WorkloadManager provisioned PicoD pod).

The actual tenant isolation and authorization (ensuring User A cannot access User B's sandbox, even if they share the same machine identity) is strictly enforced at the Application Layer via the Resource-Level Access Control (RLAC) JWT claim checks detailed in Section 3.6. This prevents the SPIRE Server from being overwhelmed by creating/destroying unique cryptographic identities for every short-lived sandbox instance!

docs/design/auth-proposal.md
mahil-2040: we are not implementing any integration code for these four mechanisms inside the AgentCube components.

The core of this design is that the AgentCube Go binaries will only read standard certificate files from the local disk using the new `--mtls-cert-file`, `--mtls-key-file`, and `--mtls-ca-file` flags.

Because AgentCube only cares about reading a file, we automatically support all four of those mechanisms out-of-the-box, as they all ultimately just write PEM files to a Kubernetes volume mount. Our implementation will strictly focus on adding the file-reading logic to the Go codebase, and providing example Kubernetes YAML manifests (like the `spiffe-helper` sidecar) to show users how to mount those files seamlessly.

docs/design/auth-proposal.md
mahil-2040: You are right, as SPIRE SVIDs rotate so aggressively (e.g., 1-hour TTL), the Go applications cannot statically hold onto the tls.Certificate they read at startup.

I've updated the Go snippets in Section 1.7 of the proposal to explicitly demonstrate using the GetCertificate and GetClientCertificate callbacks. In the actual implementation, these functions will dynamically reload the PEM files from disk on demand, guaranteeing zero-downtime rotation without ever needing to restart the Router or PicoD pods!

docs/design/auth-proposal.md
mahil-2040: I see the confusion! You are totally right that the SPIRE Agent automatically self-registers to the SPIRE Server (Node Attestation).

However, the Controller Manager is responsible for Workload Registration, not Node Registration. The Agent registers the worker node, but the Controller Manager reads the `ClusterSPIFFEID` CRDs to register the rules for the pods (e.g., "Any pod with `app: picod` gets this SPIFFE ID").

To make this clear, I have updated the proposal under Section 1.4 to include a new "End-to-End Identity Workflow" section that explicitly breaks down how Agent self-registration, Controller workload registration, and SVID delivery all link together step-by-step!

docs/design/auth-proposal.md
mahil-2040: The phrasing in the proposal made it sound like the architecture was hardcoded to a single realm name - that was definitely my mistake!

But the Go implementation actually accepts the Keycloak Issuer URL and Realm dynamically through CLI configuration flags (e.g., `--keycloak-realm`). The agentcube realm is purely just the default name created by the "getting started" Helm chart. Cluster administrators can easily alter the deployment flags to point the Router at whatever custom or multiple distinct realms they need to achieve full multi-tenant isolation.

I have updated the proposal text in Section 2.3 to explicitly state this capability so there's no confusion!

docs/design/auth-proposal.md
mahil-2040: I've looked into this carefully and The key insight is that JWT-SVIDs in SPIRE still go through the exact same Workload Attestation flow (Agent → kubelet query → selector matching), so the attestation latency is identical regardless of SVID format.

For PicoD sandboxes specifically, the `spiffe-helper` sidecar fetches the SVID concurrently with PicoD's own container initialization, so attestation (~200-500ms) overlaps with pod startup time rather than adding to it sequentially.

However, for high-throughput scenarios like agentic RL training loops where even this concurrent overhead is unacceptable, our architecture already has a built-in zero-latency alternative: the file-based certificate mode (Section 1.6). In this mode, certificates are pre-provisioned as Kubernetes Secrets and mounted as volumes — they are available the instant the container starts with absolutely no attestation step needed. This makes file-based mode the recommended choice for latency-critical sandbox provisioning.

I've added a dedicated "Latency Considerations for Sandbox Provisioning" note to the proposal to document this tradeoff explicitly.

hzxuzhonghu: you are right, the signing is similar flow as x509, the delay sometimes is un acceptable. In code interpreter, we usually need 100ms bootstrap latency. And in reality, tls handshake takes some time maybe tens of milliseconds.

With jwt, the handshake latency can be eliminated. if we have warmpool, i guess the latency is totally from runtime, not from auth system.

docs/design/auth-proposal.md
hzxuzhonghu: Yeah, i was asking whether you would provide such integration guide

docs/design/auth-proposal.md
mahil-2040: Thank you for clarifying. You're absolutely right that the real bottleneck for latency-critical scenarios like Code Interpreters isn't attestation but the TLS handshake overhead itself (~20-50ms per new connection). With JWT, the Router just sends a Bearer token over plain HTTP and PicoD does a ~1ms signature check - the handshake cost is completely eliminated.

Based on your feedback, I've updated the proposal to support two configurable Router→PicoD auth modes via a `--picod-auth-mode` flag:

1. mtls (default): Full mutual TLS for security-hardened production deployments.
2. jwt: Application-layer JWT authentication for latency-critical scenarios (Code Interpreters, agentic RL). Combined with warm pools, this ensures the auth system adds near-zero overhead.

Router↔WorkloadManager always uses mTLS since both are long-lived and the handshake cost is amortized. The existing PicoD-Plain-Auth code path is retained and improved as the JWT mode implementation rather than being deprecated.

docs/design/auth-proposal.md
hzxuzhonghu: got it, So please add a install option

hzxuzhonghu (approved):
generally LG, @YaoZengzeng to have a another look

PR #245, opened 2026-03-20
improve sandbox access resilience through ServiceFQDN

### LINKED ISSUE #205: [Feature request] Improve sandbox access resilience
We only check sandbox status and get sandbox ip once when 'createSandbox' https://github.com/volcano-sh/agentcube/blob/86841e5f753de719d111425bd2bcb0649a4f5c7d/pkg/workloadmanager/handlers.go#L208

Even after sometime, the pod fails or evicted and recreated, we donot maintain the status any more. This is a missing point from beginning.

cc @YaoZengzeng @acsoto @tjucoder

FAUST-BENCHOU:
**What type of PR is this?**

Fixes https://github.com/volcano-sh/agentcube/issues/205

```release-note

```

### DIFF
--- a/pkg/workloadmanager/handlers.go
+++ b/pkg/workloadmanager/handlers.go
@@ -195,6 +195,49 @@ func (s *Server) createK8sResources(ctx context.Context, dynamicClient dynamic.I
 	return nil
 }
 
+func (s *Server) waitForSandboxReady(ctx context.Context, sandbox *sandboxv1alpha1.Sandbox, resultChan <-chan SandboxStatusUpdate) (*sandboxv1alpha1.Sandbox, error) {
+	// Use NewTimer so we can stop it explicitly when another branch wins,
+	// preventing the runtime from retaining the timer until it fires.
+	timer := time.NewTimer(2 * time.Minute) // consistent with router settings
+
+	select {
+	case result := <-resultChan:
+		timer.Stop()
+		createdSandbox := result.Sandbox
+		klog.V(2).Infof("sandbox %s/%s reported ready, verifying entrypoints", createdSandbox.Namespace, createdSandbox.Name)
+		return createdSandbox, nil
+	case <-ctx.Done():
+		timer.Stop()
+		klog.Warningf("sandbox %s/%s wait canceled: %v", sandbox.Namespace, sandbox.Name, ctx.Err())
+		return nil, ctx.Err()
+	case <-timer.C:
+		klog.Warningf("sandbox %s/%s create timed out", sandbox.Namespace, sandbox.Name)
+		return nil, errSandboxCreationTimeout
+	}
+}
+
+// resolveSandboxAccessHost returns ServiceFQDN when set, otherwise the sandbox pod IP.
+func (s *Server) resolveSandboxAccessHost(ctx context.Context, createdSandbox, sandbox *sandboxv1alpha1.Sandbox) (string, error) {
+	if host := createdSandbox.Status.ServiceFQDN; host != "" {
+		return host, nil
+	}
+
+	// agent-sandbox create pod with same name as sandbox if no warmpool is used
+	// if warmpool is used, the pod name is stored in sandbox's annotation `agents.x-k8s.io/sandbox-pod-name`
+	sandboxPodName := sandbox.Name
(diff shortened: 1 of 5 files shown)


pkg/workloadmanager/handlers.go
```diff
@@ -200,21 +200,22 @@ func (s *Server) createSandbox(ctx context.Context, dynamicClient dynamic.Interf
...
-	if podName, exists := createdSandbox.Annotations[controllers.SandboxPodNameAnnotation]; exists {
-		sandboxPodName = podName
-	}
-
-	podIP, err := s.k8sClient.GetSandboxPodIP(ctx, sandbox.Namespace, sandbox.Name, sandboxPodName)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get sandbox %s/%s pod IP: %v", sandbox.Namespace, sandbox.Name, err)
+	// Prefer ServiceFQDN: K8s Endpoints controller keeps DNS in sync when pod is evicted/recreated.
+	// Fallback to pod IP for agent-sandbox without ServiceFQDN.
+	host := createdSandbox.Status.ServiceFQDN
```
acsoto: `createdSandbox.Status.ServiceFQDN` is preferred here, but upstream agent-sandbox currently hardcodes that field to `.svc.cluster.local`. If a cluster uses a different DNS domain, this may regress cases that previously worked via pod IP fallback.

pkg/workloadmanager/sandbox_helper.go
```diff
@@ -48,7 +49,7 @@ func buildSandboxInfo(sandbox *sandboxv1alpha1.Sandbox, podIP string, entry *san
 		accesses = append(accesses, types.SandboxEntryPoint{
 			Path:     port.PathPrefix,
 			Protocol: string(port.Protocol),
-			Endpoint: net.JoinHostPort(podIP, strconv.Itoa(int(port.Port))),
+			Endpoint: net.JoinHostPort(host, strconv.Itoa(int(port.Port))),
```
acsoto, line 84: Changing `Endpoint: net.JoinHostPort(host, ...)` means `entryPoints[].endpoint` now effectively becomes a service DNS name instead of a pod IP-style address. Router usage looks fine, but it would be good to confirm there are no external consumers depending on the old behavior and to update the API docs if this is intentional.

pkg/workloadmanager/handlers.go
FAUST-BENCHOU: Under non-default cluster domains, there is indeed a theoretical risk of rollback. Because FQDN actually introduces many uncertainties, if this design is not accepted by the community, I will consider another solution for maintaining IPs. Do you think this solution aligns with the community's design, or is abandoning FQDN more reasonable?
anyway thanks for your review! :)

pkg/workloadmanager/sandbox_helper.go
FAUST-BENCHOU: ditto, FQDN is a convenient way to maintain IPs, but it may not meet the needs of the community.

pkg/workloadmanager/handlers.go
hzxuzhonghu: Eventually the upstram should not hard code `cluster.local`, by that assumption, should we accept using service fqdn?

FAUST-BENCHOU: Sorry almost forget this pr. But to be honest , i still think ServiceFQDN is more nature

FAUST-BENCHOU:
@hzxuzhonghu @acsoto updated.PTAL

pkg/workloadmanager/handlers.go
```diff
@@ -195,40 +195,17 @@ func (s *Server) createK8sResources(ctx context.Context, dynamicClient dynamic.I
...
-			return
-		}
-		s.rollbackSandboxCreation(dynamicClient, sandbox, sandboxClaim, sandboxEntry.SessionID)
-	}()
-
-	if err := s.createK8sResources(ctx, dynamicClient, sandbox, sandboxClaim); err != nil {
-		return nil, err
-	}
-
+func (s *Server) waitForSandboxReady(ctx context.Context, sandbox *sandboxv1alpha1.Sandbox, resultChan <-chan SandboxStatusUpdate) (*sandboxv1alpha1.Sandbox, error) {
```
FAUST-BENCHOU, line 198: just refractor to bypass go-lint

PR #247, opened 2026-03-28
Add Python SDK publish workflow

### LINKED ISSUE #246: Need an action to automatically package the Python SDK

Provide a GitHub Action that packages and publishes Python code to PyPL whenever changes are made to the code in the `sdk-python/agentcube`.

**Why is this needed**:

acsoto:
**What type of PR is this?**

/kind enhancement

This PR adds an automated publishing workflow for the Python SDK.

The new workflow builds the package from `sdk-python`, reads the package version from `sdk-python/pyproject.toml`, checks whether that version already exists on PyPI, and only publishes when the version is not already present.

It also tightens the SDK packaging flow by:
- excluding Python cache artifacts from the source distribution
- updating the Python package license metadata to the modern string form
- reusing the existing `make build-python-sdk` path so the root `LICENSE` remains the single source of truth while still being included in built artifacts

Trusted publishing for PyPI has already been configured for this repository, so this workflow does not require a stored PyPI API token. It uses PyPI's trusted publisher flow via GitHub Actions OIDC and `pypa/gh-action-pypi-publish`.

Screenshot 2026-03-28 at 21 38 52

References:
- https://docs.pypi.org/trusted-publishers/adding-a-publisher/
- https://docs.pypi.org/trusted-publishers/using-a-publisher/
- https://github.com/pypa/gh-action-pypi-publish

Fixes #246

The publish job is intentionally limited to `push` on `main` and manual dispatch. Existing Python test workflows remain responsible for test coverage; this workflow focuses on build-and-publish behavior.

```release-note
Added automated PyPI publishing for the Python SDK when `sdk-python` changes are merged to `main`.
```

### DIFF
--- a/.github/workflows/python-sdk-publish.yml
+++ b/.github/workflows/python-sdk-publish.yml
@@ -1,0 +1,63 @@
+name: Python SDK Publish
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "sdk-python/**"
+      - ".github/workflows/python-sdk-publish.yml"
+  workflow_dispatch:
+
+jobs:
+  publish-python-sdk:
+    runs-on: ubuntu-latest
+    environment: pypi
+    permissions:
+      contents: read
+      id-token: write
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install build dependency
+        run: |
+          python -m pip install --upgrade pip
+          pip install build
+
+      - name: Build package
+        run: make build-python-sdk
+
+      - name: Read package version
+        id: package
+        working-directory: sdk-python
+        run: |
+          version="$(python -c 'import tomllib; print(tomllib.load(open("pyproject.toml", "rb"))["project"]["version"])')"
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+
+      - name: Check whether version already exists on PyPI
+        id: pypi
+        env:
+          PACKAGE_NAME: agentcube-sdk
+          PACKAGE_VERSION: ${{ steps.package.outputs.version }}
+        run: |
+          status="$(curl -s -o /dev/null -w "%{http_code}" "https://pypi.org/pypi/${PACKAGE_NAME}/${PACKAGE_VERSION}/json")"
+          if [ "$status" = "200" ]; then
+            echo "exists=true" >> "$GITHUB_OUTPUT"
+            echo "Version ${PACKAGE_VERSION} already exists on PyPI, skipping publish."
+          else
+            echo "exists=false" >> "$GITHUB_OUTPUT"
+            echo "Version ${PACKAGE_VERSION} is not published yet."
+          fi
+
+      - name: Publish package to PyPI
+        if: steps.pypi.outputs.exists != 'true'
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          packages-dir: sdk-python/dist/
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -319,6 +319,12 @@ e2e-clean:
...
-	@echo "Build complete. Artifacts are in sdk-python/dist/"
\ No newline at end of file
+	@tmp_file="$(PROJECT_DIR)/sdk-python/LICENSE"; \
+	cp LICENSE "$$tmp_file"; \
+	set +e; \
+	cd sdk-python && python3 -m build; \
+	status=$$?; \
+	set -e; \
+	rm -f "$$tmp_file"; \
+	exit $$status
```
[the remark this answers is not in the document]
acsoto: fixed

acsoto:
Pls take a look thanks
@LiZhenCheng9527 @YaoZengzeng

.github/workflows/python-sdk-publish.yml
```diff
@@ -0,0 +1,63 @@
...
+          pip install build
+
+      - name: Build package
+        run: make build-python-sdk
+
+      - name: Read package version
+        id: package
+        working-directory: sdk-python
+        run: |
+          version="$(python -c 'import tomllib; print(tomllib.load(open("pyproject.toml", "rb"))["project"]["version"])')"
```
hzxuzhonghu, line 41: IIRC, this is using same version for all the changes on main branch right?

.github/workflows/python-sdk-publish.yml
```diff
@@ -0,0 +1,63 @@
...
+      - name: Check whether version already exists on PyPI
+        id: pypi
+        env:
+          PACKAGE_NAME: agentcube-sdk
+          PACKAGE_VERSION: ${{ steps.package.outputs.version }}
+        run: |
+          status="$(curl -s -o /dev/null -w "%{http_code}" "https://pypi.org/pypi/${PACKAGE_NAME}/${PACKAGE_VERSION}/json")"
+          if [ "$status" = "200" ]; then
+            echo "exists=true" >> "$GITHUB_OUTPUT"
+            echo "Version ${PACKAGE_VERSION} already exists on PyPI, skipping publish."
```
LiZhenCheng9527, line 53: By default, use ‘latest’ as the release version. Since running this GitHub Action indicates that changes have been made to the python-sdk package in agentCube, it should be updated to the ‘latest’ version number even if the version remains the same.

LiZhenCheng9527: Therefore, the ‘version=latest’ should be push unconditionally, followed by the ‘version=x.y.z’ where applicable; to ensure that each agentCube version has the corresponding Python SDK version.

.github/workflows/python-sdk-publish.yml
acsoto: Yes. We use the version declared in `sdk-python/pyproject.toml`, so if that version does not change, later runs on `main` will skip

.github/workflows/python-sdk-publish.yml
acsoto: PyPI package versions are immutable, so unlike container images we cannot keep updating a mutable `latest` tag for the same package version. Re-publishing the same version is not supported. So if we want to do that we need to do a version auto updater

acsoto: However, I think that for the release security, we should explicitly control the version in  `sdk-python/pyproject.toml`

.github/workflows/python-sdk-publish.yml
acsoto: The question is whether we need automatic version updates or manual version control in `sdk-python/pyproject.toml`. I think the latter is more under control

.github/workflows/python-sdk-publish.yml
LiZhenCheng9527: So, is the best practice for PyPI to push a Python SDK version each time an agentCube release is published?

acsoto: I think so. When releasing the agentcube version, manually update the SDK version number

PR #249, opened 2026-04-01
update a python-sdk version

LiZhenCheng9527:
**What type of PR is this?**

update python-sdk project version in pyproject.toml to auto push new python sdk version to pypi

Fixes #

```release-note

```

### DIFF
--- a/sdk-python/pyproject.toml
+++ b/sdk-python/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "agentcube-sdk"
-version = "0.1.0"
+version = "0.1.1.dev0"
 description = "Python SDK for AgentCube Code Interpreter"
 readme = "README.md"
 license = "Apache-2.0"


sdk-python/pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "agentcube-sdk"
-version = "0.1.0"
+version = "0.1.0-rc1"
```
hzxuzhonghu: I think it is better to reflect the dev state.
maybe set to 0.1-dev, after 0.1.0 is released, we update to 0.2-dev， and each time we publish the sdk, we use 0.1-dev-shaID. Not sure if python package use `latest` or not.

acsoto: Python packages usually don’t rely on a `latest` concept. It’s better to encode the development state in the version, e.g. `0.1.dev0` , and bump to `0.2.dev*` after `0.1.0` is released.
I think we can follow sklearn like this:

https://github.com/scikit-learn/scikit-learn/blob/70b05c98024aa3996ef82437932df1124251029f/sklearn/__init__.py#L45

acsoto: image

LiZhenCheng9527: Done

sdk-python/pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "agentcube-sdk"
-version = "0.1.0"
+version = "0.1.dev0"
```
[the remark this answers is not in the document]
acsoto: I'm kind of confused

sdk-python/pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "agentcube-sdk"
-version = "0.1.0"
+version = "0.1.1.dev0"
```
hzxuzhonghu, line 7:
```suggestion
version = "0.1-dev"
```

LiZhenCheng9527: We now have `agentcube-sdk 0.1.0`. If a new release is made with the SDK as `0.1-dev`, it might lead people to think that `0.1.0` is the latest version, which is misleading.

hzxuzhonghu: we need to re push 0.1.0, previous one is not right

FAUST-BENCHOU (commented):
/lgtm

PR #259, opened 2026-04-07
rbac: add resourceclaims/binding for DRA granular status authorization

cairon-ab:
Starting in Kubernetes v1.36, the `DRAResourceClaimGranularStatusAuthorization` feature gate (beta, on-by-default) enforces fine-grained authorization checks for `ResourceClaim` status updates.

Schedulers that modify `status.allocation` or `status.reservedFor` now require additional permissions on the `resourceclaims/binding` synthetic subresource with `update` and `patch` verbs.

This adds the required RBAC rule for the Volcano agent scheduler. The existing `resourceclaims/status` permission is still required and remains unchanged.

These new permissions are inert on Kubernetes versions prior to v1.36, so this is safe to merge now in preparation.

Ref: kubernetes/kubernetes#138149


hzxuzhonghu (changes requested):
@JesseStutler Does agent scheduler support DRA currently?

JesseStutler:
> @JesseStutler Does agent scheduler support DRA currently?

I think yes @hzxuzhonghu, we have already imported the DRA into predicates plugin, and the agent scheduler contains predicates plugin

hzxuzhonghu (approved):
/lgtm

PR #268, opened 2026-04-12
fix: honor per-sandbox SessionTimeout in GC and CodeInterpreter builder

Sanchit2662:
## What type of PR is this?                                                                                                                                                              
  /kind bug

## What this PR does / why we need it:

So I was looking at how the garbage collector handles idle sandboxes and noticed something off. The SessionTimeout field you set on your AgentRuntime or CodeInterpreter CRD is basically ignored at the agentcube level. The GC just hardcodes 15 minutes for everything, no matter what you've configured.

There's also a second problem specific to CodeInterpreter , the SessionTimeout value never even makes it to the sandbox. If you look at buildSandboxByAgentRuntime, the timeout is properly passed into buildParams.idleTimeout. But the same lines are just missing in buildSandboxByCodeInterpreter. So the K8s annotation on the sandbox always ends up as 15m regardless of what you set.                                       
  
## What I changed:

  - Added the missing SessionTimeout propagation in buildSandboxByCodeInterpreter (both the warm-pool and non-warm-pool paths)                                  
  - Added IdleTimeout to sandboxEntry so it flows all the way into the Redis/Valkey store record
  - Both buildSandboxPlaceHolder and buildSandboxInfo now write the correct per-sandbox idle timeout into SandboxInfo                                         
  - Updated ListInactiveSandboxes in both the Redis and Valkey store implementations to use WITHSCORES so we get the actual LastActivityAt timestamp back per 
  session                                                                                                                                                       
  - The GC now checks each sandbox's own IdleTimeout against its real last-activity time instead of applying one global cutoff to everything                  
                                                                                                                                                                
Before this, setting sessionTimeout: 1h meant nothing at the GC layer , your sandbox would still die after 15 minutes of inactivity. Now it actually  respects what you configure.

 ### Special notes for your reviewer:
                                                                                                                                 
The GC still uses DefaultSandboxIdleTimeout (15m) as the query threshold to get initial candidates from the sorted set. That part hasn't changed. The per-sandbox check happens after that as a filter step. So sandboxes with a timeout shorter than 15m won't be picked up by the GC earlier than 15m : that edge case is still handled by the agent-sandbox controller through the K8s annotation. Felt like a reasonable tradeoff for now rather than rearchitecting the whole query.

Also worth noting: LastActivityAt is tagged json:"-" so it's never written into the Redis JSON blob. It's only populated transiently from the sorted set  score during ListInactiveSandboxes. The IdleTimeout field is the one that actually gets persisted.
                                                                                                                                                              
 ### Does this PR introduce a user-facing change?:

Fixed a bug where the garbage collector always deleted idle sandboxes after 15 minutes regardless of the SessionTimeout configured on the AgentRuntime or CodeInterpreter CRD. Also fixed CodeInterpreter not propagating SessionTimeout to the sandbox at all. Per-sandbox idle timeouts are now respected correctly.


Sanchit2662:
Hi @hzxuzhonghu @acsoto , Please review the fix.

pkg/workloadmanager/workload_builder.go
```diff
@@ -312,10 +313,17 @@ func buildSandboxByCodeInterpreter(namespace string, codeInterpreterName string,
...
+	}
+
 	sandboxEntry := &sandboxEntry{
-		Kind:      types.SandboxKind,
-		Ports:     codeInterpreterObj.Spec.Ports,
-		SessionID: sessionID,
+		Kind:        types.SandboxKind,
+		Ports:       codeInterpreterObj.Spec.Ports,
+		SessionID:   sessionID,
+		IdleTimeout: idleTimeout,
```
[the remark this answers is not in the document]
acsoto: +1
I think we should propagate idleTimeout through buildSandboxClaimParams and write the same annotation on SandboxClaim before merging.

Sanchit2662:
Hey @acsoto  , I've addressed all the review points , switched IdleTimeout to metav1.Duration so it persists as "15m0s" instead of a raw nanosecond integer, propagated idleTimeout through buildSandboxClaimParams so SandboxClaim gets the annotation too, added a zero-value guard on LastActivityAt in GC, and fixed the Valkey score parsing to use ParseFloat. Let me know if anything needs another look!

hzxuzhonghu (commented):
Does that mean even the idle timeout is < 15 minutes, it still takes at least 15 minutes to  delete?

Sanchit2662:
@hzxuzhonghu  You're right , the query now uses now as the upper bound so sandboxes with shorter timeouts aren't blocked by the 15-minute floor. A 5-minute timeout will correctly fire at 5 minutes now.

pkg/store/store_valkey.go
```diff
@@ -257,19 +257,51 @@ func (vs *valkeyStore) ListExpiredSandboxes(ctx context.Context, before time.Tim
...
 		return nil, nil
 	}
 
 	maxScore := before.Unix()
-	ids, err := vs.cli.Do(ctx, vs.cli.B().Zrangebyscore().Key(vs.lastActivityIndexKey).Min("-inf").Max(fmt.Sprintf("%d", maxScore)).Limit(0, limit).Build()).AsStrSlice()
+	rawResult, err := vs.cli.Do(ctx,
+		vs.cli.B().Zrangebyscore().Key(vs.lastActivityIndexKey).
+			Min("-inf").Max(fmt.Sprintf("%d", maxScore)).
+			Withscores().Limit(0, limit).Build(),
+	).AsStrSlice()
```
hzxuzhonghu: **Prefer `AsZScores()` over `AsStrSlice()` for WITHSCORES responses**

The valkey-go v1 client provides `AsZScores()` specifically for WITHSCORES responses -- it handles both RESP2 (flat interleaved array) and RESP3 (nested map) formats correctly. With RESP3, scores arrive as native doubles, not bulk strings, so `AsStrSlice()` may fail or silently produce wrong data.

Suggested replacement:
```go
zscores, err := vs.cli.Do(ctx,
    vs.cli.B().Zrangebyscore().Key(vs.lastActivityIndexKey).
        Min("-inf").Max(fmt.Sprintf("%d", maxScore)).
        Withscores().Limit(0, limit).Build(),
).AsZScores()
if err != nil {
    return nil, fmt.Errorf("ListInactiveSandboxes: ZRangeByScore WITHSCORES failed: %w", err)
}
ids := make([]string, 0, len(zscores))
scores := make(map[string]time.Time, len(zscores))
for _, z := range zscores {
    ids = append(ids, z.Member)
    scores[z.Member] = time.Unix(int64(z.Score), 0)
}
```

This also removes the `len%2` guard and the `strconv.ParseFloat` call.

pkg/workloadmanager/garbage_collection.go
```diff
@@ -63,16 +63,37 @@ func (gc *garbageCollector) run(stopCh <-chan struct{}) {
...
 	ctx, cancel := context.WithTimeout(context.Background(), gcOnceTimeout)
 	defer cancel()
-	inactiveTime := time.Now().Add(-DefaultSandboxIdleTimeout)
-	inactiveSandboxes, err := gc.storeClient.ListInactiveSandboxes(ctx, inactiveTime, 16)
+	now := time.Now()
+
+	// Query sandboxes by last-activity time, oldest-first. Using `now` as the upper
+	// bound includes all sandboxes regardless of how short their configured IdleTimeout
+	// is; the per-sandbox filter below decides whether each one has actually timed out.
+	candidates, err := gc.storeClient.ListInactiveSandboxes(ctx, now, 16)
```
hzxuzhonghu: **Incorrect PR description -- query scope was silently widened, risking GC starvation**

The PR description states: *The GC still uses `DefaultSandboxIdleTimeout` (15m) as the query threshold to get initial candidates from the sorted set. That part hasn't changed.* But that is incorrect -- the threshold was changed from `now - DefaultSandboxIdleTimeout` to `now`, which returns all sandboxes from the entire ZSet (up to the limit of 16).

Before:
```go
inactiveTime := time.Now().Add(-DefaultSandboxIdleTimeout)
gc.storeClient.ListInactiveSandboxes(ctx, inactiveTime, 16)
```

The sorted set is ordered by last-activity time ascending. With `now` as the upper bound the GC fetches the 16 oldest-by-activity sandboxes. If those 16 are still within their individual idle timeouts the per-sandbox filter removes them all, and sandboxes at positions 17+ (which may be genuinely idle) are never examined in that GC cycle.

At any scale above 16 concurrent sandboxes this can silently reduce GC effectiveness. Consider using `time.Now().Add(-minConfiguredIdleTimeout)` as a smarter upper bound, or at minimum update the PR description to accurately reflect this behavioral change.

pkg/workloadmanager/garbage_collection.go
```diff
@@ -63,16 +63,37 @@ func (gc *garbageCollector) run(stopCh <-chan struct{}) {
...
 	if err != nil {
 		klog.Errorf("garbage collector error listing inactive sandboxes: %v", err)
 	}
+
+	// Apply per-sandbox idle timeout: only include sandboxes whose own IdleTimeout
+	// (stored in the session JSON) has actually elapsed since LastActivityAt.
+	inactiveSandboxes := make([]*types.SandboxInfo, 0, len(candidates))
```
hzxuzhonghu, line 91: **Missing unit tests for the new per-sandbox filter**

The per-sandbox idle-timeout filtering logic added here is the core behavioral change of this PR, but there are no new or updated unit tests covering it. The `fakeStore` in `handlers_test.go` already stubs `ListInactiveSandboxes` and is a natural fit for a table-driven test. At minimum, please add cases for:
- Sandbox whose `IdleTimeout` has elapsed -> should be collected
- Sandbox still within its `IdleTimeout` -> should be skipped
- `IdleTimeout == 0` (nil `SessionTimeout`) -> should fall back to `DefaultSandboxIdleTimeout`
- `LastActivityAt.IsZero()` -> should be skipped with a warning log

pkg/store/store_redis.go
```diff
@@ -250,23 +250,43 @@ func (rs *redisStore) ListExpiredSandboxes(ctx context.Context, before time.Time
...
+	}
+
+	sandboxes, err := rs.loadSandboxesBySessionIDs(ctx, ids)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, s := range sandboxes {
+		if t, ok := scores[s.SessionID]; ok {
+			s.LastActivityAt = t
```
hzxuzhonghu, line 289: **`LastActivityAt` population is not covered by any test assertion**

Correctly populating `LastActivityAt` from the ZSet score is the critical new contract that `gc.once()` depends on. However, neither `TestListInactiveSandboxes` (Redis) nor `TestValkeyStore_ListInactiveSandboxes` (Valkey) assert that the returned `SandboxInfo.LastActivityAt` matches the timestamp passed to `UpdateSessionLastActivity`.

Please add assertions like:
```go
assert.Equal(t, now.Add(-3*time.Hour).Unix(), list[0].LastActivityAt.Unix())
```
Without this, the `ZRangeByScoreWithScores` / `WITHSCORES` changes are not meaningfully tested.

pkg/workloadmanager/workload_builder.go
```diff
@@ -277,9 +284,10 @@ func buildSandboxByAgentRuntime(namespace string, name string, ifm *Informers) (
 	}
 	sandbox := buildSandboxObject(buildParams)
 	entry := &sandboxEntry{
-		Kind:      types.SandboxKind,
-		Ports:     agentRuntimeObj.Spec.Ports,
-		SessionID: sessionID,
+		Kind:        types.SandboxKind,
+		Ports:       agentRuntimeObj.Spec.Ports,
+		SessionID:   sessionID,
+		IdleTimeout: buildParams.idleTimeout,
```
hzxuzhonghu: **`buildSandboxByAgentRuntime` persists `IdleTimeout: 0` when `SessionTimeout` is unset**

`buildParams.idleTimeout` is only set when `agentRuntimeObj.Spec.SessionTimeout != nil`. If it is nil, `entry.IdleTimeout` is 0 and the persisted `SandboxInfo.IdleTimeout` JSON field will be `"0s"`, forcing the GC to apply a runtime fallback.

By contrast, `buildSandboxByCodeInterpreter` always stores a meaningful non-zero value by initialising `idleTimeout` to `DefaultSandboxIdleTimeout` before the optional override. Applying the same pattern here makes both paths consistent and the zero-value guard in `buildSandboxPlaceHolder`/`buildSandboxInfo` purely defensive:
```go
idleTimeout := DefaultSandboxIdleTimeout
if agentRuntimeObj.Spec.SessionTimeout != nil {
    idleTimeout = agentRuntimeObj.Spec.SessionTimeout.Duration
}
entry := &sandboxEntry{
    Kind:        types.SandboxKind,
    Ports:       agentRuntimeObj.Spec.Ports,
    SessionID:   sessionID,
    IdleTimeout: idleTimeout,
}
```

hzxuzhonghu (changes requested):
The overall direction of this fix is correct -- per-sandbox `IdleTimeout` propagation and store-level `LastActivityAt` population were genuinely missing. Found a few issues that should be addressed before merging.

pkg/store/store_valkey.go
```diff
@@ -257,19 +257,51 @@ func (vs *valkeyStore) ListExpiredSandboxes(ctx context.Context, before time.Tim
...
+	scores := make(map[string]time.Time, len(rawResult)/2)
+	for i := 0; i < len(rawResult); i += 2 {
+		id := rawResult[i]
+		scoreFloat, err := strconv.ParseFloat(rawResult[i+1], 64)
+		if err != nil {
+			return nil, fmt.Errorf("ListInactiveSandboxes: parse score for %s: %w", id, err)
+		}
+		ids = append(ids, id)
+		scores[id] = time.Unix(int64(scoreFloat), 0)
+	}
```
hzxuzhonghu: looks this is different with redis, why?

pkg/workloadmanager/garbage_collection.go
hzxuzhonghu: This is a little bit busy if list all sandboxes. I am thinking we should probably limit the least in active time

Sanchit2662:
I've addressed all issues across the review. All changes are pushed , let me know if anything still looks off or needs another iteration!

pkg/workloadmanager/garbage_collection.go
```diff
@@ -32,6 +32,14 @@ import (
 
 const (
 	gcOnceTimeout = 2 * time.Minute
+	// gcMinInactiveLookback is the shortest idle timeout we support. The GC query
```
hzxuzhonghu: **Comment is factually incorrect: `gcMinInactiveLookback` is not the "shortest idle timeout we support"**

The comment reads: *gcMinInactiveLookback is the shortest idle timeout we support.* This is misleading — the system does support idle timeouts shorter than 1 minute (they are enforced by the K8s annotation on the agent-sandbox controller, not by this GC path). `gcMinInactiveLookback` is a **query window**, not a minimum configuration limit.

Suggested replacement:
```go
// gcMinInactiveLookback is the minimum time a sandbox must have been idle before
// it is included in the GC's inactive candidate query. Sandboxes with an IdleTimeout
// shorter than this window are not caught by this check; those rely on the
// idle-timeout annotation enforced by the agent-sandbox controller.
```

pkg/workloadmanager/garbage_collection_test.go
```diff
@@ -0,0 +1,237 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	dynamicfake "k8s.io/client-go/dynamic/fake"
+
+	"github.com/volcano-sh/agentcube/pkg/common/types"
+	"github.com/volcano-sh/agentcube/pkg/store"
+)
+
+// gcFakeStore is a controllable store for GC tests.
+type gcFakeStore struct {
+	store.Store
```
hzxuzhonghu: **Embedding a nil `store.Store` interface is fragile — future additions to `gc.once()` will panic silently**

Embedding `store.Store` (an interface) as a zero-value field makes any unimplemented method dereference a nil pointer, producing a runtime panic instead of a compile-time or test-failure error. This PR only calls three methods in `gc.once()`, and all three are overridden, so today's tests pass. But if a future change adds a fourth store call to `gc.once()` the test will panic with an opaque nil-pointer message rather than a compilation error.

The standard pattern for partial mocks in Go is to embed a concrete no-op struct rather than the interface:

```go
// nopStore provides no-op implementations of all store.Store methods
// not overridden by gcFakeStore.
type nopStore struct{}
func (nopStore) Ping(_ context.Context) error { return nil }
// ... implement the remaining methods ...

type gcFakeStore struct {
    nopStore  // concrete, not an interface
    ...
}
```

Alternatively, use [testify/mock](https://pkg.go.dev/github.com/stretchr/testify/mock) which already handles this correctly.

hzxuzhonghu (changes requested):
Good progress — all five issues from the previous round are addressed. Three new issues remain before this is ready to merge.

pkg/workloadmanager/garbage_collection.go
```diff
@@ -32,6 +32,10 @@ import (
...
+	// uses this as the lower bound so we avoid scanning sandboxes that were active
+	// very recently while still catching sub-DefaultSandboxIdleTimeout timeouts.
+	gcMinInactiveLookback = 1 * time.Minute
```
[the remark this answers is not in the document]
hzxuzhonghu: And if gcOnceTimeout = 2m and gcMinInactiveLookback = 1 * time.Minute, this is actually confusing.

pkg/common/types/sandbox.go
```diff
@@ -30,9 +32,16 @@ type SandboxInfo struct {
...
-	Status string `json:"status"`
+	// IdleTimeout is the per-sandbox idle timeout configured via SessionTimeout on the
+	// AgentRuntime or CodeInterpreter spec. It is stored in the JSON blob so the
+	// garbage collector can apply it per-sandbox rather than using a global constant.
+	// metav1.Duration marshals as a human-readable string (e.g. "15m0s") rather than
+	// a raw nanosecond integer, making the persisted JSON unambiguous.
+	IdleTimeout metav1.Duration `json:"idleTimeout,omitempty"`
+	// LastActivityAt is populated transiently from the store's last-activity sorted set
+	// during ListInactiveSandboxes. It is intentionally excluded from JSON serialization.
+	LastActivityAt time.Time `json:"-"`
```
hzxuzhonghu, line 43: if this is omitted, how could it be stored in redis/valkey?

pkg/store/store_redis.go
```diff
@@ -250,23 +250,43 @@ func (rs *redisStore) ListExpiredSandboxes(ctx context.Context, before time.Time
...
-		return nil, fmt.Errorf("ListInactiveSandboxes: ZRangeByScore failed: %w", err)
+		return nil, fmt.Errorf("ListInactiveSandboxes: ZRangeByScoreWithScores failed: %w", err)
 	}
 
-	return rs.loadSandboxesBySessionIDs(ctx, ids)
+	ids := make([]string, len(zs))
+	scores := make(map[string]time.Time, len(zs))
+	for i, z := range zs {
+		id := fmt.Sprintf("%v", z.Member)
+		ids[i] = id
```
hzxuzhonghu, line 278: **`fmt.Sprintf("%v", z.Member)` is not type-safe — use a string type assertion**

`redisv9.Z.Member` is typed as `interface{}`. Using `fmt.Sprintf("%v", ...)` produces the string representation without verifying the underlying type, and would silently produce values like `"<nil>"` or `"42"` if the member were anything other than a string.

For consistency with the Valkey path (where `valkey.ZScore.Member` is already `string` and is used directly), apply a proper type assertion here:

```go
for i, z := range zs {
    id, ok := z.Member.(string)
    if !ok {
        return nil, fmt.Errorf("ListInactiveSandboxes: unexpected member type %T", z.Member)
    }
    ids[i] = id
    scores[id] = time.Unix(int64(z.Score), 0)
}
```

pkg/common/types/sandbox.go
Sanchit2662: LastActivityAt is never stored in the JSON blob. It lives as the score in the lastActivityIndex sorted set and is populated transiently from that score during ListInactiveSandboxes (via ZRANGEBYSCORE ... WITHSCORES). The json:"-" tag is intentional: the field only exists on the in-memory struct for the GC's per-sandbox filter, not as a persisted field.

hzxuzhonghu: Does that work if we run multiple workload managers

pkg/workloadmanager/garbage_collection.go
```diff
@@ -63,16 +74,37 @@ func (gc *garbageCollector) run(stopCh <-chan struct{}) {
...
+	candidates, err := gc.storeClient.ListInactiveSandboxes(ctx, now.Add(-gcMinInactiveLookback), gcCandidateLimit)
 	if err != nil {
 		klog.Errorf("garbage collector error listing inactive sandboxes: %v", err)
 	}
+
+	// Apply per-sandbox idle timeout: only include sandboxes whose own IdleTimeout
+	// (stored in the session JSON) has actually elapsed since LastActivityAt.
+	inactiveSandboxes := make([]*types.SandboxInfo, 0, len(candidates))
+	for _, s := range candidates {
+		if s.LastActivityAt.IsZero() {
```
hzxuzhonghu: should we fallback to use s.CreatedAt

pkg/common/types/sandbox.go
Sanchit2662: Yeah, it works fine with multiple instances. LastActivityAt is just a transient view of the sorted-set score in Redis/Valkey, which is shared across all workload managers , so every instance reads the same value. The only race is two instances trying to delete the same sandbox simultaneously, but that's already safe since we treat IsNotFound as success.

pkg/workloadmanager/garbage_collection.go
Sanchit2662: Falling back to s.CreatedAt makes more sense than skipping , a sandbox with no last-activity entry would otherwise be immune to GC indefinitely. If it's newly created, CreatedAt.Add(idleTimeout) won't have elapsed yet so it's still safe. I'll apply the fallback.

Sanchit2662: Done!

pkg/common/types/sandbox.go
hzxuzhonghu: got it

hzxuzhonghu (approved):
generally lg

PR #270, opened 2026-04-13
fix: propagate url.Parse error in buildURL to prevent nil pointer pan…

### LINKED ISSUE #269: bug: router panic on malformed sandbox endpoint due to unhandled `url.Parse` error in `buildURL`
**What happened**:

The `buildURL` function in the router silently discards the error returned by `url.Parse`. When a sandbox entry point contains a malformed or empty endpoint string, `url.Parse` returns `nil` for the URL along with an error — but the error is thrown away with `_`. That `nil` URL is then returned to `determineUpstreamURL`, which wraps it in a successful `(*url.URL, nil)` return, giving the caller no indication anything went wrong. Downstream, `httputil.NewSingleHostReverseProxy(nil)` is called with that nil pointer, causing a **panic that crashes the router process**.

```go
// pkg/router/handlers.go
func buildURL(protocol, endpoint string) *url.URL {
    ...
    url, _ := url.Parse(endpoint)  // error silently dropped
    return url                      // may return nil
}
```

A malformed or empty endpoint in a sandbox's entry points should result in a descriptive HTTP error response (e.g. `500 Internal Server Error` or `502 Bad Gateway`) — not a router panic. The error from `url.Parse` should be propagated up through `determineUpstreamURL` so `forwardToSandbox` can handle it gracefully and return an error to the client.

1. Register a sandbox whose `EntryPoints` contains an entry with a malformed endpoint (e.g. `"://\x00bad"` or an empty string `""`).
2. Send any HTTP request to the router that resolves to that sandbox (e.g. `GET /agents/{namespace}/{name}/`).
3. The router calls `buildURL` → gets a `nil *url.URL` → passes it to `httputil.NewSingleHostReverseProxy(nil)` → **panic / router crash**.

The fix is straightforward — change `buildURL` to return `(*url.URL, error)`, propagate the error up through `determineUpstreamURL`, and handle it in `forwardToSandbox` the same way other URL errors are already handled (log + return HTTP error). No API surface changes required, this is entirely internal to the router package.

Relevant call chain:
`forwardToSandbox` → `determineUpstreamURL` → `buildURL` → `httputil.NewSingleHostReverseProxy`

**Environment**:

- AgentCube version: main (commit `7502b78`)
- Kubernetes version: N/A (reproducible at unit test level without a cluster)
- Others: affects the router component (`cmd/router`) only

sicaario:
**What type of PR is this?**

/kind bug

`buildURL` in the router was silently discarding the error returned by `url.Parse` using a blank identifier (`_`). This meant that when a sandbox entry point contained a malformed or empty endpoint string, `url.Parse` would return a `nil` `*url.URL` which was then passed directly to `httputil.NewSingleHostReverseProxy`. Passing a `nil` URL to the reverse proxy causes a panic at request time, which was only masked by gin's built-in recovery middleware returning a 500 — not proper error handling.

This PR fixes the issue by:
- Changing `buildURL` to return `(*url.URL, error)` and propagating the parse error to the caller
- Adding an `errNoEntryPoint` sentinel to distinguish between "no entry points configured" (404) and "malformed endpoint URL" (500) in `forwardToSandbox`
- Returning a proper `500 Internal Server Error` with a descriptive message instead of relying on panic recovery

Fixes #269

The existing test `TestForwardToSandbox_InvalidEndpoint` was already covering this scenario but was passing for the wrong reason — it expected 500 because gin's panic recovery was catching the nil pointer dereference. The test still passes after this fix, but now tests legitimate error handling instead of panic recovery behaviour.

```release-note
Fixed a router panic caused by malformed sandbox endpoint URLs. Invalid endpoints now return a 500 Internal Server Error instead of crashing the router process.
```

### DIFF
--- a/pkg/router/handlers.go
+++ b/pkg/router/handlers.go
@@ -17,6 +17,7 @@ limitations under the License.
 package router
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"net/http/httputil"
@@ -103,27 +104,33 @@ func (s *Server) handleGetSandboxError(c *gin.Context, err error) {
 	c.JSON(http.StatusInternalServerError, gin.H{"error": "internal server error"})
 }
 
+// errNoEntryPoint is returned when a sandbox has no entry points configured.
+var errNoEntryPoint = errors.New("no entry point found for sandbox")
+
 func determineUpstreamURL(sandbox *types.SandboxInfo, path string) (*url.URL, error) {
 	// prefer matched entrypoint by path
 	for _, ep := range sandbox.EntryPoints {
 		if strings.HasPrefix(path, ep.Path) {
-			return buildURL(ep.Protocol, ep.Endpoint), nil
+			return buildURL(ep.Protocol, ep.Endpoint)
 		}
 	}
 	// fallback to first entrypoint
 	if len(sandbox.EntryPoints) == 0 {
-		return nil, fmt.Errorf("no entry point found for sandbox")
+		return nil, errNoEntryPoint
 	}
 	ep := sandbox.EntryPoints[0]
-	return buildURL(ep.Protocol, ep.Endpoint), nil
+	return buildURL(ep.Protocol, ep.Endpoint)
 }
 
-func buildURL(protocol, endpoint string) *url.URL {
+func buildURL(protocol, endpoint string) (*url.URL, error) {
 	if protocol != "" && !strings.Contains(endpoint, "://") {
-		endpoint = (strings.ToLower(protocol) + "://" + endpoint)
+		endpoint = strings.ToLower(protocol) + "://" + endpoint
 	}
-	url, _ := url.Parse(endpoint)
-	return url
+	u, err := url.Parse(endpoint)
+	if err != nil {
+		return nil, fmt.Errorf("invalid endpoint URL %q: %w", endpoint, err)
+	}
+	return u, nil
 }
 
 // handleAgentInvoke handles agent invocation requests
@@ -148,9 +155,11 @@ func (s *Server) forwardToSandbox(c *gin.Context, sandbox *types.SandboxInfo, pa
 	targetURL, err := determineUpstreamURL(sandbox, path)
 	if err != nil {
 		klog.Errorf("Failed to get sandbox access address %s: %v", sandbox.SandboxID, err)
-		c.JSON(http.StatusNotFound, gin.H{
-			"error": err.Error(),
-		})
+		if errors.Is(err, errNoEntryPoint) {
+			c.JSON(http.StatusNotFound, gin.H{"error": err.Error()})
+		} else {
+			c.JSON(http.StatusInternalServerError, gin.H{"error": "invalid sandbox endpoint configuration"})
+		}
 		return
 	}
 
--- a/pkg/router/handlers.go
+++ b/pkg/router/handlers.go
@@ -180,7 +180,7 @@ func (s *Server) forwardToSandbox(c *gin.Context, sandbox *types.SandboxInfo, pa
 		if errors.Is(err, errNoEntryPoint) {
 			c.JSON(http.StatusNotFound, gin.H{"error": err.Error()})
 		} else {
-			c.JSON(http.StatusInternalServerError, gin.H{"error": fmt.Sprintf("invalid sandbox endpoint configuration: %v", err)})
+			c.JSON(http.StatusInternalServerError, gin.H{"error": "invalid sandbox endpoint configuration"})
 		}
 		return
 	}
--- a/pkg/router/handlers.go
+++ b/pkg/router/handlers.go
@@ -158,7 +158,7 @@ func (s *Server) forwardToSandbox(c *gin.Context, sandbox *types.SandboxInfo, pa
 		if errors.Is(err, errNoEntryPoint) {
 			c.JSON(http.StatusNotFound, gin.H{"error": err.Error()})
 		} else {
-			c.JSON(http.StatusInternalServerError, gin.H{"error": "invalid sandbox endpoint configuration"})
+			c.JSON(http.StatusInternalServerError, gin.H{"error": fmt.Sprintf("invalid sandbox endpoint configuration: %v", err)})
 		}
 		return
 	}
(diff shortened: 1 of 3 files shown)


pkg/router/handlers.go
```diff
@@ -148,9 +155,11 @@ func (s *Server) forwardToSandbox(c *gin.Context, sandbox *types.SandboxInfo, pa
 	targetURL, err := determineUpstreamURL(sandbox, path)
 	if err != nil {
 		klog.Errorf("Failed to get sandbox access address %s: %v", sandbox.SandboxID, err)
-		c.JSON(http.StatusNotFound, gin.H{
-			"error": err.Error(),
-		})
+		if errors.Is(err, errNoEntryPoint) {
```
acsoto, line 180: Since this branch is newly distinguishing missing entry points from malformed URLs, it would be good to add a small test that covers `EntryPoints == nil/empty` and asserts we now return 404 here.

pkg/router/handlers.go
```diff
@@ -148,9 +155,11 @@ func (s *Server) forwardToSandbox(c *gin.Context, sandbox *types.SandboxInfo, pa
...
+			c.JSON(http.StatusNotFound, gin.H{"error": err.Error()})
+		} else {
+			c.JSON(http.StatusInternalServerError, gin.H{"error": "invalid sandbox endpoint configuration"})
```
acsoto: The new behavior makes sense, but the response body here is still fairly generic. Since the PR description mentions a more descriptive error, I suggest to include a sanitized parse failure in the JSON response

pkg/router/handlers.go
hrmn3s: Added `TestForwardToSandbox_NoEntryPoints` in the latest commit (`02fae19`) which sets `EntryPoints: []` and asserts a `404 Not Found` response, covering the `errNoEntryPoint` path.

pkg/router/handlers.go
hrmn3s: @acsoto Updated in `02fae19` — the 500 response now includes the sanitized parse error via `fmt.Sprintf("invalid sandbox endpoint configuration: %v", err)`, so the caller gets context like `invalid endpoint URL "://invalid-url":` instead of the generic message.

acsoto: Thanks, the earlier comments are addressed. I think there is still one edge case left: \url.Parse may succeed for inputs like `""` or `"http://"`, but those are still not usable proxy targets. It would be better to validate `u.Scheme` and `u.Host` in `buildURL` so these cases also return 500 as invalid endpoint configuration, and add a small test for that.`

hrmn3s: @acsoto Added scheme whitelist (http/https/ws/wss) and port range (1–65535) validation in 9dfb8a4 — both cases where url.Parse succeeds but produces an unusable proxy target. Test table updated to cover all edge cases.

pkg/router/handlers.go
```diff
@@ -103,27 +105,54 @@ func (s *Server) handleGetSandboxError(c *gin.Context, err error) {
...
+		endpoint = strings.ToLower(protocol) + "://" + endpoint
+	}
+	u, err := url.Parse(endpoint)
+	if err != nil {
+		return nil, fmt.Errorf("invalid endpoint URL %q: %w", endpoint, err)
+	}
+	if u.Scheme == "" {
+		return nil, fmt.Errorf("invalid endpoint URL %q: missing scheme", endpoint)
 	}
-	url, _ := url.Parse(endpoint)
```
hzxuzhonghu, line 125: donot we have already guranteed this always succeed in 
```
func buildSandboxInfo(sandbox *sandboxv1alpha1.Sandbox, podIP string, entry *sandboxEntry) *types.SandboxInfo {
	createdAt := sandbox.GetCreationTimestamp().Time
	expiresAt := createdAt.Add(DefaultSandboxTTL)
	if sandbox.Spec.ShutdownTime != nil {
		expiresAt = sandbox.Spec.ShutdownTime.Time
	}
	accesses := make([]types.SandboxEntryPoint, 0, len(entry.Ports))
	for _, port := range entry.Ports {
		accesses = append(accesses, types.SandboxEntryPoint{
			Path:     port.PathPrefix,
			Protocol: string(port.Protocol),
			Endpoint: net.JoinHostPort(podIP, strconv.Itoa(int(port.Port))),
		})
```

hrmn3s: BuildSandboxInfo does guarantee the format. The validation still handles the case where Protocol is empty: url.Parse("host:port") succeeds but parses it as scheme="host" with an empty Host, giving a silently unusable proxy target. Also keeps buildURL safe for any future callers. Happy to move the validation upstream if you prefer though.

hzxuzhonghu: I cannot get what you mean by ` but parses it as scheme="host" with an empty Host, `.

From the api object validation, we have a hard limit that it must be set to http or https now

hrmn3s: @hzxuzhonghu to address this buildURL now takes protocol as a separate parameter and prepends it explicitly before parsing — so `"HTTP" + "host:port"` becomes `"http://host:port" `before url.Parse is called. This removes the misparse concern entirely.

The` scheme == ""` check is still kept because SandboxEntryPoint.Protocol is a plain string (not the API-validated ProtocolType), and the session manager can produce entries with an empty Protocol field — as seen in several tests. So it's a real path

pkg/router/handlers.go
```diff
@@ -103,27 +105,54 @@ func (s *Server) handleGetSandboxError(c *gin.Context, err error) {
...
+}
+
+// validProxySchemes are the URL schemes accepted for reverse-proxy targets.
+// ws/wss are included to support WebSocket upgrades.
+var validProxySchemes = map[string]bool{
+	"http": true, "https": true, "ws": true, "wss": true,
 }
 
-func buildURL(protocol, endpoint string) *url.URL {
+func buildURL(protocol, endpoint string) (*url.URL, error) {
```
hzxuzhonghu, line 132: If iirc， the endpoint is constructed by us in buildSandboxInfor, it is podIP:port. And also the protocol can be only http and https now.

So i think this is actually not worth verifying.

PR #271, opened 2026-04-13
Stop publishing Python packages from main branch pushes

acsoto:
**What type of PR is this?**

/kind bug

image

Stop publishing the Python CLI and Python SDK to PyPI on every push to `main`.

PyPI package versions are immutable, so triggering package publishes from regular `main` branch changes causes repeated failures when the package version has not been bumped. This change limits Python package publishing to release tags and manual dispatches, which matches the expected release flow more closely and avoids noisy deployment failures.

Fixes #

This is a stopgap change only. It does not define the long-term Python package versioning or release policy yet.

```release-note
NONE
```

### DIFF
--- a/.github/workflows/python-cli-publish.yml
+++ b/.github/workflows/python-cli-publish.yml
@@ -2,11 +2,8 @@ name: Python CLI Publish
 
 on:
   push:
-    branches:
-      - main
-    paths:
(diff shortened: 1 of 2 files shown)


hzxuzhonghu (approved):
/lgtm

PR #273, opened 2026-04-13
fix: stop blocking on client disconnect and sanitize internal errors in sandbox creation

Aman-Cool:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it:**

When a sandbox pod hits a terminal failure; bad image, OOM kill, eviction, anything; the session creation request would silently block for the full 2-minute timeout before handing back a generic "timed out" error. On top of that, if the client disconnected mid-wait, the server goroutine kept blocking anyway.

This PR adds a `ctx.Done()` arm to the sandbox wait loop so the server stops holding goroutines when the client is already gone, and cleans up the error handling so internal errors (store, K8s API) don't leak system details to callers while sandbox-level failures are still surfaced directly.

**Which issue(s) this PR fixes:**
NA

**Special notes for your reviewer:**

The `ConditionFalse` fast-fail was dropped after review feedback; detecting terminal pod states by substring-matching controller-internal messages couples too tightly to the `agent-sandbox` implementation. Stuck sandboxes are still handled by the 2-minute timeout.

**Does this PR introduce a user-facing change?**

Session creation now returns immediately when the client disconnects instead of blocking for up to 2 minutes. Internal error details are no longer leaked to callers.


Aman-Cool:
@hzxuzhonghu, happy to update anything here ; the main call worth double-checking is the `ConditionFalse` + non-empty `Reason` heuristic for "terminal failure". That's based on how `agent-sandbox` sets conditions today, but if there are cases where a non-empty `Reason` appears during normal startup I'd want to know. Open to tightening that check if needed.

Aman-Cool:
Addressed all the copilot suggestions:

Propagated the actual error message to the HTTP response so callers get something useful instead of a generic 500. Swapped `time.After` for `time.NewTimer` with explicit stops in each branch (no defer, avoids the ordering issue). Dropped the pointer indirection on `SandboxStatusUpdate` in favor of a value + bool flag. Included namespace/name in the reconciler error so it's easier to correlate across concurrent creates. Also extended the status tests to assert the failure message, added an extra case for when `Message` is empty and we fall back to `Reason`.

acsoto:
Thanks. Pls fix lint and e2e

pkg/workloadmanager/handlers.go
```diff
@@ -164,45 +175,32 @@ func (s *Server) createSandbox(ctx context.Context, dynamicClient dynamic.Interf
...
 		if !needRollbackSandbox {
 			return
 		}
-		sandboxRollbackFunc()
+		s.rollbackSandboxCreation(dynamicClient, sandbox, sandboxClaim, sandboxEntry.SessionID)
 	}()
 
+	// Use NewTimer so we can stop it explicitly when another branch wins,
+	// preventing the runtime from retaining the timer until it fires.
+	timer := time.NewTimer(2 * time.Minute) // consistent with router settings
```
hzxuzhonghu, line 221: defer timer.Stop

pkg/workloadmanager/sandbox_helper.go
```diff
@@ -60,17 +61,41 @@ func buildSandboxInfo(sandbox *sandboxv1alpha1.Sandbox, podIP string, entry *san
...
-		if condition.Type == string(sandboxv1alpha1.SandboxConditionReady) && condition.Status == metav1.ConditionTrue {
-			return "running"
+		if condition.Type != string(sandboxv1alpha1.SandboxConditionReady) {
+			continue
+		}
+		if condition.Status == metav1.ConditionTrue {
+			return "running", ""
+		}
+		// ConditionFalse with a reason indicates a terminal failure, not transient pending.
+		if condition.Status == metav1.ConditionFalse && condition.Reason != "" {
```
hzxuzhonghu: I donot think we should return failure when a condition false occur, if you look at the agent-sandbox condition set, it is reconciling there.

YaoZengzeng:
friendly ping @Aman-Cool any updates?

Aman-Cool:
@YaoZengzeng, please take a look🙏

hzxuzhonghu:
Please update the pr title. not sure what it mean.

pkg/workloadmanager/handlers.go
```diff
@@ -132,8 +132,19 @@ func (s *Server) handleSandboxCreate(c *gin.Context, kind string) {
...
+		}
 		klog.Errorf("create sandbox failed %s/%s: %v", sandbox.Namespace, sandbox.Name, err)
-		respondError(c, http.StatusInternalServerError, "internal server error")
+		// Internal errors (store, K8s API) must not leak system details to callers;
+		// sandbox-level failures (terminal pod state, timeout) are safe to surface.
+		msg := err.Error()
+		if apierrors.IsInternalError(err) {
+			msg = "internal server error"
+		}
+		respondError(c, http.StatusInternalServerError, msg)
```
acsoto, line 168: You wrap Internal error but many errors are plain fmt.Errorf

pkg/workloadmanager/handlers.go
```diff
@@ -132,8 +132,19 @@ func (s *Server) handleSandboxCreate(c *gin.Context, kind string) {
 
 	response, err := s.createSandbox(c.Request.Context(), dynamicClient, sandbox, sandboxClaim, sandboxEntry, resultChan)
 	if err != nil {
+		// Client disconnected — nothing to write back, avoid a misleading 500 in metrics.
+		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
```
acsoto: I guess these two types of errors need to be handled separately? context.DeadlineExceeded ma y need a return

test/e2e/e2e_test.go
```diff
@@ -339,41 +339,67 @@ func (e *testEnv) createCodeInterpreterSession(namespace, name string) (string,
...
+		if resp.StatusCode == http.StatusInternalServerError {
+			// Retry on all 500s under load. "Pod exists with phase: Pending" failures
+			// look terminal per request but are transient cluster-wide: the GC cleans
+			// up stuck pods and a subsequent attempt will succeed. Not retrying here
+			// produces a 0% success rate under concurrent load.
+			if attempt < maxRetries-1 {
+				time.Sleep(backoff)
+				backoff *= 2
+				continue
+			}
```
acsoto, line 383: What is the intention here?

Aman-Cool:
Hey @hzxuzhonghu @acsoto, thanks a lot for taking the time to review this; really appreciate the detailed feedback, it helped me catch a few things I'd missed.

Went ahead and fixed the error wrapping issue; you're right that `GetSandboxPodIP`, `waitForSandboxEntryPointsReady`, and `UpdateSandbox` were returning plain `fmt.Errorf` errors that wouldn't be caught by `apierrors.IsInternalError`, so those internal details could've leaked to callers. Wrapped all three with `api.NewInternalError` now.

On `defer timer.Stop`; I kept the explicit `timer.Stop()` calls in both the `resultChan` and `ctx.Done()` branches of the select instead of using a defer, since every exit path through the select already stops the timer. Happy to switch to a defer if you feel that's cleaner though, just wanted to understand your preference there.

For the `ConditionFalse` concern; I totally get the hesitation. The current logic only treats `ConditionFalse` as terminal when a non-empty `Reason` is set, and it explicitly carves out known transient states like "Operation cannot be fulfilled", "Pod exists with phase: Pending", and "Pod is Running but not Ready" as unknown rather than failed. Curious though; are there other cases in the `agent-sandbox` condition set you had in mind that this might be catching incorrectly? Would love to understand the reconciliation flow better there.

For the `context.Canceled` / `context.DeadlineExceeded` handling; both currently short-circuit with a plain return before we hit `respondError`, so neither ends up in the 500 path. But I might be missing something; did you have a specific scenario in mind where you'd want them handled differently?

And for the e2e retry block; the intention comment right above it tries to explain the reasoning (transient "Pod exists with phase: Pending" failures under load that look terminal per-request but resolve after GC), but let me know if that's not clear enough and I can expand it.

Thanks again, open to any pushback on any of this!

Aman-Cool:
@hzxuzhonghu @acsoto, On further thought, I want to correct myself on the `context.DeadlineExceeded` point; I said both errors are "client disconnected" but that's not quite right. `context.Canceled` is the client dropping, but `context.DeadlineExceeded` is actually a server-side request timeout firing, so silently returning nothing there might not be the right call. Should that be returning a 504 instead? Happy to split the handling if so.

hzxuzhonghu (commented):
## 代码审查

**总体评价：修复了真实的用户体验问题，代码质量较高，建议合并。**

### 问题分析

原始问题：sandbox pod 进入终态失败（镜像拉取失败、OOM、驱逐等）时，session 创建请求会阻塞满 2 分钟超时才返回，而且返回的是无意义的 "timed out" 错误，客户端无从判断是否应该重试。

### 代码变更分析

**handlers.go 改进：**
1. 新增 ctx.Done() 分支 — 客户端断连时不再傻等，提前退出，避免无效的 goroutine 驻留。
2. time.After 换为 time.NewTimer + 显式 Stop() — 防止 timer 在非超时分支退出后泄漏，资源管理更严格。
3. 错误信息区分机制：内部错误统一返回 "internal server error"，沙箱级别失败透传具体消息，安全性和可调试性兼顾。

**rollback 重构：**
使用独立的 context.Background() + 30s 超时执行回滚，确保客户端断连不影响资源清理，这是正确的做法。

**测试：**
新增了对内部错误脱敏和非内部错误透传的验证，覆盖完整。

LGTM

hzxuzhonghu (commented):
LGTM. The fix improves UX by notifying clients immediately. Error classification (internal vs sandbox) is handled safely, and using an independent context for rollback is correct.

pkg/workloadmanager/sandbox_helper.go
```diff
@@ -93,20 +98,52 @@ func buildSandboxInfo(sandbox *sandboxv1alpha1.Sandbox, podIP string, entry *san
...
+				return sandboxStatusUnknown, ""
+			}
+			// "Pod exists with phase: Pending" and "Pod is Running but not Ready" are
+			// intermediate states emitted by the sandbox controller during pod startup.
+			// Both are transient — the pod will eventually transition to Running+Ready.
+			// Treat as unknown so the 2-minute timeout decides the outcome.
+			if strings.Contains(msg, "Pod exists with phase: Pending") ||
+				strings.Contains(msg, "Pod is Running but not Ready") {
+				return sandboxStatusUnknown, ""
+			}
```
hzxuzhonghu: why do you only check "Operation cannot be fulfilled" and "Pod exists with phase: Pending"

hzxuzhonghu (changes requested):
I do think this is ai generated without personal thinking at all

pkg/workloadmanager/sandbox_helper.go
Aman-Cool: These three are the specific messages the `agent-sandbox` controller emits during normal pod startup; "Operation cannot be fulfilled" comes from a K8s 409 conflict when the controller races to update the pod (it retries automatically), and "Pod exists with phase: Pending" / "Pod is Running but not Ready" are just intermediate states the controller sets while the pod is coming up. Everything else with a non-empty `Reason`; like `ErrImagePull`, `OOMKilled`, eviction, etc. ; falls through as terminal, which is the intended behaviour.

hzxuzhonghu: Do you read the agent-sandbox implementation? 

We donot want to couple much with the details

Aman-Cool: Fair point, I didn't go through the `agent-sandbox` controller source before adding those checks. 
Looking at the API, there are no formal reason constants for those intermediate states, so I was essentially coupling to internal message strings that could change. 
Would it make more sense to drop the `ConditionFalse` fast-fail entirely and just let the timeout handle stuck sandboxes, keeping only the `ctx.Done()` arm for client disconnects? Or is there a cleaner way you had in mind to detect terminal failures without poking at controller internals?

hzxuzhonghu: From my perspective, it would be ok to merge without this part change.

Aman-Cool:
@hzxuzhonghu, thanks a lot for the thorough review as always.😊

Dropped the `ConditionFalse` fast-fail entirely; you were right that substring-matching on controller-internal messages was too tightly coupled to the `agent-sandbox` implementation. Stuck sandboxes will still be caught by the 2-minute timeout.

Also fixed the error wrapping inconsistency; `GetSandboxPodIP`, `waitForSandboxEntryPointsReady`, and `UpdateSandbox` were returning plain `fmt.Errorf` errors that wouldn't be caught by `apierrors.IsInternalError`, so internal details could've leaked to callers. Wrapped all three with `api.NewInternalError` now.

Updated the PR description and title accordingly to reflect what the PR actually does now.

pkg/workloadmanager/sandbox_controller.go
```diff
@@ -48,31 +48,26 @@ func (r *SandboxReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		delete(r.watchers, req.NamespacedName)
+	}
+	r.mu.Unlock()
+
+	if exists {
+		// WatchSandboxOnce always creates a buffered channel of size 1, and the
+		// map entry is deleted before this point so only one sender can ever
+		// reach here for a given key. The buffer is therefore always empty and
+		// this send never blocks.
+		resultChan <- SandboxStatusUpdate{Sandbox: sandbox}
```
hzxuzhonghu: please add back the default branch, in case the receiver is not there.

pkg/workloadmanager/handlers.go
acsoto: @Aman-Cool I have no idea why this comment is addressed without any actions. Could you explain this to me?

pkg/workloadmanager/sandbox_controller.go
Aman-Cool: Done 👍

acsoto:
image

Aman-Cool:
@acsoto; does everything looks good now?

acsoto:
> @acsoto; does everything looks good now?

@Aman-Cool still a few questions haven't been answered by you

image

image

Aman-Cool:
>  guess these two types of errors need to be handled separately? context.DeadlineExceeded ma y need a return

@acsoto, Apologies for not replying before resolving, that was my bad.

You're right that they're semantically different errors, and I didn't think about that distinction when I wrote this. After your comment I looked into it more carefully, and the reason I didn't split them is that the context being passed into `createSandbox` is `c.Request.Context()` with no additional wrapping; no `context.WithTimeout` or `context.WithDeadline` on top of it. So both `Canceled` and `DeadlineExceeded` here come from the same source: the HTTP request lifecycle itself. When either fires, the underlying connection is already gone and writing a response back would fail silently anyway. The `return` on line 138 already covers both branches.

If there was an internal deadline wrapping the request context, I'd absolutely split them and send back a 504 for the timeout case. But since that's not the case here, the behavior is the same for both. 
Let me know if you see it differently and I'm happy to revisit...

Aman-Cool:
> You wrap Internal error but many errors are plain fmt.Errorf

@acsoto, Sorry for resolving this also without explaining.

You were right. When I introduced the `apierrors.IsInternalError` check to selectively expose error messages, I'd wrapped the store and K8s API errors in `createSandbox` but missed a couple; specifically the pod IP fetch and the store update cache call, which were still plain `fmt.Errorf` and would have leaked internal details. I caught that after your comment and wrapped those in `api.NewInternalError` as well (the pod IP and `UpdateSandbox` errors). The only unwrapped one left is `"sandbox creation timed out"`, which is intentional; it's a safe message to surface to the caller.

pkg/workloadmanager/sandbox_helper.go
```diff
@@ -33,6 +33,9 @@ const (
 	defaultSandboxReadyProbeTimeout  = 15 * time.Second
 	defaultSandboxReadyProbeInterval = 1 * time.Second
 	defaultSandboxReadyDialTimeout   = 1 * time.Second
+
+	sandboxStatusRunning = "running"
+	sandboxStatusUnknown = "unknown"
```
LiZhenCheng9527: IMO, `unknow` is not a good expression.
Maybe `Ready` and `UnReady` is better.
https://github.com/agent-sandbox/agent-sandbox/blob/ff327f8cf827582b74a3dae72f4310c7a458c2a7/pkg/sandbox/sandbox.go#L34

acsoto:
> > You wrap Internal error but many errors are plain fmt.Errorf
> 
> @acsoto, Sorry for resolving this also without explaining.
> 
> You were right. When I introduced the `apierrors.IsInternalError` check to selectively expose error messages, I'd wrapped the store and K8s API errors in `createSandbox` but missed a couple; specifically the pod IP fetch and the store update cache call, which were still plain `fmt.Errorf` and would have leaked internal details. I caught that after your comment and wrapped those in `api.NewInternalError` as well (the pod IP and `UpdateSandbox` errors). The only unwrapped one left is `"sandbox creation timed out"`, which is intentional; it's a safe message to surface to the caller.

Thanks. For the wrapping point, using %w before api.NewInternalError preserves the cause only until the error is converted into apimachinery StatusError. apierrors.NewInternalError does not expose the wrapped error through errors.Unwrap, so handleSandboxCreate still cannot match context.Canceled/context.DeadlineExceeded once those errors are converted.

Aman-Cool:
@LiZhenCheng9527 agreed, "unknown" was misleading; it implies we don't know the state rather than "not ready". Went ahead and renamed both constants to ready/not-ready to make the pair consistent.

Aman-Cool:
@acsoto, you're right that once a context error gets wrapped by `api.NewInternalError`, it becomes invisible to `errors.Is` since `StatusError` doesn't implement `Unwrap`. 
Fixed by adding a context error check before every `api.NewInternalError` wrap in `createSandbox`, so cancellations during store and K8s calls are returned directly and stay detectable in the handler.

acsoto (approved):
LGTM, thanks

Aman-Cool:
@hzxuzhonghu, could you also take a look 🙏

hzxuzhonghu (approved):
Thank you everyone

PR #276, opened 2026-04-15
fix: honor MaxSessionDuration in placeholder expiry ZSet score

Sanchit2662:
  /kind bug                           
                                                                                                                                    
so basically `buildSandboxPlaceHolder` was always setting `ExpiresAt` to `time.Now().Add(DefaultSandboxTTL)` which is a hardcoded 8h, no matter what `MaxSessionDuration` the user had set on their CodeInterpreter or AgentRuntime CR.                                                                                 
the problem is `StoreSandbox` writes that expiry score into the Redis/Valkey ZSet right at placeholder time, and `UpdateSandbox` (called later when the sandbox is actually ready) only updates the JSON blob ; it doesn't touch the ZSet. so the GC  always saw 8h, not whatever the user configured.                                                                                                                                                                                                                                                  
                                                                                                                     

the fix is pretty small , just read `Spec.Lifecycle.ShutdownTime` from the sandbox CR in `buildSandboxPlaceHolder` if it's set, and only fall back to `DefaultSandboxTTL` when it's nil. the sandbox CR already has `ShutdownTime` populated by `buildSandboxObject` before `buildSandboxPlaceHolder` is called, so no extra plumbing needed.                                                                                          also added table-driven tests for this in `sandbox_helper_test.go` covering the three cases: no ShutdownTime (fallback), 24h duration, and 30m duration.                                                                                                                                                                                
                                                                                                                                                         
  
###  **Special notes for your reviewer**:                                                                                                                                 
the pattern I used in `buildSandboxPlaceHolder` is the same one already in `buildSandboxInfo` , so it should look familiar. no changes to the store interface  or any other layer.                                                                                                                                                

### **Does this PR introduce a user-facing change?**:                                                                                                                  
  ```release-note                                                                                                                                                    
  Fixed a bug where `MaxSessionDuration` set on CodeInterpreter or AgentRuntime was
  being ignored by the GC. Sandboxes were always expiring at 8h regardless of what                                                                                   
  was configured. Now the expiry ZSet score correctly reflects the user's configured                                                                                 
  `MaxSessionDuration`.

### DIFF
--- a/pkg/workloadmanager/sandbox_helper.go
+++ b/pkg/workloadmanager/sandbox_helper.go
@@ -27,9 +27,11 @@ import (
 )
 
 func buildSandboxPlaceHolder(sandboxCR *sandboxv1alpha1.Sandbox, entry *sandboxEntry) *types.SandboxInfo {
-	expiresAt := time.Now().Add(DefaultSandboxTTL)
+	var expiresAt time.Time
 	if sandboxCR.Spec.Lifecycle.ShutdownTime != nil {
 		expiresAt = sandboxCR.Spec.Lifecycle.ShutdownTime.Time
+	} else {
+		expiresAt = time.Now().Add(DefaultSandboxTTL)
 	}
 	return &types.SandboxInfo{
 		Kind:             entry.Kind,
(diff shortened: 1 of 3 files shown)


Sanchit2662:
hey @hzxuzhonghu  , I have addressed a few of the Copilot suggestions. switched to if/else in buildSandboxPlaceHolder to skip the unnecessary time.Now() call when ShutdownTime is already set. also fixed the warm-pool path in workload_builder.go since simpleSandbox was being built without ShutdownTime, so MaxSessionDuration was still being ignored there even after the original fix. tightened the test assertions to assert.Equal for the exact-value cases and added a regression test for the warm-pool flow too.

hzxuzhonghu (approved):
/lgtm

PR #281, opened 2026-04-16
fix: skip redis.Nil in pipeline Exec to prevent GC batch abort on orphaned zset entries

### LINKED ISSUE #279: bug: Redis pipeline Nil error aborts entire GC batch when zset index has orphaned entries
`loadSandboxesBySessionIDs` pipelines `GET` calls for all session IDs returned by the sorted-set range query. `go-redis/v9` `Pipeline.Exec` returns the first error from any queued command, including `redis.Nil` for a missing key. The early-return on `pipeErr != nil` fires before the per-command loop that would have correctly skipped missing keys with `continue`, so the entire batch is aborted.

Over time, orphaned index entries accumulate and the GC becomes progressively more broken — eventually failing on every run.

Sandboxes whose sorted-set index entry exists but whose hash key has been evicted should be silently skipped. The remaining valid sandboxes in the batch should still be returned and garbage collected normally.

1. Configure Redis with `maxmemory-policy allkeys-lru` and a low `maxmemory` limit.
2. Create several sandboxes so that session hash keys are written to Redis.
3. Allow Redis to evict one or more session hash keys while their sorted-set entries (`session:expiry` / `session:last_activity`) remain.
4. Wait for or trigger a GC cycle.
5. Observe that no sandboxes are deleted and the error above is logged.

Alternatively, reproduce in a unit test by storing a sandbox, deleting only its hash key via `DEL session:<id>` (leaving the zset entry), then calling `ListExpiredSandboxes` or `ListInactiveSandboxes`.

- The Valkey store (`store_valkey.go`) uses `MGET` which returns empty strings for missing keys and is **not** affected by this bug.
- The fix is a one-line guard: `!errors.Is(pipeErr, redisv9.Nil)` on the early-return in `loadSandboxesBySessionIDs`.
- A regression test using `miniredis` that directly reproduces the failure is included in the accompanying PR.

**Environment**:

- agentcube version:
- Kubernetes version:
- Others: Redis with `allkeys-lru` or any eviction policy that can remove hash keys independently of sorted-set entries

sicaario:
**What type of PR is this?**

/kind bug

`loadSandboxesBySessionIDs` uses a Redis pipeline to batch-fetch session hash
keys. `go-redis/v9` `Pipeline.Exec` returns the first error from any queued
command — including `redis.Nil` for a missing key. The existing early-return on
`pipeErr != nil` fired before the per-command loop that correctly skipped
missing keys with `continue`, causing the entire GC batch to be aborted
whenever any session's hash key was absent.

In production this surfaces when Redis evicts hash keys under memory pressure
(`allkeys-lru`) while leaving sorted-set index entries intact. Every subsequent
GC batch that includes those orphaned entries fails, preventing expired and idle
sandboxes from being cleaned up and causing orphaned index entries to accumulate
indefinitely.

Fix: guard the early return with `!errors.Is(pipeErr, redisv9.Nil)` so only
real errors (network failures, etc.) abort the batch. Missing keys continue to
be skipped per-command as intended. A regression test using `miniredis` is
included that directly reproduces the failure.

Fixes #279 

The valkey store (`store_valkey.go`) uses `MGET` which returns empty strings for
missing keys and is not affected by this bug.

```release-note
Fixed a bug where Redis LRU eviction of session hash keys caused garbage
collection to silently skip entire batches of expired/idle sandboxes.

### DIFF
--- a/pkg/store/store_redis.go
+++ b/pkg/store/store_redis.go
@@ -89,7 +89,7 @@ func (rs *redisStore) loadSandboxesBySessionIDs(ctx context.Context, sessionIDs 
 		sandboxCommands[i] = pipe.Get(ctx, sessionKey)
 	}
 	_, pipeErr := pipe.Exec(ctx)
-	if pipeErr != nil {
+	if pipeErr != nil && !errors.Is(pipeErr, redisv9.Nil) {
 		return nil, fmt.Errorf("redis pipeline exec failed: %w", pipeErr)
 	}
 
--- a/pkg/store/store_redis_test.go
+++ b/pkg/store/store_redis_test.go
@@ -240,6 +240,44 @@ func TestListInactiveSandboxes(t *testing.T) {
 	}
 }
 
+// TestLoadSandboxesBySessionIDs_OrphanedZSetEntry verifies that
+// loadSandboxesBySessionIDs skips session IDs whose hash key has been evicted
+// from Redis (orphaned sorted-set entry) instead of aborting the entire batch.
+//
+// This scenario occurs in production when Redis evicts hash keys under memory
+// pressure (allkeys-lru policy) while leaving sorted-set index entries intact,
+// causing garbage collection to fail for the whole batch.
+func TestLoadSandboxesBySessionIDs_OrphanedZSetEntry(t *testing.T) {
+	ctx := context.Background()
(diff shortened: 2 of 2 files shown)


hzxuzhonghu (approved):
/lgtm

PR #282, opened 2026-04-16
test: add unit tests for garbageCollector.once()

Sanchit2662:

  /kind test
  
so `garbageCollector.once()` is basically the core function that decides which sandboxes get cleaned up , it queries inactive and expired sandboxes from the store and deletes them. but there were zero tests for it. none. any regression in this path would just silently ship.i added a proper test file for it. i built a `gcFakeStore` that lets me control what `ListInactiveSandboxes`, `ListExpiredSandboxes`, and `DeleteSandboxBySessionID` return independently, and paired it with a fake dynamic client so no real cluster is needed to run these.

                                                                                                                                                       
###  **Special notes for your reviewer**:                                                                                                                           
the fake dynamic client returns "not found" for all deletes since there are no pre-loaded objects and the GC already treats not-found as success, so the happy-path tests work cleanly without needing to register any CRD types in the scheme.
the `inactiveCutoffUsesDefaultIdleTimeout` test captures the actual `before` timestamp passed to `ListInactiveSandboxes` and asserts it matches `now - DefaultSandboxIdleTimeout` within a small window ; this locks in the timeout constant so if someone changes it accidentally the test breaks.

### DIFF
--- a/pkg/workloadmanager/garbage_collection_test.go
+++ b/pkg/workloadmanager/garbage_collection_test.go
@@ -1,0 +1,214 @@
+/*
+Copyright The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


Sanchit2662:
Closing this as superseded. The garbage_collection_test.go coverage for garbageCollector.once() has already landed on main via #268

PR #283, opened 2026-04-16
fix: skip redis.Nil in pipeline exec to prevent GC list failures when a session key is concurrently deleted

Storm1289:
/kind bug

`Pipeline.Exec()` in go-redis v9 returns the first non-nil error from any command, which includes `redis.Nil` when a `GET` hits a missing key. In `loadSandboxesBySessionIDs`, the pipeline-level error check was firing on `redis.Nil` and returning early — before the per-command loop (which already handles `redis.Nil` via `continue`) could run.

This caused both `ListExpiredSandboxes` and `ListInactiveSandboxes` to return an error and zero results whenever a session key was concurrently deleted between the sorted-set scan and the GET pipeline. A realistic race: `DeleteSandboxBySessionID` uses a non-atomic pipeline, so a parallel GC run can remove the key while its sorted-set entry is still visible. The net effect was the garbage collector silently doing nothing and expired/inactive sandboxes accumulating.

## Code Change (`pkg/store/store_redis.go`)

```go
// Before
_, pipeErr := pipe.Exec(ctx)
if pipeErr != nil {
    return nil, fmt.Errorf("redis pipeline exec failed: %w", pipeErr)
}

// After
_, pipeErr := pipe.Exec(ctx)
// redis.Nil means a GET found no key (concurrent delete); skip it, the per-command loop handles it.
if pipeErr != nil && !errors.Is(pipeErr, redisv9.Nil) {
    return nil, fmt.Errorf("redis pipeline exec failed: %w", pipeErr)
}
```

## Which issue(s) this PR fixes

None — self-contained bug found during code review.

## Special notes for reviewer

`TestListExpiredSandboxes_MissingKey` reproduces the race exactly: stores two sandboxes, manually deletes one session key while leaving its sorted-set entry intact, then asserts the list returns the remaining sandbox without error. This test fails against the previous code.

## Release Note

```release-note
Fix Redis store GC silently skipping all expired/inactive sandbox cleanup when any session key is concurrently deleted
```

### DIFF
--- a/pkg/store/store_redis.go
+++ b/pkg/store/store_redis.go
@@ -89,7 +89,8 @@ func (rs *redisStore) loadSandboxesBySessionIDs(ctx context.Context, sessionIDs 
 		sandboxCommands[i] = pipe.Get(ctx, sessionKey)
 	}
 	_, pipeErr := pipe.Exec(ctx)
-	if pipeErr != nil {
+	// redis.Nil means a GET found no key (concurrent delete); skip it, the per-command loop handles it.
+	if pipeErr != nil && !errors.Is(pipeErr, redisv9.Nil) {
 		return nil, fmt.Errorf("redis pipeline exec failed: %w", pipeErr)
 	}
 
--- a/pkg/store/store_redis_test.go
+++ b/pkg/store/store_redis_test.go
@@ -183,6 +183,43 @@ func TestListExpiredSandboxes(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


hzxuzhonghu (commented):
isn't this duplicate with #281

Storm1289:
Hi @hzxuzhonghu, sorry I did not know there was already a PR (#281) open for this bug. Thank you for the review.

PR #287, opened 2026-04-17
fix: copy pod labels in buildSandboxObject to avoid mutating the informer cache

Storm1289:
/kind bug

`buildSandboxObject` injected `SessionIdLabelKey` and `SandboxNameLabelKey` directly into `params.podLabels` in-place. Both `buildSandboxByAgentRuntime` and `buildSandboxByCodeInterpreter` assigned the labels map from the informer-cached CRD object to `params.podLabels` without copying it first.

Because `params.podLabels` and `sandbox.Spec.PodTemplate.ObjectMeta.Labels` pointed at the same underlying map as the informer cache, every sandbox creation silently wrote session-specific keys back into the shared cache. Under concurrent requests for the same AgentRuntime or CodeInterpreter, multiple goroutines mutated the same map simultaneously  a data race  and each request could overwrite the session label left by the previous one, causing pods to carry the wrong session ID.

Note: `envVars` in `buildSandboxByCodeInterpreter` was already explicitly copied with `make` + `copy` to avoid this exact problem. The labels map was missed.

## Code Change

```go
// Before: params.podLabels points directly at the informer-cached CRD labels map
sandbox.Spec.PodTemplate.ObjectMeta.Labels[SessionIdLabelKey] = params.sessionID
sandbox.Spec.PodTemplate.ObjectMeta.Labels[SandboxNameLabelKey] = params.sandboxName

// After: always allocate a fresh map so we never mutate the informer-cached object
podLabels := make(map[string]string, len(params.podLabels)+2)
maps.Copy(podLabels, params.podLabels)
podLabels[SessionIdLabelKey] = params.sessionID
podLabels[SandboxNameLabelKey] = params.sandboxName
```

## What tests cover this fix

New test file `pkg/workloadmanager/workload_builder_test.go` with two cases:

- `TestBuildSandboxObject_DoesNotMutateCallerLabels`  passes a labels map, calls `buildSandboxObject`, and asserts the original map is unchanged. Fails on old code, passes on the fix.
- `TestBuildSandboxObject_NilLabels`  verifies that a nil `podLabels` input still produces a sandbox with the correct injected session labels.

## Special notes for reviewer

PR #268 also modifies `workload_builder.go` (SessionTimeout propagation). If that lands first there may be a trivial merge conflict in this file the two fixes touch different lines and are independent.

## Release Note

```release-note
fix: pod labels injected by workload-manager no longer race against concurrent sandbox creation requests for the same AgentRuntime or CodeInterpreter
```

### DIFF
--- a/pkg/workloadmanager/workload_builder.go
+++ b/pkg/workloadmanager/workload_builder.go
@@ -167,8 +167,11 @@ func buildSandboxObject(params *buildSandboxParams) *sandboxv1alpha1.Sandbox {
 	podLabels[SessionIdLabelKey] = params.sessionID
 	podLabels[SandboxNameLabelKey] = params.sandboxName
 
-	podAnnotations := make(map[string]string, len(params.podAnnotations))
-	maps.Copy(podAnnotations, params.podAnnotations)
+	var podAnnotations map[string]string
+	if params.podAnnotations != nil {
+		podAnnotations = make(map[string]string, len(params.podAnnotations))
+		maps.Copy(podAnnotations, params.podAnnotations)
+	}
 
 	// Create Sandbox object using agent-sandbox types
 	sandbox := &sandboxv1alpha1.Sandbox{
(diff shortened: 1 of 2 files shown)


hzxuzhonghu (approved):
/lgtm

PR #291, opened 2026-04-20
feat: add NetworkPolicy support for sandbox isolation

### LINKED ISSUE #216: RFC: Support network policy for sandbox

Added network policy for sandbox, k8s has existing networkpolicy. And even in some service mesh, some more high level limitation is supported.

I would suggest we can think about this feature.

**Why is this needed**:

Sandbox require strong isolation and security, in terms of this, we should provide limitations on ingress and egress network access.

Sanchit2662:

/kind feature

Sandboxes had zero network restrictions. A pod inside a sandbox could reach anything, internal or external. Since sandboxes run untrusted code this is a real isolation gap, so I picked up issue #216 to add NetworkPolicy support.

I added a `SandboxNetworkPolicy` field to both `SandboxTemplate` (AgentRuntime) and `CodeInterpreterSandboxTemplate`. Enforcement is controlled by whether the field is set:
- **Omitted (nil)** — no NetworkPolicy is created; existing setups are unaffected
- **Set (non-nil)** — a deny-all baseline NetworkPolicy is created with two built-in allow rules: DNS egress on port 53 (UDP+TCP) so pods keep name resolution, and router ingress so the router can always reach the sandbox. Additional CIDR+port rules can be added via the `Egress` and `Ingress` fields.

A few specific things I was careful about:

**Cross-namespace router ingress** — using just `PodSelector` would only match pods in the same namespace as the NP. In a real deployment the router usually lives in a different namespace than the sandbox pods, so that would silently break the router rule. I fixed this by adding a `NamespaceSelector` using `kubernetes.io/metadata.name` so it works across namespaces.

**Router selector is configurable** — you can set it in `values.yaml` under `workloadmanager.networkPolicy.routerSelector` and `routerNamespace`, or pass `--router-selector` and `--router-namespace` flags. Defaults to `app=agentcube-router` and the workloadmanager's own namespace.

**Per-session override** — added `NetworkPolicy` field to `CreateSandboxRequest`. When set it fully replaces the template-level policy for that session only. No merging, just a clean replace. Overrides are rejected with 403 when `EnableAuth=true`.

**Cleanup** — NPs are deleted when you delete the sandbox through the API, on rollback if creation fails partway through, and in the GC loop. I also fixed a bug where the GC was returning early on `IsNotFound` before deleting the NP. Since the store record gets purged right after, that would've leaked the NP permanently. Now NP cleanup runs regardless of whether the sandbox was already gone.

**Rollback** — simplified to always attempt NP delete when `networkPolicy != nil` rather than tracking a `npCreated` flag. This also handles the edge case where the k8s write succeeded but returned an error to us.

**RBAC** — added `networking.k8s.io/networkpolicies` create/get/list/watch/patch/delete to the workloadmanager ClusterRole.

### **Which issue(s) this PR fixes**:
Fixes #216

## **Special notes for your reviewer**:

Known gap: for the warm-pool path (SandboxClaim), agent-sandbox doesn't propagate `SandboxNameLabelKey` onto warm-pool pods, so a NetworkPolicy selecting on that label would never match them. To avoid a false sense of isolation, the warm-pool path now skips NP creation entirely and logs a warning when a policy is configured but can't be enforced. Proper enforcement needs label propagation on the agent-sandbox side first — happy to do that as a follow-up.

## **Does this PR introduce a user-facing change?**:
```release-note
Added NetworkPolicy support for sandbox isolation. Set the `networkPolicy` field on AgentRuntime or CodeInterpreter templates to enable a deny-all policy with DNS and router ingress allowed by default. Additional CIDR+port rules can be specified via the `ingress` and `egress` fields. Omitting the field leaves existing setups unaffected. Router selector and namespace are configurable via chart values.
```


hzxuzhonghu (commented):
## 代码审查

**总体评价：大型 feature PR，设计周到，实现较完整，但体量较大建议分步合并。**

### 亮点

1. **三种模式设计合理**：None（默认不变）/ Restricted（deny-all + DNS + 路由器放行）/ Custom（透传原生规则），向后兼容，用户可渐进采用。
2. **跨命名空间路由器 ingress 处理**：使用 NamespaceSelector + kubernetes.io/metadata.name，正确解决了同命名空间 PodSelector 失效的问题。
3. **GC 清理 bug 修复**：原来在 IsNotFound 时提前 return，会导致 NetworkPolicy 泄漏，此 PR 已修复。
4. **回滚简化**：由追踪 npCreated 标志改为"只要 networkPolicy != nil 就尝试删除"，也覆盖了"k8s 写入成功但返回错误"的边缘情况，更健壮。
5. **RBAC 补全**：ClusterRole 中添加了 networking.k8s.io/networkpolicies 权限，不遗漏。

### 已知问题（PR 作者已说明）

- warm-pool 路径的 NP 创建后 PodSelector 无法匹配（label 未传播），暂不生效，待后续 follow-up 修复。建议在代码中添加 TODO 注释，或在跟踪 issue 中明确记录。

### 建议

1. 建议为 NetworkPolicy 对象添加 OwnerReference，虽然显式清理已处理大多数路径，但 OwnerReference 可作为最终兜底保障。
2. Custom 模式下未做规则校验，恶意或格式错误的规则可能直接写入 k8s，建议增加基本的 Ingress/Egress 规则合法性校验。
3. 单元测试覆盖 NetworkPolicy 生命周期（创建/更新/回滚/GC）。

整体实现思路正确，主要风险点已在 PR 描述中说明。

hzxuzhonghu (commented):
Good design with three modes that maintain backward compatibility. The GC leak fix is a nice bonus. Please confirm behavior of NetworkPolicy for the warm-pool path before merging.

Sanchit2662:
Hey @hzxuzhonghu 
For the warm-pool path, the NP is created but the PodSelector (SandboxNameLabelKey: <sandboxName>) won't match anything. Here's why: warm-pool pods are pre-created   from the SandboxTemplate before any session exists, so they carry whatever labels the template defines , not our per-session SandboxNameLabelKey. The label only gets set explicitly in buildSandboxObject (the direct-sandbox path), where we control the pod template at creation time. In the warm-pool path we create a SandboxClaim and agent-sandbox binds a pre-warmed pod to it, but that binding doesn't patch our label onto the pod.                                                

Practically: the NP gets created, sits there with no matching pods, and does nothing. The sandbox runs unprotected in the warm-pool case even if Mode=Restricted. I've got a TODO in the code at the build site marking this.
                                                                                                                                                                       
Two possible fixes I'm aware of:                                                                                                                                      
1. Post-ready pod patch , after the sandbox reports ready, we know the pod name (from the SandboxPodNameAnnotation). We could patch the pod to add SandboxNameLabelKey and then create the NP. Needs pod patch RBAC, and it's a race-y two-step between pod running and NP creation.

2. Use whatever labels agent-sandbox applies at bind time , if agent-sandbox patches session-identifying labels onto the pod when binding a claim, we can use those   as the selector instead. I'd need to look at the agent-sandbox controller code to confirm what labels end up on a bound warm-pool pod.                               

My suggestion: merge this PR as-is since the direct-sandbox path works correctly, and track the warm-pool fix as a follow-up once we confirm what labels agent-sandbox applies. I'd rather not block the whole feature on the warm-pool edge case, but happy to dig into it more if you'd prefer it handled in this PR.

pkg/workloadmanager/handlers.go
```diff
@@ -140,37 +145,75 @@ func (s *Server) handleSandboxCreate(c *gin.Context, kind string) {
...
 			return
 		}
-		s.sandboxRollback(sandboxClaim, dynamicClient, sandbox, sandboxEntry)
+		s.sandboxRollback(sandboxClaim, dynamicClient, sandbox, networkPolicy, sandboxEntry)
 	}()
 
+	// Create NetworkPolicy if one was built from the template spec.
+	// Uses the workloadmanager's own client (not the user client) since NP is infra.
+	if networkPolicy != nil {
+		if err := createNetworkPolicy(ctx, s.k8sClient.clientset, networkPolicy); err != nil {
```
hzxuzhonghu: **Security: `Mode=Custom` in `CreateSandboxRequest` is a privilege-escalation vector when `EnableAuth=true`.**

The NetworkPolicy is created here using `s.k8sClient.clientset` — the workloadmanager's own elevated service account (which now has `networkpolicies` create/delete RBAC per this PR's manifest changes). The NP rules, however, come entirely from the user-supplied `sandboxReq.NetworkPolicy` field.

This means a user who can call the WM sandbox-create API but does **not** have `networkpolicies/create` permission in k8s can bypass that restriction by setting `networkPolicy.mode=Custom` with arbitrary egress rules. The practical impact: they can weaken the isolation the cluster operator intended (e.g., override a template-level `Restricted` policy with open egress rules that reach the k8s API server, cloud metadata endpoints, or internal services).

Suggested fixes:
1. **Disallow `Custom` mode in `CreateSandboxRequest`** when `EnableAuth=true`; reserve `Custom` for cluster-admin-controlled CRD templates only and allow only `None`/`Restricted` per-session overrides.
2. **SubjectAccessReview gate**: before creating the NP, perform a SAR check on behalf of the requesting user to verify they hold `networking.k8s.io/networkpolicies` `create` in the sandbox namespace.

pkg/workloadmanager/networkpolicy_builder.go
```diff
@@ -0,0 +1,149 @@
...
+
+func buildRestrictedIngressRules(spec *runtimev1alpha1.SandboxNetworkPolicy, routerSelector map[string]string, routerNamespace string) []networkingv1.NetworkPolicyIngressRule {
+	// Router ingress is always allowed in Restricted mode. NamespaceSelector is set
+	// so the rule works even when the router runs in a namespace distinct from the
+	// sandbox (the common production layout: router in agentcube-system, sandboxes
+	// in user namespaces).
+	routerPeer := networkingv1.NetworkPolicyPeer{
+		PodSelector: &metav1.LabelSelector{MatchLabels: routerSelector},
+	}
+	if routerNamespace != "" {
```
hzxuzhonghu: When `routerNamespace` is an empty string, `NamespaceSelector` is **not** added to the peer. Kubernetes NetworkPolicy semantics for a peer with only `PodSelector` (no `NamespaceSelector`): *match pods in the **same namespace** as the NetworkPolicy*.

If the router is in a different namespace and `routerNamespace` is empty, the router ingress rule silently becomes a same-namespace-only rule — cross-namespace router traffic is blocked, defeating the whole point of this cross-namespace logic.

In production this is guarded by `NewServer` defaulting `RouterNamespace` to `IdentitySecretNamespace`, so the empty case never fires at runtime. But the function itself is silently wrong for empty input, and any test or future caller that passes `""` will get incorrect policy without any signal. Consider either panic/returning an error for empty `routerNamespace`, or at minimum adding a comment: `// routerNamespace must be non-empty; an empty value restricts the rule to the sandbox's own namespace`.

pkg/workloadmanager/networkpolicy_builder.go
```diff
@@ -0,0 +1,149 @@
...
+		},
+	}
+
+	switch spec.Mode {
+	case runtimev1alpha1.NetworkPolicyModeRestricted:
+		np.Spec.Ingress = buildRestrictedIngressRules(spec, routerSelector, routerNamespace)
+		np.Spec.Egress = buildRestrictedEgressRules(spec)
+	case runtimev1alpha1.NetworkPolicyModeCustom:
+		if spec.Custom != nil {
+			np.Spec.Ingress = spec.Custom.Ingress
```
hzxuzhonghu: `PolicyTypes: [Ingress, Egress]` is set unconditionally (lines 53-56) for **both** `Restricted` and `Custom` modes. In `Custom` mode, `np.Spec.Ingress` is set directly from `spec.Custom.Ingress`. If the caller provides only egress rules and leaves `Custom.Ingress` nil, the resulting NP has `PolicyTypes: [Ingress]` with **no ingress rules — deny-all ingress**, including from the router.

Unlike `Restricted` mode, `Custom` mode does **not** automatically inject the router ingress rule. This is not documented anywhere in the type or the mode description, so an operator writing:
```yaml
networkPolicy:
  mode: Custom
  custom:
    egress:
      - ports: [{port: 443}]
```
...would unknowingly block the router from reaching the sandbox and see requests time out silently.

Please add a note to `SandboxNetworkPolicyCustomRules` or `NetworkPolicyModeCustom` warning that callers must explicitly include any required ingress rules (including router ingress) when using Custom mode.

hzxuzhonghu (changes requested):
The overall design is solid — per-session override, Restricted-mode DNS egress defaults, cross-namespace router peer with NamespaceSelector, and GC/rollback NP cleanup are all well thought out. A few issues below.

**Also note dead code in `handleDeleteSandbox`**: `deleteSandboxClaim` and `deleteSandbox` in `k8s_client.go` already return `nil` for `IsNotFound` errors, so the `if apierrors.IsNotFound(err)` branches at lines ~322 and ~334 of `handlers.go` are **unreachable** — `err` can never be a NotFound error at that point. The log lines there will never fire. The dead branches should be removed to avoid confusing future readers who might assume NotFound could surface there.

pkg/workloadmanager/handlers.go
```diff
@@ -140,37 +145,75 @@ func (s *Server) handleSandboxCreate(c *gin.Context, kind string) {
...
+			return nil, api.NewInternalError(err)
```
hzxuzhonghu: **Security: per-session `Custom` mode allows RBAC bypass when `EnableAuth=true`.**

This call uses `s.k8sClient.clientset` (the workloadmanager's own elevated SA, which has `networking.k8s.io/networkpolicies` create RBAC per the manifest added in this PR). A user who calls this API endpoint with `networkPolicy.mode=Custom` can therefore create an arbitrary NetworkPolicy in the sandbox namespace, even if their own principal lacks that RBAC permission.

In a multi-tenant cluster where an operator sets `Mode=Restricted` to lock down sandbox egress (e.g., block the metadata endpoint or internal services), any API caller can override that with `Mode=Custom` and open egress at will — effectively a privilege escalation.

Fix options:
1. **Disallow `Custom` in per-session overrides.** Only allow it in the operator-controlled template spec; session overrides can only set `Mode=Restricted/None` and add `allowedEgress` entries.
2. **SubjectAccessReview gate.** Before using the elevated client, perform an `authorizationv1.SubjectAccessReview` to confirm the requesting user has `networking.k8s.io/networkpolicies` create permission in the target namespace.

pkg/workloadmanager/networkpolicy_builder.go
```diff
@@ -0,0 +1,149 @@
...
+			np.Spec.Egress = spec.Custom.Egress
```
hzxuzhonghu: **`PolicyTypes: [Ingress, Egress]` is always set; nil `Custom.Ingress` = deny-all ingress, including from the router.**

`PolicyTypes` is set unconditionally to `[Ingress, Egress]` for all non-None modes. In `Custom` mode, if `spec.Custom.Ingress` is nil (the user only wants to configure egress), `np.Spec.Ingress` is nil here. Per k8s NetworkPolicy semantics, `PolicyType=Ingress` with no ingress rules means **deny all ingress** — including from the router. The sandbox becomes completely unreachable.

This affects a user who passes `{mode: Custom, custom: {egress: [...]}}` expecting to add egress allow rules while keeping normal inbound traffic.

The kubebuilder validation only requires `custom != null`, so `{mode: Custom, custom: {}}` is accepted and silently creates a deny-all policy.

Fix: add a warning to the `SandboxNetworkPolicyCustomRules` godoc:
```
// When mode is Custom, ALL ingress is denied unless rules are explicitly
// specified here. The router ingress rule injected by Restricted mode is NOT
// automatic — include it in Ingress if sandbox reachability is required.
```

pkg/workloadmanager/networkpolicy_builder.go
```diff
@@ -0,0 +1,149 @@
...
+func buildRestrictedIngressRules(spec *runtimev1alpha1.SandboxNetworkPolicy, routerSelector map[string]string, routerNamespace string) []networkingv1.NetworkPolicyIngressRule {
+	// Router ingress is always allowed in Restricted mode. NamespaceSelector is set
+	// so the rule works even when the router runs in a namespace distinct from the
+	// sandbox (the common production layout: router in agentcube-system, sandboxes
+	// in user namespaces).
+	routerPeer := networkingv1.NetworkPolicyPeer{
+		PodSelector: &metav1.LabelSelector{MatchLabels: routerSelector},
+	}
+	if routerNamespace != "" {
+		routerPeer.NamespaceSelector = &metav1.LabelSelector{
```
hzxuzhonghu: **Silent wrong semantics when `routerNamespace` is empty.**

When `routerNamespace == ""`, `NamespaceSelector` is skipped. Per k8s NetworkPolicy semantics, a `NetworkPolicyPeer` with only `PodSelector` and no `NamespaceSelector` selects pods in the **same namespace as the NetworkPolicy** — not all namespaces.

So if `routerNamespace` is empty, the rule only allows router pods co-located in the sandbox namespace. If the router is in a different namespace (the common multi-tenant layout the comment above describes), it silently cannot reach the sandbox — exactly the case the `NamespaceSelector` was added to prevent.

`server.go/NewServer` ensures `RouterNamespace` always defaults to at least `IdentitySecretNamespace`, so production is safe today. But the function itself has incorrect semantics for the empty case, making it a hazard for tests and any future direct callers.

Fix: guard at the top of `buildRestrictedIngressRules`:
```go
if routerNamespace == "" {
    // A missing namespace selector restricts the rule to the sandbox namespace only,
    // defeating cross-namespace router ingress. Callers must supply a non-empty value.
    klog.Fatalf("routerNamespace must not be empty")
}
```
or return an error up the call chain.

hzxuzhonghu (changes requested):
The overall design is solid — per-session override semantics, rollback logic, GC cleanup, and the cross-namespace router NamespaceSelector are all well-thought-out.


**Dead code in `handleDeleteSandbox` (lines ~322/334 — outside this diff's hunks):** The `if apierrors.IsNotFound(err)` branches after `deleteSandboxClaim`/`deleteSandbox` calls are unreachable. Both helpers already swallow `IsNotFound` and return `nil` for it. Any error that reaches those `if err != nil` guards is always a real, non-NotFound error. The branches are harmless today but would cause a silent behavior change if the helpers ever stop suppressing NotFound. Fix: remove the dead `IsNotFound` branches, or remove NotFound-suppression from the helpers and handle it at each call site.

pkg/apis/runtime/v1alpha1/networkpolicy_types.go
```diff
@@ -0,0 +1,113 @@
...
+// SandboxNetworkPolicy defines the network isolation policy applied to each sandbox pod.
+// +kubebuilder:validation:XValidation:rule="self.mode != 'Custom' || (has(self.custom) && self.custom != null)",message="custom must be set when mode is Custom"
+type SandboxNetworkPolicy struct {
+	// Mode controls the network policy enforcement level.
+	// "None" (default): no NetworkPolicy is created.
+	// "Restricted": deny-all baseline; DNS egress and router ingress are always allowed.
+	// "Custom": use the raw Ingress/Egress rules in the Custom field.
+	// +kubebuilder:default=None
+	// +optional
+	Mode NetworkPolicyMode `json:"mode,omitempty"`
```
hzxuzhonghu: Why do you need this field, i thin none can be represented with unset policy

`Custom` can be represented by `AllowedIngress ` `AllowedEgress`

`Restricted` should be like `None`

pkg/apis/runtime/v1alpha1/networkpolicy_types.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// AllowedEgress lists additional egress allow rules applied when Mode=Restricted.
+	// +optional
+	AllowedEgress []SandboxEgressRule `json:"allowedEgress,omitempty"`
```
hzxuzhonghu:
```suggestion
	Egress []SandboxEgressRule `json:"allowedEgress,omitempty"`
```

pkg/apis/runtime/v1alpha1/networkpolicy_types.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// AllowDNS controls whether a DNS egress rule (port 53 UDP/TCP) is added when
+	// Mode=Restricted. Defaults to true.
+	// +optional
+	AllowDNS *bool `json:"allowDNS,omitempty"`
```
hzxuzhonghu: not needed?

pkg/apis/runtime/v1alpha1/networkpolicy_types.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// AllowedIngress lists additional ingress allow rules when Mode=Restricted.
+	// Router ingress is always permitted in Restricted mode regardless of this field.
+	// +optional
+	AllowedIngress []SandboxIngressRule `json:"allowedIngress,omitempty"`
```
hzxuzhonghu:
```suggestion
	Ingress []SandboxIngressRule `json:"ingress,omitempty"`
```

pkg/apis/runtime/v1alpha1/networkpolicy_types.go
```diff
@@ -0,0 +1,113 @@
...
+type SandboxIngressRule struct {
+	// CIDRs is the list of source CIDR blocks.
+	// Each entry must be a valid CIDR notation (e.g. "10.0.0.0/8" or "2001:db8::/32").
+	// +optional
+	// +listType=atomic
+	// +kubebuilder:validation:items:Pattern=`^(\d{1,3}\.){3}\d{1,3}\/\d{1,2}$|^[0-9a-fA-F:]+\/\d{1,3}$`
+	CIDRs []string `json:"cidrs,omitempty"`
+	// Ports restricts the rule to these destination ports. An empty list matches all ports.
+	// +optional
+	Ports []SandboxNetworkPolicyPort `json:"ports,omitempty"`
```
hzxuzhonghu, line 49: ?? I think the comment is wrong , copied from SandboxEgressRule

pkg/apis/runtime/v1alpha1/networkpolicy_types.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// SandboxIngressRule describes an ingress allow rule for a sandbox pod.
+type SandboxIngressRule struct {
```
hzxuzhonghu: This struct is same as SandboxEgressRule, please check  k8s `NetworkPolicySpec` api

pkg/workloadmanager/handlers.go
```diff
@@ -140,37 +145,75 @@ func (s *Server) handleSandboxCreate(c *gin.Context, kind string) {
...
+		return nil, api.NewInternalError(fmt.Errorf("store sandbox placeholder failed: %v", err))
 	}
 
-	if sandboxClaim != nil {
-		if err := createSandboxClaim(ctx, dynamicClient, sandboxClaim); err != nil {
-			err = api.NewInternalError(fmt.Errorf("create sandbox claim %s/%s failed: %v", sandboxClaim.Namespace, sandboxClaim.Name, err))
-			return nil, err
-		}
-	} else {
-		if _, err := createSandbox(ctx, dynamicClient, sandbox); err != nil {
```
hzxuzhonghu, line 214: why change it

Sanchit2662: This was extracted from createSandbox to get its cyclomatic complexity under the gocyclo lint limit of 15 , it was hitting 16 before. The logic is exactly the same, just pulled out into its own helper. The early return is just how I naturally wrote it in Go, happy to match whatever style you prefer here.

Sanchit2662:
Hi @hzxuzhonghu , thanks for the thorough review! I've addressed all the feedback

pkg/workloadmanager/handlers.go
```diff
@@ -140,37 +144,75 @@ func (s *Server) handleSandboxCreate(c *gin.Context, kind string) {
...
 		if !needRollbackSandbox {
 			return
 		}
-		s.sandboxRollback(sandboxClaim, dynamicClient, sandbox, sandboxEntry)
+		s.sandboxRollback(sandboxClaim, dynamicClient, sandbox, networkPolicy, sandboxEntry)
 	}()
 
+	// Create NetworkPolicy if one was built from the template spec.
+	// Uses the workloadmanager's own client (not the user client) since NP is infra.
+	if networkPolicy != nil {
```
acsoto, line 339: Can we create the NetworkPolicy before creating the Sandbox/SandboxClaim, or otherwise block pod startup until the policy exists? Creating the workload first leaves a window where untrusted sandbox code can run before the deny-all policy is applied.

pkg/workloadmanager/workload_builder.go
```diff
@@ -388,7 +400,12 @@ func buildSandboxByCodeInterpreter(namespace string, codeInterpreterName string,
 			simpleSandbox.Spec.Lifecycle.ShutdownTime = &shutdownTime
 		}
 		sandboxEntry.Kind = types.SandboxClaimsKind
-		return simpleSandbox, sandboxClaim, sandboxEntry, nil
+		np := buildNetworkPolicy(sandboxName, namespace, effectiveNetworkPolicy(sessionNetworkPolicy, codeInterpreterObj.Spec.Template.NetworkPolicy), routerSelector, routerNamespace)
```
acsoto: Can we avoid creating a NetworkPolicy for the warm-pool path until it actually selects the bound pod? As written, this creates a policy object but the TODO says the pod does not carry `SandboxNameLabelKey`, so users can enable isolation and still get no enforcement.

Sanchit2662: I've addressed this ;  Warm-pool pods don't carry SandboxNameLabelKey, so the NP's PodSelector would never match them , creating the object only gave a false sense of isolation. I've removed NP creation from the warm-pool path entirely and added a warning log when a policy is configured but can't be enforced. Proper enforcement   will need agent-sandbox to propagate claim labels onto the bound pod; left a note in the code for that.

pkg/workloadmanager/handlers.go
Sanchit2662: Done , NP is now created before the Sandbox/SandboxClaim so the deny-all is in place before the CNI wires up the pod's network interface. The OwnerRef is patched onto the NP after the sandbox UID is known.

Sanchit2662:
Hey @hzxuzhonghu @acsoto , just checking in any updates on this PR or are there any further changes needed.

hzxuzhonghu:
@Sanchit2662 Would suggest first resolve the ci failure, and all he comments above. Then we can move faster

manifests/charts/base/crds/runtime.agentcube.volcano.sh_agentruntimes.yaml
```diff
@@ -63,6 +63,93 @@ spec:
...
+                        items:
+                          description: SandboxNetworkPolicyRule describes a single
+                            ingress or egress allow rule for a sandbox pod.
+                          properties:
+                            cidrs:
+                              description: |-
+                                CIDRs is the list of CIDR blocks for this rule (source for ingress, destination for egress).
+                                Each entry must be valid CIDR notation (e.g. "10.0.0.0/8" or "2001:db8::/32").
+                              items:
+                                pattern: ^(\d{1,3}\.){3}\d{1,3}\/\d{1,2}$|^[0-9a-fA-F:]+\/\d{1,3}$
```
hzxuzhonghu: The IPv4 part of this regex (`(\d{1,3}\.){3}\d{1,3}\/\d{1,2}`) accepts octets up to 999 — so `999.999.999.999/0` would pass CRD validation. Not a security issue since K8s will reject it at the NetworkPolicy level, but it might be confusing to users who get a mysterious downstream error instead of a clear validation message. A stricter octet bound like `(25[0-5]|2[0-4]\d|[01]?\d\d?)` would catch it earlier.

cmd/workload-manager/main.go
```diff
@@ -95,6 +103,11 @@ func main() {
 		os.Exit(1)
```
hzxuzhonghu, line 120: Malformed pairs in `--router-selector` are silently skipped (only an empty-result warning is logged). If someone passes `app=agentcube-router,badpair` they'd get `app=agentcube-router` silently, which could be confusing. Maybe worth a `klog.Warningf` per skipped pair so it's visible in logs?

hzxuzhonghu (commented):
Solid feature PR — the three-mode design (None/Restricted/Custom) is clean and the cross-namespace router ingress fix with NamespaceSelector is important for real deployments. The `--router-selector` / `--router-namespace` flags are a nice way to keep this configurable. Two things worth looking at below.

FAUST-BENCHOU:
friendly ping @Sanchit2662 .Are u still working on it?

Sanchit2662:
@FAUST-BENCHOU , yes i am working on it ,I will push the changes shorlty.

FAUST-BENCHOU:
Thanks btw dont forget to rebase first

Sanchit2662:
Hi @hzxuzhonghu , I have made the required change.

hzxuzhonghu:
Please resolve the conflicts, i will take a look

Sanchit2662:
I resolved the conflicts.

FAUST-BENCHOU:
make sure ci green first

Sanchit2662:
Hey @hzxuzhonghu , this is ready for another look whenever you get a chance. The CI is green an, PR description has been updated to match what actually ships and the conflicts have been resolved too. Since your last review I've rebased everything onto main as a single commit and worked through all the open comments. Happy to address anything else!

PR #292, opened 2026-04-21
fix: propagate ctx into waitForCacheSync to respect caller cancellation

Storm1289:
## Summary

- Fix RunAndWaitForCacheSync to respect the provided context for timeout handling instead of using context.Background()
- Ensures cache sync wait exits early when the parent context is cancelled (e.g., during graceful shutdown)

## What

RunAndWaitForCacheSync accepted a ctx (context.Context) parameter but incorrectly derived its timeout from context.Background():

// before
ctxTimeout, cancel := context.WithTimeout(context.Background(), 1*time.Minute)

// after
ctxTimeout, cancel := context.WithTimeout(ctx, 1*time.Minute)

## Why

Cancelling the parent context (for example, on SIGTERM during graceful shutdown) had no effect on the cache-sync wait. The function would block for the full 1-minute timeout regardless.

The same ctx was already correctly passed to ifm.run(ctx.Done()) earlier, making this inconsistency evident. This change ensures consistent and expected context propagation.

## Changes

- Updated RunAndWaitForCacheSync to derive timeout from the passed ctx
- Preserves existing timeout behavior while allowing early cancellation
- Aligns context usage with ifm.run(ctx.Done())

## Test Plan

- [x] Added TestRunAndWaitForCacheSync_RespectsContextCancellation in informers_test.go

### Test Details

- Uses a neverSyncedInformer stub (HasSynced always returns false) to keep sync pending
- Cancels the parent context immediately
- Asserts the function returns within 2 seconds (well before the 1-minute timeout)

## Impact

- Improves graceful shutdown behavior
- Prevents unnecessary blocking during termination
- No breaking changes (backward compatible)

### DIFF
--- a/pkg/workloadmanager/informers.go
+++ b/pkg/workloadmanager/informers.go
@@ -83,12 +83,21 @@ func (ifm *Informers) run(stopCh <-chan struct{}) {
 
 func (ifm *Informers) waitForCacheSync(ctx context.Context) error {
 	if !cache.WaitForCacheSync(ctx.Done(), ifm.AgentRuntimeInformer.HasSynced) {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
 		return fmt.Errorf("timed out waiting for %v caches to sync", AgentRuntimeGVR)
 	}
 	if !cache.WaitForCacheSync(ctx.Done(), ifm.CodeInterpreterInformer.HasSynced) {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
 		return fmt.Errorf("timed out waiting for %v caches to sync", CodeInterpreterGVR)
 	}
 	if !cache.WaitForCacheSync(ctx.Done(), ifm.PodInformer.HasSynced) {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
 		return fmt.Errorf("timed out waiting for pod informer cache to sync")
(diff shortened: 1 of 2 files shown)


hzxuzhonghu (commented):
LGTM

hzxuzhonghu (commented):
LGTM. One-line fix propagating context into waitForCacheSync correctly resolves the blocking issue during graceful shutdown. Interface abstraction is clean and tests are thorough.

pkg/workloadmanager/informers.go
```diff
@@ -49,11 +48,16 @@ var (
 	}
 )
 
+// informerStarter is the subset of SharedInformerFactory used by Informers.
+type informerStarter interface {
```
hzxuzhonghu: you donot need this wrapper, i guess you can initiate a SharedInformerFactory with k8s fake client

Storm1289:
Hi @hzxuzhonghu, thanks for the review .
I have removed the informerStarter wrapper and replaced it with fake.NewSimpleClientset() +informers.NewSharedInformerFactory(fakeClient, 0) so the test uses a real factory backed by a k8s fake client. No custom abstraction needed.

hzxuzhonghu (approved):
/lgtm

Issue #243, opened 2026-03-12
LFX 2026 Term 1 Project - Establishing AgentCube's Authentication and Authorization Capabilities


mahil-2040:
# Establishing AgentCube's Authentication and Authorization Capabilities

*V=volcano-sh/agentcube: Establish authentication and authorization (2026 Term 1)**

AgentCube currently lacks a unified authentication and authorization model. This issue tracks the implementation of:

- **Internal workload identity** using X.509 mTLS between Router, WorkloadManager, and PicoD (SPIRE recommended, file-based certs also supported).
- **Hybrid Sandbox Authentication** using Application-layer JWTs for latency-critical code interpreters.
- **External user authentication** using Keycloak (OIDC/OAuth2) at the Router.
- **Role-based authorization** using Keycloak's built-in RBAC.
- **Resource-Level Access Control (RLAC)** for strict tenant isolation.
- **Cloud provider federation** via Keycloak identity brokering (stretch goal).

Design proposal: [auth-proposal.md](https://github.com/volcano-sh/agentcube/blob/main/docs/design/auth-proposal.md)

### Internal Workload Authentication (X.509 mTLS & JWT)

- [x] **1. Auth design proposal**
  Publish the design document covering X.509 mTLS (with SPIRE and file-based certificate modes), Keycloak authentication, Keycloak RBAC authorization, and cloud provider federation.

- [x] **2. SPIRE infrastructure deployment**
  Add SPIRE Server (StatefulSet), Agent (DaemonSet), and Controller Manager (Deployment) manifests to `agentcube-system`. Create ServiceAccounts, RBAC, ConfigMaps for server/agent configuration, and `ClusterSPIFFEID` CRDs for declarative workload registration. SPIFFE IDs follow the Istio convention (`spiffe://<trust-domain>/ns/<namespace>/sa/<service-account>`).

- [ ] **3. Certificate source abstraction and dynamic reloading**
  Add `--mtls-cert-file`, `--mtls-key-file`, and `--mtls-ca-file` flags to Router, WorkloadManager, and PicoD. Implement a unified TLS configuration loader using Go's standard `crypto/tls` package. Configure `GetCertificate` and `GetClientCertificate` callbacks for dynamic zero-downtime certificate reloading. For SPIRE mode, the [spiffe-helper](https://github.com/spiffe/spiffe-helper) sidecar writes SVIDs to disk as PEM files, consumed by the same code path. Unit tests for the configuration loader.

[...]

PR #293, opened 2026-04-21
Feat: support spiffee mtls between router and workloadmanager

mahil-2040:
**What type of PR is this?**

/kind feature

### Description
This PR implements the foundational mTLS certificate abstraction layer (`pkg/mtls`) and actively wires the internal HTTP/gRPC listeners across the AgentCube architecture. Following the secure workload identity framework outlined in `auth-proposal.md`, this PR establishes the zero-trust, mutually authenticated TLS connections between the **Router** and **WorkloadManager** components. 

*(Note: Earlier PicoD mTLS/sidecar wiring has been reverted/decoupled from this PR to isolate scope cleanly to control-plane infrastructure. PicoD will use JWT application-layer logic for now).*

This officially completes Phase 1: Internal Workload Authentication (SPIRE) of the Auth Proposal.

### Core Changes
1. **Dynamic Zero-Downtime Rotation & Auto-Detection**
Introduced a highly robust, fsnotify-driven `CertWatcher` capable of hot-reloading certificates and dynamic CA rotation from disk without dropping active connections. mTLS configuration now uses auto-detection.

2. **Strict SPIFFE Identity Verification**
Added `pkg/mtls/spiffeid.go` with hardcoded, secure SPIFFE identity constants. 
- The WorkloadManager server is hard-wired to strictly reject connections that do not present the `RouterSPIFFEID`.
- The Router client cryptographically verifies it is communicating with the authentic `WorkloadManagerSPIFFEID` before proxying requests.

3. **Component Wiring & Resilient Startup**
Wired the `--tls-cert`, `--tls-key`, and `--tls-ca` flags directly into the Router and WorkloadManager Config structs. Added a pre-flight wait loop (`waitForMTLSFiles`) during binary startup to patiently wait for SPIRE agent to provision the files, ensuring no crash-loops on startup.

4. **E2E Testing & Namespace Decoupling**
- Integrated E2E tests for when mTLS is enabled.
- Split E2E testing namespaces: Introduced `WORKLOAD_NAMESPACE` distinct from `AGENTCUBE_NAMESPACE` (system namespace) to appropriately separate system services from the sandboxed CodeInterpreter workflows. 
- Updated `run_e2e.sh` and Python fixtures to strictly respect workload boundaries.

5. **Helm Charts Updates**
Updated the base template manifests in `manifests/charts/base/templates/` to seamlessly reflect current mTLS setups and sidecar behaviors for the Router and WorkloadManager.

### Verification
- All AgentCube binaries build successfully with the new CLI flags.
- `pkg/mtls` test suite passes (including rotational tests, intermediate cert validation logic, and auto-detection).
- E2E tests fully pass with `MTLS_ENABLED=true` securely routing `Router -> WorkloadManager`.
- Verified component resilience: Router and WorkloadManager wait patiently and securely bootstrap when SPIRE injects certs dynamically.

### Which issue(s) this PR fixes:
Fixes Part of #243

### Special notes for your reviewer:
PicoD sidecar builder logic and mode toggles (`--picod-auth-mode`) were reverted to keep this PR heavily focused and stable on the Router <-> WorkloadManager boundary. If multi-cluster or configurable trust domains are needed in the future, the hardcoded constants in `spiffeid.go` should be exposed via CLI flags.

### Does this PR introduce a user-facing change?:
yes

```release-note
Implemented foundational zero-trust Workload Identity via SPIRE. The Router and WorkloadManager components now support mutually authenticated TLS (mTLS) with strict SPIFFE ID enforcement and dynamic certificate rotation. Helm charts and deployment testing mechanisms have also been compartmentalized strictly into system vs. workload deployments.
```

### DIFF
--- a/cmd/picod/main.go
+++ b/cmd/picod/main.go
@@ -17,7 +17,11 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"flag"
+	"os"
+	"os/signal"
+	"syscall"
 
 	"k8s.io/klog/v2"
 
@@ -37,10 +41,32 @@ func main() {
 		Workspace: *workspace,
 	}
 
-	// Create and start server
+	// Create server
 	server := picod.NewServer(config)
 
-	if err := server.Run(); err != nil {
-		klog.Fatalf("Failed to start server: %v", err)
+	// Setup signal handling with context cancellation
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	defer cancel()
+
+	// Start PicoD server in goroutine
+	errCh := make(chan error, 1)
+	go func() {
+		klog.Infof("Starting PicoD server on port %d", *port)
+		if err := server.Start(ctx); err != nil {
+			errCh <- err
+		}
+		close(errCh)
+	}()
+
+	// Wait for signal or fatal error
+	select {
+	case <-ctx.Done():
+		klog.Info("Received shutdown signal, shutting down gracefully...")
+		cancel()
+		<-errCh
+	case err := <-errCh:
+		klog.Fatalf("Server error: %v", err)
 	}
+
+	klog.Info("PicoD server stopped")
 }
--- a/cmd/router/main.go
+++ b/cmd/router/main.go
@@ -19,12 +19,14 @@ package main
 import (
 	"context"
 	"flag"
+	"fmt"
 	"os"
 	"os/signal"
 	"syscall"
 
 	"k8s.io/klog/v2"
 
+	"github.com/volcano-sh/agentcube/pkg/mtls"
 	"github.com/volcano-sh/agentcube/pkg/router"
 )
 
@@ -32,10 +34,16 @@ func main() {
 	var (
 		port                  = flag.String("port", "8080", "Router API server port")
 		enableTLS             = flag.Bool("enable-tls", false, "Enable TLS (HTTPS)")
-		tlsCert               = flag.String("tls-cert", "", "Path to TLS certificate file")
-		tlsKey                = flag.String("tls-key", "", "Path to TLS key file")
+		tlsCert               = flag.String("tls-cert", "", "Path to TLS certificate file for the Router listener")
+		tlsKey                = flag.String("tls-key", "", "Path to TLS key file for the Router listener")
 		debug                 = flag.Bool("debug", false, "Enable debug mode")
 		maxConcurrentRequests = flag.Int("max-concurrent-requests", 1000, "Maximum number of concurrent requests that a router server can handle (0 = unlimited)")
+
+		// mTLS client certificates used by the Router to authenticate with upstream WorkloadManager.
+		// These are distinct from --tls-cert/--tls-key which serve the Router's own HTTPS listener.
+		mtlsCert = flag.String("mtls-cert", "", "Path to mTLS client certificate for upstream WorkloadManager connections")
+		mtlsKey  = flag.String("mtls-key", "", "Path to mTLS client key for upstream WorkloadManager connections")
+		mtlsCA   = flag.String("mtls-ca", "", "Path to mTLS CA bundle for verifying upstream WorkloadManager identity")
 	)
 
 	// Initialize klog flags
@@ -43,6 +51,21 @@ func main() {
 
 	// Parse command line flags
 	flag.Parse()
+
+	tlsConfig := mtls.Config{
+		CertFile: *mtlsCert,
+		KeyFile:  *mtlsKey,
+		CAFile:   *mtlsCA,
+	}
+	if err := validateMTLSFlags(tlsConfig); err != nil {
+		klog.Fatalf("Invalid mTLS configuration: %v", err)
+	}
+	if tlsConfig.CertFile != "" && tlsConfig.KeyFile != "" && tlsConfig.CAFile != "" {
+		klog.Infof("Waiting for Router mTLS cert/key/CA files")
+		if err := mtls.WaitForCertificateFiles(tlsConfig, mtls.DefaultCertificateFileWaitTimeout); err != nil {
+			klog.Fatalf("Invalid mTLS configuration: %v", err)
+		}
+	}
 
 	// Create Router API server configuration
 	config := &router.Config{
@@ -52,6 +75,7 @@ func main() {
 		TLSCert:               *tlsCert,
 		TLSKey:                *tlsKey,
 		MaxConcurrentRequests: *maxConcurrentRequests,
+		MTLSConfig:            tlsConfig,
 	}
 
 	// Create Router API server
@@ -87,4 +111,14 @@ func main() {
 	}
 
 	klog.Info("Router server stopped")
+}
+
+func validateMTLSFlags(cfg mtls.Config) error {
+	if cfg.CertFile == "" && cfg.KeyFile == "" && cfg.CAFile == "" {
+		return nil
+	}
+	if cfg.CertFile == "" || cfg.KeyFile == "" || cfg.CAFile == "" {
+		return fmt.Errorf("mtls-cert, mtls-key, and mtls-ca must all be specified together")
+	}
+	return nil
 }
--- a/cmd/workload-manager/main.go
+++ b/cmd/workload-manager/main.go
@@ -36,6 +36,7 @@ import (
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
 	runtimev1alpha1 "github.com/volcano-sh/agentcube/pkg/apis/runtime/v1alpha1"
+	"github.com/volcano-sh/agentcube/pkg/mtls"
 	"github.com/volcano-sh/agentcube/pkg/workloadmanager"
 )
 
@@ -57,7 +58,8 @@ func main() {
 		enableTLS        = flag.Bool("enable-tls", false, "Enable TLS (HTTPS)")
 		tlsCert          = flag.String("tls-cert", "", "Path to TLS certificate file")
 		tlsKey           = flag.String("tls-key", "", "Path to TLS key file")
-		enableAuth       = flag.Bool("enable-auth", false, "Enable Authentication")
+		tlsCA            = flag.String("tls-ca", "", "Path to TLS CA certificate file")
+		enableAuth       = flag.Bool("enable-auth", false, "Enable authorization")
 	)
 
 	// Initialize klog flags
@@ -65,6 +67,21 @@ func main() {
 
 	// Parse command line flags
 	flag.Parse()
+
+	tlsConfig := mtls.Config{
+		CertFile: *tlsCert,
+		KeyFile:  *tlsKey,
+		CAFile:   *tlsCA,
+	}
+	if err := validateTLSFlags(*enableTLS, tlsConfig); err != nil {
+		klog.Fatalf("Invalid TLS configuration: %v", err)
+	}
+	if tlsConfig.CertFile != "" && tlsConfig.KeyFile != "" && tlsConfig.CAFile != "" {
+		klog.Infof("Waiting for WorkloadManager mTLS cert/key/CA files")
+		if err := mtls.WaitForCertificateFiles(tlsConfig, mtls.DefaultCertificateFileWaitTimeout); err != nil {
+			klog.Fatalf("Invalid mTLS configuration: %v", err)
+		}
+	}
 
 	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
 
@@ -103,6 +120,7 @@ func main() {
 		TLSCert:          *tlsCert,
 		TLSKey:           *tlsKey,
 		EnableAuth:       *enableAuth,
+		MTLSConfig:       tlsConfig,
 	}
 
 	// Create and initialize API server
@@ -179,5 +197,18 @@ func setupControllers(mgr ctrl.Manager, sandboxReconciler *workloadmanager.Sandb
 		return fmt.Errorf("unable to create codeinterpreter controller: %w", err)
 	}
 
+	return nil
+}
+
+func validateTLSFlags(enableTLS bool, cfg mtls.Config) error {
+	if cfg.CAFile != "" {
+		if cfg.CertFile == "" || cfg.KeyFile == "" {
+			return fmt.Errorf("tls-cert, tls-key, and tls-ca must all be specified together for mTLS")
+		}
+		return nil
+	}
+	if enableTLS && (cfg.CertFile == "" || cfg.KeyFile == "") {
+		return fmt.Errorf("both --tls-cert and --tls-key must be provided when --enable-tls is set")
+	}
 	return nil
 }
--- a/manifests/charts/base/templates/agentcube-router.yaml
+++ b/manifests/charts/base/templates/agentcube-router.yaml
@@ -22,6 +22,10 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- if .Values.spire.enabled }}
+      securityContext:
+        fsGroup: 1000
+      {{- end }}
       containers:
         {{- if .Values.spire.enabled }}
         - name: spiffe-helper
@@ -64,12 +68,21 @@ spec:
               value: {{ .Values.redis.password | quote }}
             - name: WORKLOAD_MANAGER_URL
               value: "http://workloadmanager.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.workloadmanager.service.port }}"
+            {{- if .Values.spire.enabled }}
+            - name: AGENTCUBE_SPIFFE_TRUST_DOMAIN
+              value: {{ .Values.spire.trustDomain | quote }}
+            {{- end }}
             {{- with .Values.router.extraEnv }}
             {{- toYaml . | nindent 12 }}
             {{- end }}
           args:
             - --port={{ .Values.router.service.targetPort }}
             - --debug
+            {{- if .Values.spire.enabled }}
+            - --mtls-cert={{ .Values.spire.spiffeHelper.certDir }}/{{ .Values.spire.spiffeHelper.certFileName }}
+            - --mtls-key={{ .Values.spire.spiffeHelper.certDir }}/{{ .Values.spire.spiffeHelper.keyFileName }}
+            - --mtls-ca={{ .Values.spire.spiffeHelper.certDir }}/{{ .Values.spire.spiffeHelper.bundleFileName }}
+            {{- end }}
           resources:
             {{- toYaml .Values.router.resources | nindent 12 }}
           livenessProbe:
--- a/manifests/charts/base/templates/spire/cluster-spiffe-ids.yaml
+++ b/manifests/charts/base/templates/spire/cluster-spiffe-ids.yaml
@@ -1,7 +1,7 @@
 {{- if .Values.spire.enabled }}
 # Prerequisite: spire.spiffe.io CRDs must be present in the cluster before these resources can be created.
 # The spire-controller-manager sidecar registers them automatically on first boot.
-# For a fresh cluster, run: kubectl apply -f https://github.com/spiffe/spire-controller-manager/releases/download/v{{ .Values.spire.controllerManager.image.tag }}/crds.yaml
+# For a fresh cluster, run: kubectl apply -k "https://github.com/spiffe/spire-controller-manager/config/crd?ref=v{{ .Values.spire.controllerManager.image.tag }}"
 # Router registration
 apiVersion: spire.spiffe.io/v1alpha1
 kind: ClusterSPIFFEID
@@ -30,16 +30,4 @@ spec:
   namespaceSelector:
     matchLabels:
       kubernetes.io/metadata.name: {{ .Release.Namespace }}
-
----
-# PicoD (Sandbox) registration - namespace-agnostic
-apiVersion: spire.spiffe.io/v1alpha1
-kind: ClusterSPIFFEID
-metadata:
-  name: {{ .Release.Name }}-agentcube-sandbox
-spec:
-  spiffeIDTemplate: "spiffe://{{ .Values.spire.trustDomain }}/sa/{{ "{{ .PodSpec.ServiceAccountName }}" }}"
-  podSelector:
-    matchLabels:
-      app: picod
 {{- end }}
--- a/manifests/charts/base/templates/workloadmanager.yaml
+++ b/manifests/charts/base/templates/workloadmanager.yaml
@@ -20,6 +20,10 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- if .Values.spire.enabled }}
+      securityContext:
+        fsGroup: 1000
+      {{- end }}
       containers:
         {{- if .Values.spire.enabled }}
         - name: spiffe-helper
@@ -60,24 +64,43 @@ spec:
               value: {{ .Values.redis.password | quote }}
             - name: REDIS_ADDR
               value: {{ .Values.redis.addr | quote }}
+            {{- if .Values.spire.enabled }}
+            - name: AGENTCUBE_SPIFFE_TRUST_DOMAIN
+              value: {{ .Values.spire.trustDomain | quote }}
+            {{- end }}
             {{- with .Values.workloadmanager.extraEnv }}
             {{- toYaml . | nindent 12 }}
             {{- end }}
           args:
             - --port={{ .Values.workloadmanager.service.port }}
             - --runtime-class-name=
+            {{- if .Values.spire.enabled }}
+            - --tls-cert={{ .Values.spire.spiffeHelper.certDir }}/{{ .Values.spire.spiffeHelper.certFileName }}
+            - --tls-key={{ .Values.spire.spiffeHelper.certDir }}/{{ .Values.spire.spiffeHelper.keyFileName }}
+            - --tls-ca={{ .Values.spire.spiffeHelper.certDir }}/{{ .Values.spire.spiffeHelper.bundleFileName }}
+            {{- end }}
           resources:
             {{- toYaml .Values.workloadmanager.resources | nindent 12 }}
           livenessProbe:
+            {{- if .Values.spire.enabled }}
+            tcpSocket:
+              port: {{ .Values.workloadmanager.service.port }}
+            {{- else }}
             httpGet:
               path: /health
               port: {{ .Values.workloadmanager.service.port }}
+            {{- end }}
             initialDelaySeconds: 10
             periodSeconds: 10
           readinessProbe:
+            {{- if .Values.spire.enabled }}
+            tcpSocket:
+              port: {{ .Values.workloadmanager.service.port }}
+            {{- else }}
             httpGet:
               path: /health
               port: {{ .Values.workloadmanager.service.port }}
+            {{- end }}
             initialDelaySeconds: 5
             periodSeconds: 5
       {{- if .Values.spire.enabled }}
--- a/pkg/mtls/config.go
+++ b/pkg/mtls/config.go
@@ -1,0 +1,29 @@
+/*
+Copyright The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mtls
+
+// Config holds the mTLS certificate file paths for a component.
+// The code is certificate-source agnostic — it simply loads from the provided paths,
+// regardless of whether SPIRE, cert-manager, or a self-signed CA provisioned them.
+type Config struct {
+	// CertFile is the path to the mTLS certificate (PEM).
+	CertFile string
+	// KeyFile is the path to the mTLS private key (PEM).
+	KeyFile string
+	// CAFile is the path to the mTLS CA bundle for peer verification (PEM).
+	CAFile string
+}
--- a/pkg/mtls/loader.go
+++ b/pkg/mtls/loader.go
@@ -1,0 +1,186 @@
+/*
+Copyright The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mtls
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+)
+
+// LoadServerConfig returns a tls.Config for a server that requires mTLS.
+// Certs and the CA bundle are loaded dynamically via CertWatcher to support
+// zero-downtime rotation of both identity certificates and trust bundles.
+// Any client presenting a valid certificate signed by the trusted CA is accepted.
+// Authorization should be handled at the application layer, not at the TLS level.
+func LoadServerConfig(cfg *Config) (*tls.Config, *CertWatcher, error) {
+	if cfg == nil {
+		return nil, nil, fmt.Errorf("server TLS config: mtls.Config is nil")
+	}
+	if cfg.CAFile == "" {
+		return nil, nil, fmt.Errorf("server TLS config: CAFile is required for mTLS verification")
+	}
+
+	watcher, err := NewCertWatcher(cfg.CertFile, cfg.KeyFile, cfg.CAFile)
+	if err != nil {
+		return nil, nil, fmt.Errorf("server TLS cert watcher: %w", err)
+	}
+
+	tlsCfg := &tls.Config{
+		GetCertificate: watcher.GetCertificate,
+		// Use RequireAnyClientCert instead of RequireAndVerifyClientCert so we can
+		// verify the chain against the dynamically-reloaded CA pool in VerifyPeerCertificate.
+		ClientAuth: tls.RequireAnyClientCert,
+		MinVersion: tls.VersionTLS13,
+		// Manually verify the client certificate chain against the
+		// watcher's dynamic CA pool, enabling trust bundle rotation without restart.
+		VerifyPeerCertificate: verifyClientCert(watcher),
+	}
+
+	return tlsCfg, watcher, nil
+}
+
+// LoadClientConfig returns a tls.Config for a client that presents its cert and verifies the server.
+// Certs and the CA bundle are loaded dynamically via CertWatcher to support
+// zero-downtime rotation of both identity certificates and trust bundles.
+//
+// The expectedServerID is required; the server must present a matching SPIFFE ID in its URI SAN.
+// This sets InsecureSkipVerify=true to bypass standard DNS hostname checking
+// and manually verifies the cryptographic chain and the SPIFFE ID instead.
+func LoadClientConfig(cfg *Config, expectedServerID string) (*tls.Config, *CertWatcher, error) {
+	if expectedServerID == "" {
+		return nil, nil, fmt.Errorf("client TLS config: expectedServerID is required for SPIFFE verification")
+	}
+
+	if cfg == nil {
+		return nil, nil, fmt.Errorf("client TLS config: mtls.Config is nil")
+	}
+	if cfg.CAFile == "" {
+		return nil, nil, fmt.Errorf("client TLS config: CAFile is required for mTLS verification")
+	}
+
+	watcher, err := NewCertWatcher(cfg.CertFile, cfg.KeyFile, cfg.CAFile)
+	if err != nil {
+		return nil, nil, fmt.Errorf("client TLS cert watcher: %w", err)
+	}
+
+	tlsCfg := &tls.Config{
+		GetClientCertificate: watcher.GetClientCertificate,
+		MinVersion:           tls.VersionTLS13,
+		//nolint:gosec // G402: we use VerifyPeerCertificate for custom SPIFFE ID verification instead
+		InsecureSkipVerify: true,
+		// Manually verify the server certificate chain and SPIFFE ID against the
+		// watcher's dynamic CA pool, enabling trust bundle rotation without restart.
+		VerifyPeerCertificate: verifyServerCert(watcher, expectedServerID),
+	}
+
+	return tlsCfg, watcher, nil
+}
+
+// verifyClientCert verifies the client's certificate chain against the dynamic CA pool.
+// Used server-side where ClientAuth is RequireAnyClientCert (chain verification is done
+// manually, not by the TLS stack). Any client with a valid certificate signed by the
+// trusted CA is accepted; authorization is handled at the application layer.
+func verifyClientCert(watcher *CertWatcher) func([][]byte, [][]*x509.Certificate) error {
+	return func(rawCerts [][]byte, _ [][]*x509.Certificate) error {
+		if len(rawCerts) == 0 {
+			return fmt.Errorf("no client certificate presented")
+		}
+
+		peerCert, err := x509.ParseCertificate(rawCerts[0])
+		if err != nil {
+			return fmt.Errorf("parse client certificate: %w", err)
+		}
+
+		intermediates := x509.NewCertPool()
+		for i, rawCert := range rawCerts[1:] {
+			ic, err := x509.ParseCertificate(rawCert)
+			if err != nil {
+				return fmt.Errorf("parse client intermediate certificate %d: %w", i+1, err)
+			}
+			intermediates.AddCert(ic)
+		}
+
+		caPool, err := requireCAPool(watcher)
+		if err != nil {
+			return fmt.Errorf("verify client certificate chain: %w", err)
+		}
+		opts := x509.VerifyOptions{
+			Roots:         caPool,
+			Intermediates: intermediates,
+			KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
+		}
+		if _, err := peerCert.Verify(opts); err != nil {
+			return fmt.Errorf("verify client certificate chain: %w", err)
+		}
+
+		return nil
+	}
+}
+
+// verifyServerCert verifies the server's certificate chain against the dynamic CA pool
+// and checks the server's SPIFFE ID. Used client-side where InsecureSkipVerify is true
+// (chain verification and SPIFFE ID check are done manually).
+func verifyServerCert(watcher *CertWatcher, expectedID string) func([][]byte, [][]*x509.Certificate) error {
+	return func(rawCerts [][]byte, _ [][]*x509.Certificate) error {
+		if len(rawCerts) == 0 {
+			return fmt.Errorf("no server certificate presented")
+		}
+
+		peerCert, err := x509.ParseCertificate(rawCerts[0])
+		if err != nil {
+			return fmt.Errorf("parse server certificate: %w", err)
+		}
+
+		intermediates := x509.NewCertPool()
+		for i, rawCert := range rawCerts[1:] {
+			intermediateCert, err := x509.ParseCertificate(rawCert)
+			if err != nil {
+				return fmt.Errorf("parse server intermediate certificate %d: %w", i+1, err)
+			}
+			intermediates.AddCert(intermediateCert)
+		}
+
+		caPool, err := requireCAPool(watcher)
+		if err != nil {
+			return fmt.Errorf("verify server certificate chain: %w", err)
+		}
+		opts := x509.VerifyOptions{
+			Roots:         caPool,
+			Intermediates: intermediates,
+			KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		}
+		if _, err := peerCert.Verify(opts); err != nil {
+			return fmt.Errorf("verify server certificate chain: %w", err)
+		}
+
+		for _, uri := range peerCert.URIs {
+			if uri.String() == expectedID {
+				return nil
+			}
+		}
+		return fmt.Errorf("server certificate SPIFFE ID does not match expected %q", expectedID)
+	}
+}
+
+func requireCAPool(watcher *CertWatcher) (*x509.CertPool, error) {
+	caPool := watcher.GetCAPool()
+	if caPool == nil {
+		return nil, fmt.Errorf("CA pool is required for mTLS verification")
+	}
+	return caPool, nil
+}
--- a/pkg/mtls/loader_test.go
+++ b/pkg/mtls/loader_test.go
@@ -1,0 +1,335 @@
+/*
+Copyright The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mtls
+
+import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"math/big"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+)
+
+// generateTestCerts creates a self-signed CA and a leaf certificate signed by that CA.
+func generateTestCerts(t *testing.T) (certFile, keyFile, caFile string) {
+	t.Helper()
+	return generateTestCertsWithSPIFFEID(t, "")
+}
+
+// generateTestCertsWithSPIFFEID creates a CA + leaf cert with an optional SPIFFE ID URI SAN.
+func generateTestCertsWithSPIFFEID(t *testing.T, spiffeID string) (certFile, keyFile, caFile string) {
+	t.Helper()
+	dir := t.TempDir()
+
+	// Generate CA
+	caKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	if err != nil {
+		t.Fatalf("generate CA key: %v", err)
+	}
+	caTemplate := &x509.Certificate{
+		SerialNumber:          big.NewInt(1),
+		Subject:               pkix.Name{Organization: []string{"Test CA"}},
+		NotBefore:             time.Now().Add(-1 * time.Hour),
+		NotAfter:              time.Now().Add(24 * time.Hour),
+		KeyUsage:              x509.KeyUsageCertSign | x509.KeyUsageCRLSign,
+		IsCA:                  true,
+		BasicConstraintsValid: true,
+	}
+	caCertDER, err := x509.CreateCertificate(rand.Reader, caTemplate, caTemplate, &caKey.PublicKey, caKey)
+	if err != nil {
+		t.Fatalf("create CA cert: %v", err)
+	}
+	caFile = filepath.Join(dir, "ca.pem")
+	if err := writePEM(caFile, "CERTIFICATE", caCertDER); err != nil {
+		t.Fatalf("write CA file: %v", err)
+	}
+	caCert, err := x509.ParseCertificate(caCertDER)
+	if err != nil {
+		t.Fatalf("parse CA cert: %v", err)
+	}
+
+	// Generate leaf cert
+	leafKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	if err != nil {
+		t.Fatalf("generate leaf key: %v", err)
+	}
+	leafTemplate := &x509.Certificate{
+		SerialNumber: big.NewInt(2),
+		Subject:      pkix.Name{Organization: []string{"Test Leaf"}},
+		NotBefore:    time.Now().Add(-1 * time.Hour),
+		NotAfter:     time.Now().Add(24 * time.Hour),
+		KeyUsage:     x509.KeyUsageDigitalSignature,
+		ExtKeyUsage:  []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth, x509.ExtKeyUsageClientAuth},
+	}
+	if spiffeID != "" {
+		u, err := url.Parse(spiffeID)
+		if err != nil {
+			t.Fatalf("parse SPIFFE ID URL: %v", err)
+		}
+		leafTemplate.URIs = []*url.URL{u}
+	}
+	leafCertDER, err := x509.CreateCertificate(rand.Reader, leafTemplate, caCert, &leafKey.PublicKey, caKey)
+	if err != nil {
+		t.Fatalf("create leaf cert: %v", err)
+	}
+	certFile = filepath.Join(dir, "cert.pem")
+	if err := writePEM(certFile, "CERTIFICATE", leafCertDER); err != nil {
+		t.Fatalf("write cert file: %v", err)
+	}
+	keyFile = filepath.Join(dir, "key.pem")
+	keyDER, err := x509.MarshalECPrivateKey(leafKey)
+	if err != nil {
+		t.Fatalf("marshal leaf key: %v", err)
+	}
+	if err := writePEM(keyFile, "EC PRIVATE KEY", keyDER); err != nil {
+		t.Fatalf("write key file: %v", err)
+	}
+
+	return certFile, keyFile, caFile
+}
+
+func writePEM(path, blockType string, data []byte) error {
+	f, err := os.Create(path)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	return pem.Encode(f, &pem.Block{Type: blockType, Bytes: data})
+}
+
+// --- LoadServerConfig ---
+
+func TestLoadServerConfig(t *testing.T) {
+	certFile, keyFile, caFile := generateTestCerts(t)
+	cfg := &Config{CertFile: certFile, KeyFile: keyFile, CAFile: caFile}
+
+	tlsCfg, watcher, err := LoadServerConfig(cfg)
+	if err != nil {
+		t.Fatalf("LoadServerConfig() error: %v", err)
+	}
+	defer watcher.Stop()
+	// RequireAnyClientCert: require a cert, but verify it manually via VerifyPeerCertificate
+	if tlsCfg.ClientAuth != tls.RequireAnyClientCert {
+		t.Errorf("ClientAuth = %v, want RequireAnyClientCert", tlsCfg.ClientAuth)
+	}
+	// ClientCAs is nil — chain verification happens dynamically via watcher.GetCAPool()
+	if tlsCfg.ClientCAs != nil {
+		t.Error("ClientCAs should be nil (CA verification is done dynamically)")
+	}
+	if tlsCfg.GetCertificate == nil {
+		t.Error("GetCertificate callback is nil")
+	}
+	if tlsCfg.MinVersion != tls.VersionTLS13 {
+		t.Errorf("MinVersion = %d, want TLS 1.3", tlsCfg.MinVersion)
+	}
+	// VerifyPeerCertificate is always set (chain verification)
+	if tlsCfg.VerifyPeerCertificate == nil {
+		t.Error("VerifyPeerCertificate should always be set for dynamic CA verification")
+	}
+}
+
+func TestLoadClientConfig_EmptyServerID(t *testing.T) {
+	certFile, keyFile, caFile := generateTestCerts(t)
+	cfg := &Config{CertFile: certFile, KeyFile: keyFile, CAFile: caFile}
+
+	_, _, err := LoadClientConfig(cfg, "")
+	if err == nil {
+		t.Fatal("LoadClientConfig() expected error for empty ServerID, got nil")
+	}
+	if !strings.Contains(err.Error(), "expectedServerID is required") {
+		t.Errorf("error = %q, want mention of expectedServerID", err.Error())
+	}
+}
+
+func TestLoadClientConfig_WithSPIFFEID(t *testing.T) {
+	certFile, keyFile, caFile := generateTestCerts(t)
+	cfg := &Config{CertFile: certFile, KeyFile: keyFile, CAFile: caFile}
+
+	tlsCfg, watcher, err := LoadClientConfig(cfg, "spiffe://cluster.local/ns/default/sa/wm")
+	if err != nil {
+		t.Fatalf("LoadClientConfig() error: %v", err)
+	}
+	defer watcher.Stop()
+	if !tlsCfg.InsecureSkipVerify {
+		t.Error("InsecureSkipVerify should be true when expectedServerID is set")
+	}
+	if tlsCfg.VerifyPeerCertificate == nil {
+		t.Error("VerifyPeerCertificate should be set when expectedServerID provided")
+	}
+}
+
+// --- Error cases ---
+
+func TestLoadServerConfig_NilConfig(t *testing.T) {
+	_, _, err := LoadServerConfig(nil)
+	if err == nil {
+		t.Fatal("expected error for nil config")
+	}
+	if !strings.Contains(err.Error(), "mtls.Config is nil") {
+		t.Errorf("error = %q, want mention of nil config", err.Error())
+	}
+}
+
+func TestLoadClientConfig_NilConfig(t *testing.T) {
+	_, _, err := LoadClientConfig(nil, "spiffe://cluster.local/ns/default/sa/test")
+	if err == nil {
+		t.Fatal("expected error for nil config")
+	}
+	if !strings.Contains(err.Error(), "mtls.Config is nil") {
+		t.Errorf("error = %q, want mention of nil config", err.Error())
+	}
+}
+
+func TestLoadServerConfig_EmptyCAFile(t *testing.T) {
+	certFile, keyFile, _ := generateTestCerts(t)
+	cfg := &Config{CertFile: certFile, KeyFile: keyFile}
+
+	_, _, err := LoadServerConfig(cfg)
+	if err == nil {
+		t.Fatal("expected error for empty CA file")
+	}
+	if !strings.Contains(err.Error(), "CAFile is required") {
+		t.Errorf("error = %q, want mention of required CAFile", err.Error())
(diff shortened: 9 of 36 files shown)


pkg/mtls/watcher.go
```diff
@@ -0,0 +1,149 @@
...
+			if event.Has(fsnotify.Write) || event.Has(fsnotify.Create) {
+				if err := cw.reload(); err != nil {
+					klog.Errorf("Failed to reload certificate: %v", err)
+				}
+			}
+			// Re-watch after atomic rename (spiffe-helper does atomic renames for both cert and key)
+			if event.Has(fsnotify.Rename) || event.Has(fsnotify.Remove) {
+				targetFile := event.Name
+				_ = cw.watcher.Remove(targetFile)
+				if err := cw.watcher.Add(targetFile); err != nil {
```
hzxuzhonghu: If `watcher.Add(targetFile)` fails here (for example, spiffe-helper renames cert.pem away *before* atomically placing the new one), the file is permanently dropped from the watch set and no further events will be received for it — cert hot-reload silently stops working.

The current code only logs the error, so a rotation failure could go unnoticed until the cert expires. Two safer options:
1. **Watch the parent directory** rather than individual files (standard `cert-manager`/`spiffe-helper` approach).
2. **Implement a short retry loop** (e.g. try re-Add up to 5× with 50 ms sleep) so a transient race during atomic rename is recovered.

At minimum, downgrade the log to `klog.V(4)` for the expected case and add a clear `klog.Errorf` when the retry budget is exhausted so operators know cert reload is degraded.

pkg/mtls/loader.go
```diff
@@ -0,0 +1,163 @@
...
+		return nil, nil, fmt.Errorf("client TLS cert watcher: %w", err)
+	}
+
+	tlsCfg := &tls.Config{
+		GetClientCertificate: watcher.GetClientCertificate,
+		RootCAs:              caPool,
+		MinVersion:           tls.VersionTLS12,
+	}
+
+	if expectedServerID != "" {
```
hzxuzhonghu: When `expectedServerID == ""` the `else` branch (lines 78-82) returns a `tls.Config` with standard hostname verification enabled (`InsecureSkipVerify=false`, `ServerName=""`). SPIFFE-issued certificates carry only URI SANs — no DNS SANs — so Go's hostname check will fail with `x509: certificate is valid for ...` against any SPIFFE endpoint. The config returned by this branch is essentially unusable with the mTLS setup this package targets.

Options:
1. Require `expectedServerID` to be non-empty when mTLS is configured, and return an `error` if it is empty.
2. Document clearly in the godoc that `expectedServerID=""` is only valid when connecting to non-SPIFFE endpoints that have a matching DNS SAN, and consider setting `tlsCfg.ServerName` to prevent accidental misuse.

pkg/picod/server.go
```diff
@@ -30,6 +30,15 @@ import (
 type Config struct {
 	Port      int    `json:"port"`
 	Workspace string `json:"workspace"`
+
+	// mTLS configuration (certificate source abstraction)
+
+	// MTLSCertFile is the path to the mTLS certificate (--mtls-cert-file)
+	MTLSCertFile string `json:"mtlsCertFile"`
```
hzxuzhonghu: These three fields (`MTLSCertFile`, `MTLSKeyFile`, `MTLSCAFile`) are plumbed through and validated in `cmd/picod/main.go`, but `Run()` always calls `server.ListenAndServe()` (plain HTTP) regardless of whether they are set. The same pattern applies in `pkg/workloadmanager/server.go` (`Start()`) and `pkg/router` (`server.go`/`Start()`).

Please add either:
- A `klog.Warningf("mTLS cert paths configured but listener not yet wired — ignored until follow-up PR")` guard in `Run()`/`Start()` when `MTLSCertFile != ""`, or
- A `// TODO(<author>): wire mtls.LoadServerConfig here — follow-up PR` comment at the `ListenAndServe` call site

Otherwise operators silently pass flags that have zero effect.

pkg/mtls/loader_test.go
```diff
@@ -0,0 +1,347 @@
...
+	return pem.Encode(f, &pem.Block{Type: blockType, Bytes: data})
+}
+
+// --- LoadServerConfig ---
+
+func TestLoadServerConfig(t *testing.T) {
+	certFile, keyFile, caFile := generateTestCerts(t)
+	cfg := &Config{CertFile: certFile, KeyFile: keyFile, CAFile: caFile}
+
+	tlsCfg, watcher, err := LoadServerConfig(cfg, nil)
```
hzxuzhonghu: This test only exercises `LoadServerConfig(cfg, nil)` (no `expectedClientIDs`). The code path where `VerifyPeerCertificate` is injected for SPIFFE enforcement (`LoadServerConfig(cfg, []string{spiffeID})`) has no test. Since this callback is the central security enforcement for server-side identity, please add a test analogous to `TestVerifyPeerSPIFFEID_MatchingID`/`NonMatchingID` but going through `LoadServerConfig`:
```go
func TestLoadServerConfig_WithExpectedClientIDs(t *testing.T) {
    certFile, keyFile, caFile := generateTestCerts(t)
    cfg := &Config{CertFile: certFile, KeyFile: keyFile, CAFile: caFile}
    tlsCfg, watcher, err := LoadServerConfig(cfg, []string{"spiffe://cluster.local/test"})
    ...
    if tlsCfg.VerifyPeerCertificate == nil {
        t.Error("VerifyPeerCertificate should be set when expectedClientIDs provided")
    }
    // Also exercise the callback directly to confirm reject/accept behaviour.
}
```

hzxuzhonghu (changes requested):
The `pkg/mtls` package is well-structured — clean separation of `Config`, `CertWatcher`, and loader helpers, good test coverage, and the `sync.Once`-guarded `Stop()` is correct. A few issues to address before merging.

**Across all three binaries** (`pkg/picod/server.go`, `pkg/workloadmanager/server.go`, `pkg/router/config.go`): the mTLS fields are validated at startup but never consumed by the HTTP listeners. An operator who deploys with `--mtls-cert-file` etc. will see the flags silently have no effect — no error, no warning. This is the stated intent (follow-up PR will wire it), but please at minimum add a startup `klog.Warningf` or a `// TODO(author): wire mTLS in follow-up PR` comment at each `ListenAndServe` / `Start` call site so the intent is clear in the code.

pkg/picod/server.go
mahil-2040: I've added a startup klog.Warningf and a TODO guard at all three ListenAndServe call sites to explicitly warn operators until the follow-up wiring PR.

pkg/mtls/watcher.go
mahil-2040: I went with your 2nd suggested approach and implemented a 5-iteration retry loop with a 50ms sleep, plus a CRITICAL log if the budget is exhausted.

pkg/mtls/loader.go
mahil-2040: I've strictly enforced Option 1: LoadClientConfig now returns an explicit error immediately if expectedServerID is empty, completely preventing accidental fallback to standard DNS SAN verification.

pkg/mtls/loader_test.go
mahil-2040: I've added TestLoadServerConfig_WithExpectedClientIDs which directly exercises the injected VerifyPeerCertificate callback to confirm it correctly accepts and rejects matching/non-matching identities.

pkg/mtls/watcher.go
```diff
@@ -0,0 +1,165 @@
...
+
+func (cw *CertWatcher) watchLoop() {
+	for {
+		select {
+		case event, ok := <-cw.watcher.Events:
+			if !ok {
+				return
+			}
+			// Reload on write or create
+			if event.Has(fsnotify.Write) || event.Has(fsnotify.Create) {
```
hzxuzhonghu: **Bug – cert/key rotation race (no debounce).**

Both cert and key files are watched independently, and every Write/Create event immediately calls `reload()` → `tls.LoadX509KeyPair(certFile, keyFile)`. During an atomic rotation (e.g. spiffe-helper renames cert, *then* renames key), the first event loads the **new** cert with the **old** key → `"private key does not match public key"` error is logged, and the watcher falls back to the stale certificate.

The second file event self-heals, so it's not an outage, but with SPIRE SVIDs (1-hour TTL) you'd get ~24 spurious `klog.Errorf` lines/day/component — noisy for alerting.

**Suggested fix:** add a short debounce (e.g. a `time.AfterFunc` of 200–300 ms that resets on each event) so both files settle before calling `reload()`. controller-runtime's [certwatcher](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/certwatcher/certwatcher.go) uses exactly this pattern.

pkg/mtls/watcher.go
```diff
@@ -0,0 +1,165 @@
...
+	for i := 0; i < 5; i++ {
+		if err := cw.watcher.Add(targetFile); err != nil {
+			klog.V(4).Infof("Failed to re-watch %s (attempt %d/5): %v", targetFile, i+1, err)
+			time.Sleep(50 * time.Millisecond)
+			continue
+		}
+		added = true
+		break
+	}
+	if !added {
```
hzxuzhonghu: **Bug – permanent watch loss after retry exhaustion; "degraded" is misleading.**

If all 5 retries (total budget 250 ms) fail, the log says *"Certificate hot-reload is degraded!"* — but in reality the file watch for `targetFile` is **permanently lost**. No future changes to that file will ever be detected; the component silently serves a stale certificate until it expires, then connections fail with no self-healing path (restart required).

Two issues:
1. **Retry budget is very tight.** 5 × 50 ms = 250 ms. On network-mounted volumes or under I/O pressure, file creation after an atomic swap can take longer. Consider backing off (e.g. 100, 200, 400, 800, 1600 ms) and/or bumping retries.
2. **No recovery path.** Once the watch is lost the only fix is a process restart. Consider a background re-try goroutine, or at minimum change the error message from "degraded" to something like "permanently broken — process restart required to restore certificate rotation."

pkg/mtls/loader.go
```diff
@@ -0,0 +1,165 @@
...
+
+// LoadServerConfig returns a tls.Config for a server that requires mTLS.
+// Certs are loaded dynamically via GetCertificate to support rotation.
+// If expectedClientIDs is non-empty, clients must present a matching SPIFFE ID.
+func LoadServerConfig(cfg *Config, expectedClientIDs []string) (*tls.Config, *CertWatcher, error) {
+	if cfg == nil {
+		return nil, nil, fmt.Errorf("server TLS config: mtls.Config is nil")
+	}
+
+	caPool, err := loadCAPool(cfg.CAFile)
```
hzxuzhonghu: **Design gap – CA bundle is loaded once and never refreshed.**

`loadCAPool` reads the CA file at startup and bakes the resulting `*x509.CertPool` into `tls.Config.ClientCAs` / `RootCAs`. If the CA rotates (standard in SPIFFE trust domains — trust bundle can change when a root or intermediate CA is re-keyed), the stale pool will reject certs signed by the new CA, causing a **full connection outage** once the old CA's certs expire.

Since this is a foundational PR, this doesn't need to be fixed here, but please document the limitation prominently (e.g. in the `LoadServerConfig` / `LoadClientConfig` godocs) so the follow-up PR that wires listeners knows to address it. A `CAWatcher` analogous to `CertWatcher` would close the gap.

pkg/mtls/loader.go
```diff
@@ -0,0 +1,165 @@
...
+			if err != nil {
+				return fmt.Errorf("parse peer intermediate certificate %d: %w", i+1, err)
+			}
+			intermediates.AddCert(intermediateCert)
+		}
+
+		opts := x509.VerifyOptions{
+			Roots:         caPool,
+			Intermediates: intermediates,
+		}
```
hzxuzhonghu, line 166: **Compatibility – `VerifyOptions.KeyUsages` defaults to `ExtKeyUsageServerAuth` only.**

When `KeyUsages` is nil/empty, Go's `x509.Certificate.Verify` [defaults to `[]ExtKeyUsage{ExtKeyUsageServerAuth}`](https://pkg.go.dev/crypto/x509#VerifyOptions). The SPIFFE X.509-SVID spec says leaf SVIDs **SHOULD** (not MUST) include `id-kp-serverAuth`. Certs from non-SPIRE CAs or custom PKIs that omit this EKU will be silently rejected here.

Since the library aims to be "certificate-source agnostic", consider explicitly setting:
```go
opts := x509.VerifyOptions{
    Roots:         caPool,
    Intermediates: intermediates,
    KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageAny},
}
```
This makes chain verification purely about trust, leaving EKU policy to the caller/SPIFFE layer.

pkg/router/server.go
```diff
@@ -184,6 +184,10 @@ func (s *Server) Start(ctx context.Context) error {
 	klog.Infof("Router server listening on %s", addr)
 
 	// Start HTTP or HTTPS server
+	if s.config.MTLSCertFile != "" || s.config.MTLSKeyFile != "" || s.config.MTLSCAFile != "" {
+		klog.Warningf("TODO(author): wire mTLS in follow-up PR. mTLS flags are currently parsed but not consumed by the Router listener.")
```
hzxuzhonghu: **Nit – `TODO(author)` violates K8s convention.**

Per [K8s coding conventions](https://www.kubernetes.dev/docs/guide/coding-convention/), TODOs should use a real username: `TODO(mahil-2040):`. Same applies to the identical TODOs in `pkg/picod/server.go:119` and `pkg/workloadmanager/server.go:177`.

hzxuzhonghu (commented):
Foundational mTLS library looks solid overall — good test coverage, correct use of `InsecureSkipVerify` with manual chain verification, and clean SPIFFE ID checking. Flagging a few issues that could bite in production rotation scenarios.

cmd/picod/main.go
```diff
@@ -21,20 +21,39 @@ import (
...
 )
 
 func main() {
 	port := flag.Int("port", 8080, "Port for the PicoD server to listen on")
 	workspace := flag.String("workspace", "", "Root directory for file operations (default: current working directory)")
 
+	// mTLS flags (certificate source abstraction)
+	mtlsCertFile := flag.String("mtls-cert-file", "", "Path to mTLS certificate file")
+	mtlsKeyFile := flag.String("mtls-key-file", "", "Path to mTLS private key file")
+	mtlsCAFile := flag.String("mtls-ca-file", "", "Path to mTLS CA bundle file")
```
hzxuzhonghu: can change to not use pointer type

cmd/picod/main.go
```diff
@@ -21,20 +21,39 @@ import (
...
 	"github.com/volcano-sh/agentcube/pkg/picod"
 )
 
 func main() {
 	port := flag.Int("port", 8080, "Port for the PicoD server to listen on")
 	workspace := flag.String("workspace", "", "Root directory for file operations (default: current working directory)")
 
+	// mTLS flags (certificate source abstraction)
+	mtlsCertFile := flag.String("mtls-cert-file", "", "Path to mTLS certificate file")
+	mtlsKeyFile := flag.String("mtls-key-file", "", "Path to mTLS private key file")
```
hzxuzhonghu: BTW. I forget have you mounted this dir anywhere?

pkg/mtls/loader.go
```diff
@@ -0,0 +1,165 @@
...
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"os"
+)
+
+// LoadServerConfig returns a tls.Config for a server that requires mTLS.
+// Certs are loaded dynamically via GetCertificate to support rotation.
+// If expectedClientIDs is non-empty, clients must present a matching SPIFFE ID.
+func LoadServerConfig(cfg *Config, expectedClientIDs []string) (*tls.Config, *CertWatcher, error) {
```
hzxuzhonghu: Not sure i missed, where do you call this function?

hzxuzhonghu (commented):
Please also make the e2e tests with mtls

cmd/picod/main.go
mahil-2040: For the control plane (Router/WorkloadManager), we already mounted the spiffe-helper volumes in the previous SPIRE PR. For PicoD, the sandbox pod is built dynamically in Go code. I haven't wired the volume mounts yet, but I added a specific TODO in workload_builder.go so we don't forget to mount the SPIRE CSI driver (or sidecar) in the follow-up PR where the listeners are wired!

pkg/mtls/loader.go
mahil-2040: It's not called anywhere in this PR. As mentioned in the PR description, this PR strictly introduces the foundational flags, loader abstractions, and CertWatcher logic to keep the review size small. The actual 'wiring' (calling LoadServerConfig and injecting the tls.Config into the HTTP/gRPC listeners) will be done in the immediate follow-up PR. I have added the TODO(mahil-2040) and klog.Warningf guards at the ListenAndServe call sites in the router/workloadmanager/picod specifically to mark where these calls will eventually go!

mahil-2040:
> Please also make the e2e tests with mtls

The main cluster e2e tests are in test/e2e/, but since we haven't wired mTLS into the Router/PicoD listeners yet (that's the follow-up PR!), I wrote a comprehensive module-level e2e test suite in pkg/mtls/mtls_e2e_test.go. It spins up real TLS servers and clients to exercise the full handshake pipeline and SPIFFE ID validation across 5 different security scenarios (success, wrong client ID, wrong server ID, untrusted CA, and missing client cert). Once we wire up the listeners in the follow-up PR, i will add mTLS coverage to the main test/e2e/ suite.

pkg/mtls/loader.go
hzxuzhonghu: How would you do that, how about leaving this implement there.

hzxuzhonghu:
Imean at least in this pr. You can enable spre cert provision and mtls within e2e setup, you donot actually introduce any new test cases. 

One concern here is: for picod,  i am not sure how you provide the certs to codeinterpreter{picod}, it is declared by enduser.

mahil-2040:
> Imean at least in this pr. You can enable spre cert provision and mtls within e2e setup, you donot actually introduce any new test cases.
> 
> One concern here is: for picod, i am not sure how you provide the certs to codeinterpreter{picod}, it is declared by enduser.

That makes total sense regarding the e2e tests. But i think doing both the listener wiring and upgrading the entire e2e test harness to deploy SPIRE infrastructure will make the scope of this PR quite large. Since this PR currently focuses heavily on the Go crypto/tls and CertWatcher abstractions, would you be okay if we keep this PR focused purely on establishing this foundation? for now i will remove the e2e test file that i have added.

Then I can open the follow-up PR which will be dedicated entirely to: 1. . Wiring LoadServerConfig into the HTTP listeners (attaching the resulting tls.Config to the http.Server and switching to ListenAndServeTLS. 2. Injecting the SPIRE CSI driver (or spiffe-helper) volume mounts into the dynamic CodeInterpreter pod spec in workload_builder.go. 3. Modifying the e2e test harness to deploy SPIRE and run the existing tests natively over mTLS.

pkg/mtls/watcher.go
```diff
@@ -0,0 +1,194 @@
...
+	return nil
+}
+
+func (cw *CertWatcher) watchLoop() {
+	var debounceTimer *time.Timer
+	scheduleReload := func() {
+		if debounceTimer != nil {
+			debounceTimer.Stop()
+		}
+		debounceTimer = time.AfterFunc(200*time.Millisecond, func() {
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

pkg/mtls/loader.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("server TLS cert watcher: %w", err)
+	}
+
+	tlsCfg := &tls.Config{
+		GetCertificate: watcher.GetCertificate,
+		ClientCAs:      caPool,
+		ClientAuth:     tls.RequireAndVerifyClientCert,
+		MinVersion:     tls.VersionTLS12,
+	}
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

pkg/mtls/spiffeid.go
```diff
@@ -0,0 +1,35 @@
...
+const (
+	// RouterSPIFFEID is the SPIFFE identity for the Router component.
+	RouterSPIFFEID = "spiffe://cluster.local/ns/agentcube-system/sa/agentcube-router"
+
+	// WorkloadManagerSPIFFEID is the SPIFFE identity for the WorkloadManager component.
+	WorkloadManagerSPIFFEID = "spiffe://cluster.local/ns/agentcube-system/sa/workloadmanager"
+
+	// SandboxSPIFFEID is the SPIFFE identity for PicoD sandbox pods.
+	// Note: PicoD omits the namespace segment because sandboxes can run in any namespace.
+	SandboxSPIFFEID = "spiffe://cluster.local/sa/agentcube-sandbox"
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

pkg/workloadmanager/workload_builder.go
```diff
@@ -437,3 +456,114 @@ func buildSandboxByCodeInterpreter(namespace string, codeInterpreterName string,
...
+	// spireAgentSocketVolumeName is the volume exposing the SPIRE Agent socket.
+	spireAgentSocketVolumeName = "spire-agent-socket"
+	// spireAgentSocketPath is the host path where the SPIRE Agent socket lives.
+	spireAgentSocketPath = "/run/spire/sockets"
+	// spiffeHelperConfigVolumeName is the volume holding the spiffe-helper configuration.
+	spiffeHelperConfigVolumeName = "spiffe-helper-config"
+	// spiffeHelperConfigMapName is the ConfigMap name created by the Helm chart.
+	spiffeHelperConfigMapName = "spiffe-helper-config"
+	// spiffeHelperImage is the container image for the spiffe-helper sidecar.
+	spiffeHelperImage = "ghcr.io/spiffe/spiffe-helper:0.8.0"
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

pkg/workloadmanager/workload_builder.go
```diff
@@ -437,3 +456,114 @@ func buildSandboxByCodeInterpreter(namespace string, codeInterpreterName string,
...
+			continue
+		}
+
+		podSpec.Containers[i].VolumeMounts = append(podSpec.Containers[i].VolumeMounts, corev1.VolumeMount{
+			Name:      spireCertVolumeName,
+			MountPath: spireCertMountPath,
+			ReadOnly:  true,
+		})
+
+		// Append mTLS flags ONLY if the container is a known PicoD entrypoint
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

hzxuzhonghu (commented):
test

pkg/picod/server.go
```diff
@@ -106,9 +133,57 @@ func (s *Server) Run() error {
...
+		return fmt.Errorf("--enable-mtls requires --mtls-cert-file, --mtls-key-file, and --mtls-ca-file to be set")
+	}
+
+	// The spiffe-helper sidecar writes the certificates asynchronously on startup.
+	// We must retry loading them here to avoid crashing before the sidecar has provisioned them.
+	var serverTLS *tls.Config
+	var watcher *mtls.CertWatcher
+	var err error
+	backoff := 10 * time.Millisecond
+	for i := 0; i < 50; i++ {
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

pkg/workloadmanager/workload_builder.go
```diff
@@ -422,6 +437,10 @@ func buildSandboxByCodeInterpreter(namespace string, codeInterpreterName string,
 		},
 	}
 
+	if enableMTLS {
```
hzxuzhonghu: test

pkg/workloadmanager/workload_builder.go
```diff
@@ -422,6 +437,10 @@ func buildSandboxByCodeInterpreter(namespace string, codeInterpreterName string,
...
+		injectMTLSVolumes(&podSpec)
+	}
+
 	buildParams := &buildSandboxParams{
 		sandboxName:    sandboxName,
```
hzxuzhonghu, line 436: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
 
 	"k8s.io/klog/v2"
 
+	"github.com/volcano-sh/agentcube/pkg/mtls"
 	"github.com/volcano-sh/agentcube/pkg/picod"
 )
 
 func main() {
 	port := flag.Int("port", 8080, "Port for the PicoD server to listen on")
 	workspace := flag.String("workspace", "", "Root directory for file operations (default: current working directory)")
```
hzxuzhonghu, line 33: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+	enableMTLS := flag.Bool("enable-mtls", false, "Enable mutual TLS on the PicoD listener")
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+	// mTLS flags (certificate source abstraction)
```
hzxuzhonghu: test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+	var mtlsCertFile, mtlsKeyFile, mtlsCAFile string
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+	flag.StringVar(&mtlsCertFile, "mtls-cert-file", "", "Path to mTLS certificate file")
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+	flag.StringVar(&mtlsKeyFile, "mtls-key-file", "", "Path to mTLS private key file")
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+	flag.StringVar(&mtlsCAFile, "mtls-ca-file", "", "Path to mTLS CA bundle file")
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
```
hzxuzhonghu, line 34: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
 
 	// Initialize klog flags
```
hzxuzhonghu, line 35: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
 	klog.InitFlags(nil)
```
hzxuzhonghu, line 36: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
 	flag.Parse()
```
hzxuzhonghu, line 37: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
```
hzxuzhonghu, line 38: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
 
+	// Validate mTLS configuration early (fail fast on bad flags)
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+	mTLSCfg := mtls.Config{
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+		CertFile: mtlsCertFile,
```
hzxuzhonghu: test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+		KeyFile:  mtlsKeyFile,
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+		CAFile:   mtlsCAFile,
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+	}
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+	if err := mTLSCfg.Validate(); err != nil {
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+		klog.Fatalf("Invalid mTLS configuration: %v", err)
```
hzxuzhonghu: test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+	}
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
 	config := picod.Config{
```
hzxuzhonghu, line 39: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
-		Port:      *port,
-		Workspace: *workspace,
+		Port:           *port,
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+		Workspace:      *workspace,
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+		EnableMTLS:     *enableMTLS,
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+		MTLSCertFile:   mtlsCertFile,
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+		MTLSKeyFile:    mtlsKeyFile,
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
+		MTLSCAFile:     mtlsCAFile,
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
 	}
```
hzxuzhonghu, line 42: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
```
hzxuzhonghu, line 43: test

hzxuzhonghu (commented):
test

cmd/picod/main.go
```diff
@@ -21,20 +21,42 @@ import (
...
 
 	// Create and start server
```
hzxuzhonghu: test

hzxuzhonghu (commented):
test

pkg/mtls/watcher.go
hzxuzhonghu: **Use-after-stop race.** `time.AfterFunc` fires in a new goroutine. If `Stop()` is called while the 200ms debounce is in flight, the closure will still invoke `cw.reload()` after the watcher is torn down. Add a done-channel check inside the closure:
```go
debounceTimer = time.AfterFunc(200*time.Millisecond, func() {
    select {
    case <-cw.done:
        return
    default:
    }
    if err := cw.reload(); err != nil { ... }
})
```

pkg/mtls/loader.go
hzxuzhonghu: Prefer `tls.VersionTLS13` for internal service-to-service mTLS. All three components are modern Go, and TLS 1.3 removes weak cipher suites still negotiable under TLS 1.2.

pkg/mtls/spiffeid.go
hzxuzhonghu: `SandboxSPIFFEID` omits the `ns/<namespace>` segment. The SPIRE Kubernetes PSAT node attestor typically issues SVIDs as `spiffe://<trust-domain>/ns/<namespace>/sa/<service-account>`. If the actual SVID differs from this constant, every PicoD mTLS handshake will fail at `VerifyPeerCertificate`. Please confirm this matches the SPIRE registration entry exactly.

pkg/workloadmanager/workload_builder.go
hzxuzhonghu: `spiffeHelperImage` is hardcoded to `ghcr.io/spiffe/spiffe-helper:0.8.0`. Operators in air-gapped environments or upgrading SPIFFE helper versions must recompile. Move this into `Config.SPIFFEHelperImage` with this string as the default.

pkg/workloadmanager/workload_builder.go
hzxuzhonghu: Silent failure: if no container is named `"picod"` or `"code-interpreter"`, the `--enable-mtls` args are never injected but there's no warning emitted. AgentRuntime container names are user-defined, so this will silently produce sandboxes with mTLS volumes but no mTLS flags. Add a warning log if the loop finds zero matching containers.

pkg/workloadmanager/workload_builder.go
hzxuzhonghu: **mTLS not injected on the warm-pool (SandboxClaim) path.** When `WarmPoolSize > 0`, `buildSandboxByCodeInterpreter` returns early at line ~413 before reaching this `injectMTLSVolumes` call. Claim-provisioned sandbox pods will not get the spiffe-helper sidecar or cert mounts. If the SandboxTemplate already embeds the sidecar, add a comment here. Otherwise `injectMTLSVolumes` needs to run on the claim-path pod spec too.

cmd/picod/main.go
hzxuzhonghu: **No graceful shutdown — certWatcher leaks on SIGTERM.** `server.Run()` blocks with no signal handler. When the pod is killed, the process exits immediately without calling `certWatcher.Stop()`, leaking the fsnotify goroutine and file descriptors. Add signal handling (see how `cmd/router/main.go` does it) and call a `Shutdown()`/`Stop()` on the server to clean up the watcher.

pkg/picod/server.go
hzxuzhonghu: The retry loop is silent — a 25s stall waiting for SPIRE certs is invisible to operators. Add `klog.V(2).Infof` on each failed attempt. Also consider starting `backoff` at 100ms instead of 10ms; the SPIRE helper sidecar is unlikely to provision certs in under 100ms.

hzxuzhonghu (commented):
Overall solid implementation — the abstraction layer is clean, the CertWatcher pattern is correct, and wiring mTLS across all three components consistently is the right approach. A few correctness issues below.

cmd/picod/main.go
hzxuzhonghu, line 1: can we revert picod  side, this belong to users' workload. This is hacky, correct me if i am wrong

cmd/router/main.go
```diff
@@ -36,22 +37,58 @@ func main() {
 		tlsKey                = flag.String("tls-key", "", "Path to TLS key file")
 		debug                 = flag.Bool("debug", false, "Enable debug mode")
 		maxConcurrentRequests = flag.Int("max-concurrent-requests", 1000, "Maximum number of concurrent requests that a router server can handle (0 = unlimited)")
+		enableMTLS            = flag.Bool("enable-mtls", false, "Enable mutual TLS for Router outbound connections")
```
hzxuzhonghu: The description is not correct, router can access workload manager, we can enforce mtls.

it can also access agentruntime, codeinterpreter, the image may belong to user, may not support mtls at all

pkg/mtls/spiffeid.go
mahil-2040: It omit `/ns/<namespace>` because sandbox pods can run in any namespace. The standard Istio-convention format (`/ns/<ns>/sa/<sa>`) is used for Router and WorkloadManager since they always run in `agentcube-system`. I expanded the comment to clearly explain

cmd/picod/main.go
mahil-2040: The shutdown sequence on `SIGTERM` is now:

1. Signal received → context cancelled
2. HTTP server drains in-flight requests (up to 10s)
3. `certWatcher.Stop()` closes the fsnotify watcher and its goroutine
4. Process exits cleanly

hzxuzhonghu (commented):
I will continue reading the rest parts this afternoon

cmd/router/main.go
```diff
@@ -36,22 +37,58 @@ func main() {
...
 		Debug:                 *debug,
 		EnableTLS:             *enableTLS,
 		TLSCert:               *tlsCert,
 		TLSKey:                *tlsKey,
 		MaxConcurrentRequests: *maxConcurrentRequests,
+		EnableMTLS:            *enableMTLS,
+		PicodAuthMode:         *picodAuthMode,
+		MTLSCertFile:          mtlsCertFile,
+		MTLSKeyFile:           mtlsKeyFile,
+		MTLSCAFile:            mtlsCAFile,
```
hzxuzhonghu: since you initiated `mTLSCfg`, why not use that struct

cmd/router/main.go
```diff
@@ -36,22 +37,58 @@ func main() {
...
 
+	flag.StringVar(&mtlsCertFile, "mtls-cert-file", "", "Path to mTLS certificate file")
+	flag.StringVar(&mtlsKeyFile, "mtls-key-file", "", "Path to mTLS private key file")
+	flag.StringVar(&mtlsCAFile, "mtls-ca-file", "", "Path to mTLS CA bundle file")
+
 	// Parse command line flags
 	flag.Parse()
 
+	// Validate mTLS configuration early (fail fast on bad flags)
+	mTLSCfg := mtls.Config{
```
hzxuzhonghu:
```suggestion
	tlsConfig:= mtls.Config{
```

cmd/picod/main.go
mahil-2040: You're right, i have reverted it. The previous approach had WorkloadManager reaching into PicoD's container spec to inject `--enable-mtls` / `--mtls-cert-file` flags, which couples infrastructure to the workload's CLI interface.

The fix: PicoD now auto-detects mTLS on startup by checking if the well-known SPIRE cert files exist at `/run/spire/certs/` (provisioned by the spiffe-helper sidecar). If all three SVID files are present, mTLS is enabled automatically, no flags needed. WorkloadManager only injects the sidecar container, volumes, and cert mounts (the standard sidecar injection pattern) and never touches the workload container's args.

cmd/workload-manager/main.go
```diff
@@ -57,15 +58,38 @@ func main() {
 		enableTLS        = flag.Bool("enable-tls", false, "Enable TLS (HTTPS)")
 		tlsCert          = flag.String("tls-cert", "", "Path to TLS certificate file")
 		tlsKey           = flag.String("tls-key", "", "Path to TLS key file")
-		enableAuth       = flag.Bool("enable-auth", false, "Enable Authentication")
+		enableAuth        = flag.Bool("enable-auth", false, "Enable Authentication")
```
hzxuzhonghu: To be clear, this is enabling authorization

cmd/router/main.go
mahil-2040: updated the description!

cmd/workload-manager/main.go
```diff
@@ -57,15 +58,38 @@ func main() {
...
+		enableMTLS        = flag.Bool("enable-mtls", false, "Enable mutual TLS on the WorkloadManager listener")
```
hzxuzhonghu:
```suggestion
		enableMTLS        = flag.Bool("enable-mtls", false, "Whether to enable mutual TLS on the WorkloadManager api server, default to false and start a plain text http server.")
```

hzxuzhonghu: what's the releationship with `enableTLS`, i am thinking they are some kind of duplicate

cmd/workload-manager/main.go
```diff
@@ -57,15 +58,38 @@ func main() {
...
+		enableSandboxMTLS = flag.Bool("enable-sandbox-mtls", false, "Enable automatic injection of SPIRE mTLS sidecars into sandbox pods")
+
+		// mTLS flags (certificate source abstraction)
+		mtlsCertFile, mtlsKeyFile, mtlsCAFile string
```
hzxuzhonghu: we already have tls-key, tls-cert, not reintroduce another flag

cmd/router/main.go
```diff
@@ -36,22 +37,58 @@ func main() {
...
+		// mTLS flags (certificate source abstraction)
+		mtlsCertFile, mtlsKeyFile, mtlsCAFile string
 	)
 
 	// Initialize klog flags
 	klog.InitFlags(nil)
 
+	flag.StringVar(&mtlsCertFile, "mtls-cert-file", "", "Path to mTLS certificate file")
+	flag.StringVar(&mtlsKeyFile, "mtls-key-file", "", "Path to mTLS private key file")
+	flag.StringVar(&mtlsCAFile, "mtls-ca-file", "", "Path to mTLS CA bundle file")
```
hzxuzhonghu: IIUC this is the cert accessing upstream workload manager. So please distinguish with previous `tlsCert `

I think the name is not explicit now

pkg/mtls/config.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// Enabled returns true if all mTLS certificate paths are configured.
+func (c *Config) Enabled() bool {
+	return c.CertFile != "" && c.KeyFile != "" && c.CAFile != ""
+}
+
+// Validate checks that the configuration is internally consistent.
+// If any path is provided, all three must be specified together.
+func (c *Config) Validate() error {
```
hzxuzhonghu: i donot think it make much sense. you donot verify their exist

pkg/mtls/loader.go
hzxuzhonghu: prefer implement all in this pr. so it can be tested

pkg/mtls/loader.go
hzxuzhonghu: please check tls config, tls.Config.GetCertificate

hzxuzhonghu (commented):
Please revert all the picod related change, router -> agent or codeinterpreter, this path is special, we cannot fully control it

hzxuzhonghu:
Please also update the e2e test setup, so mtls can be tested

cmd/workload-manager/main.go
mahil-2040: `enableTLS` and `enableMTLS` serve two different security levels, although they are mutually exclusive in practice:

1. `--enable-tls` is for standard HTTPS (one-way TLS). It uses a standard server certificate (`--tls-cert` and `--tls-key`). It only encrypts the traffic and doesn't strongly verify the identity of the client via the TLS handshake.
2. `--enable-mtls` is for SPIFFE-based mutual TLS. It uses SPIRE-provisioned certificates and strictly enforces that the incoming connection is from a client presenting the `spiffe://.../router` identity (the Router). This provides both transport encryption and strong cryptographic identity verification.

In `pkg/workloadmanager/server.go`, if both are set, mTLS takes precedence (as it's the more secure option). They aren't exactly duplicates since one allows any client (relying on JWT/ServiceAccount auth) while the other strictly limits the transport to authenticated SPIFFE identities.

mahil-2040:
> Please also update the e2e test setup, so mtls can be tested

i am working on e2e tests

cmd/router/main.go
```diff
@@ -25,17 +25,25 @@ import (
...
 	var (
 		port                  = flag.String("port", "8080", "Router API server port")
 		enableTLS             = flag.Bool("enable-tls", false, "Enable TLS (HTTPS)")
-		tlsCert               = flag.String("tls-cert", "", "Path to TLS certificate file")
-		tlsKey                = flag.String("tls-key", "", "Path to TLS key file")
+		tlsCert               = flag.String("tls-cert", "", "Path to TLS certificate file for the Router listener")
+		tlsKey                = flag.String("tls-key", "", "Path to TLS key file for the Router listener")
 		debug                 = flag.Bool("debug", false, "Enable debug mode")
 		maxConcurrentRequests = flag.Int("max-concurrent-requests", 1000, "Maximum number of concurrent requests that a router server can handle (0 = unlimited)")
+		enableMTLS            = flag.Bool("enable-mtls", false, "Enable mutual TLS on the Router listener and for Router-to-WorkloadManager connections")
```
hzxuzhonghu: why do we need this flag

cmd/router/main.go
```diff
@@ -44,6 +52,20 @@ func main() {
...
+	tlsConfig := mtls.Config{
+		CertFile: *mtlsCert,
+		KeyFile:  *mtlsKey,
+		CAFile:   *mtlsCA,
+	}
+	if err := tlsConfig.Validate(); err != nil {
+		klog.Fatalf("Invalid mTLS configuration: %v", err)
+	}
+
+	if *enableMTLS && !tlsConfig.Enabled() {
```
hzxuzhonghu: I donot like such hacky implement, `tlsConfig.Enabled()` ?  Prefer Clean code

cmd/workload-manager/main.go
```diff
@@ -52,12 +53,16 @@ func init() {
...
-		tlsKey           = flag.String("tls-key", "", "Path to TLS key file")
-		enableAuth       = flag.Bool("enable-auth", false, "Enable Authentication")
+		port              = flag.String("port", "8080", "API server port")
+		runtimeClassName  = flag.String("runtime-class-name", "kuasar-vmm", "RuntimeClassName for sandbox pods")
+		enableTLS         = flag.Bool("enable-tls", false, "Enable TLS (HTTPS)")
+		tlsCert           = flag.String("tls-cert", "", "Path to TLS certificate file")
+		tlsKey            = flag.String("tls-key", "", "Path to TLS key file")
+		tlsCA             = flag.String("tls-ca", "", "Path to TLS CA certificate file")
+		enableAuth        = flag.Bool("enable-auth", false, "Enable authorization")
+		enableMTLS        = flag.Bool("enable-mtls", false, "Whether to enable mutual TLS on the WorkloadManager api server, default to false and start a plain text http server.")
```
hzxuzhonghu: can you expalin enableTLS  and enableMTLS?   I think we only need one or neither of them, we can infer tls from certs existence

manifests/charts/base/templates/agentcube-router.yaml
```diff
@@ -22,6 +22,10 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- if .Values.spire.enabled }}
+      securityContext:
+        fsGroup: 1000
```
hzxuzhonghu, line 27: ??

manifests/charts/base/templates/workloadmanager.yaml
```diff
@@ -66,18 +70,34 @@ spec:
...
+            - --tls-cert={{ .Values.spire.spiffeHelper.certDir }}/{{ .Values.spire.spiffeHelper.certFileName }}
+            - --tls-key={{ .Values.spire.spiffeHelper.certDir }}/{{ .Values.spire.spiffeHelper.keyFileName }}
+            - --tls-ca={{ .Values.spire.spiffeHelper.certDir }}/{{ .Values.spire.spiffeHelper.bundleFileName }}
+            {{- end }}
           resources:
             {{- toYaml .Values.workloadmanager.resources | nindent 12 }}
           livenessProbe:
+            {{- if .Values.spire.enabled }}
+            tcpSocket:
+              port: {{ .Values.workloadmanager.service.port }}
```
hzxuzhonghu, line 87: help me understand what port is it?

pkg/mtls/config.go
```diff
@@ -0,0 +1,75 @@
...
+	CertFile string
+	// KeyFile is the path to the mTLS private key (PEM).
+	KeyFile string
+	// CAFile is the path to the mTLS CA bundle for peer verification (PEM).
+	CAFile string
+}
+
+// Enabled returns true if all mTLS certificate paths are configured.
+func (c *Config) Enabled() bool {
+	return c.CertFile != "" && c.KeyFile != "" && c.CAFile != ""
```
hzxuzhonghu: we donot need such method

pkg/mtls/config.go
```diff
@@ -0,0 +1,75 @@
...
+	// Check all-or-nothing: if any path is set, all must be set.
+	set := 0
+	for _, p := range paths {
+		if p != "" {
+			set++
+		}
+	}
+	if set > 0 && set < 3 {
+		return fmt.Errorf("cert, key, and ca must all be specified together")
+	}
```
hzxuzhonghu: clean code

pkg/router/session_manager.go
```diff
@@ -68,10 +74,37 @@ func NewSessionManager(storeClient store.Store) (SessionManager, error) {
...
+	var certWatcher *mtls.CertWatcher
+
+	if enableMTLS {
+		// Ensure the URL uses https:// to prevent plaintext bypass of mTLS
+		if strings.HasPrefix(workloadMgrAddr, "http://") {
+			workloadMgrAddr = strings.Replace(workloadMgrAddr, "http://", "https://", 1)
+			klog.Infof("Rewriting WORKLOAD_MANAGER_URL to use https:// due to mTLS being enabled")
+		} else if !strings.HasPrefix(workloadMgrAddr, "https://") {
+			workloadMgrAddr = "https://" + workloadMgrAddr
+			klog.Infof("Prepending https:// to WORKLOAD_MANAGER_URL due to mTLS being enabled")
```
hzxuzhonghu: Simplify

pkg/mtls/spiffeid.go
hzxuzhonghu: can we remove it now, since no used any more

pkg/mtls/watcher.go
```diff
@@ -0,0 +1,254 @@
...
+
+	// Setup file watcher
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return nil, fmt.Errorf("create file watcher: %w", err)
+	}
+	cw.watcher = watcher
+
+	// Watch the cert file (key file is typically updated atomically with cert)
+	if err := watcher.Add(certFile); err != nil {
```
hzxuzhonghu: This is not correct, we cannot add the file path, we need to add the parent direct, because in k8s the mounted file update is via symlink.

hzxuzhonghu (commented):
My biggest concern is the pr is not very clean code

manifests/charts/base/templates/agentcube-router.yaml
mahil-2040: This is needed because router and workloadmanager run as non-root UID 1000, while spiffe-helper writes the SVID private key into a shared volume with mode **0640**. `fsGroup: 1000` lets the main container read `/run/spire/certs/svid_key.pem` without making the private key world-readable.

manifests/charts/base/templates/workloadmanager.yaml
mahil-2040: This is the WorkloadManager API port, `.Values.workloadmanager.service.port` which is default to `8080`. When SPIRE is enabled, that port serves mTLS, so the existing HTTP `/health` endpoint sends plain HTTP request to an HTTPS/mTLS listener which causes TLS handshake errors. The TCP probe is used only in SPIRE mode to verify the listener is up without requiring a client certificate.

cmd/workload-manager/main.go
mahil-2040: I have completely removed `enableMTLS` flag and used certs existence checks to infer tls.

hzxuzhonghu:
/gemini review

@mahil-2040 please address comments first

hzxuzhonghu:
@mahil-2040 please resolve all the bot comments and obey clean code

manifests/charts/base/templates/spire/cluster-spiffe-ids.yaml
```diff
@@ -30,16 +30,4 @@ spec:
   namespaceSelector:
     matchLabels:
       kubernetes.io/metadata.name: {{ .Release.Namespace }}
```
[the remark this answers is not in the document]
mahil-2040: This PR intentionally scopes mTLS to Router→WorkloadManager only. Router→PicoD/Sandbox traffic is not part of the mTLS boundary, so sandbox pods do not require SPIRE SVID registration here. We can add a separate sandbox ClusterSPIFFEID later if/when Router→PicoD mTLS is introduced.

mahil-2040:
@hzxuzhonghu PTAL!

cmd/workload-manager/main.go
```diff
@@ -181,3 +201,38 @@ func setupControllers(mgr ctrl.Manager, sandboxReconciler *workloadmanager.Sandb
...
+		}
+		return nil
+	}
+	if enableTLS && (cfg.CertFile == "" || cfg.KeyFile == "") {
+		return fmt.Errorf("both --tls-cert and --tls-key must be provided when --enable-tls is set")
+	}
+	return nil
+}
+
+func waitForMTLSFiles(cfg mtls.Config, timeout time.Duration) error {
```
hzxuzhonghu: `waitForMTLSFiles` and `mtlsFilesExist` are copy-pasted verbatim into both `cmd/router/main.go` (line 129) and here. Since `pkg/mtls` is exactly the right place for shared mTLS helpers, could these move there? Keeps the cmd packages thin and avoids the two copies drifting.

hzxuzhonghu (commented):
Big PR, but the core abstraction in `pkg/mtls` is clean and the fsnotify-based cert rotation is a nice touch. The startup wait loop for SPIRE file provisioning solves a real race condition. A couple of structural observations below.

cmd/router/main.go
```diff
@@ -19,23 +19,34 @@ package main
...
 	"syscall"
+	"time"
 
 	"k8s.io/klog/v2"
 
+	"github.com/volcano-sh/agentcube/pkg/mtls"
 	"github.com/volcano-sh/agentcube/pkg/router"
 )
 
+const mtlsFileWaitTimeout = 30 * time.Second
```
hzxuzhonghu: how could the cert provision so slow?  Can you run the spire agent as init container?

manifests/charts/base/templates/agentcube-router.yaml
hzxuzhonghu, line 27: can you explain about why set this?

pkg/mtls/watcher.go
```diff
@@ -0,0 +1,216 @@
...
+		if debounceTimer != nil {
+			debounceTimer.Stop()
+		}
+		debounceTimer = time.AfterFunc(200*time.Millisecond, func() {
+			select {
+			case <-cw.done:
+				return
+			default:
+			}
+			cw.reloadAll()
```
hzxuzhonghu: nit: if there are more than one event triggers, it will  reloadAll multi times. The expected behavior should be once with the debounce

hzxuzhonghu: You may use a debounce channel, use it in the for select loop

```
var debounceC <-chan time.Time
		if debounceC == nil {
					debounceC = time.After(watchDebounceDelay)
		}
```

pkg/mtls/watcher.go
```diff
@@ -0,0 +1,216 @@
...
+// reloadAll reloads the cert/key pair and (if configured) the CA bundle.
+func (cw *CertWatcher) reloadAll() {
+	if err := cw.reload(); err != nil {
+		klog.Errorf("Failed to reload certificate: %v", err)
+	}
+	if cw.caFile != "" {
+		if err := cw.reloadCA(); err != nil {
+			klog.Errorf("Failed to reload CA bundle: %v", err)
+		}
+	}
```
hzxuzhonghu, line 174: make the call atomic

pkg/router/session_manager.go
```diff
@@ -44,19 +45,24 @@ type SessionManager interface {
...
+	// Close cleans up any background resources (like mTLS certificate watchers)
+	Close() error
 }
 
 // manager is the default implementation of the SessionManager interface.
 type manager struct {
 	storeClient     store.Store
 	workloadMgrAddr string
 	httpClient      *http.Client
+	certWatcher     *mtls.CertWatcher // non-nil when mTLS is enabled
```
hzxuzhonghu: Seem you only need to call stop on it. No need to store complete watcher

test/e2e/run_e2e.sh
```diff
@@ -92,6 +100,23 @@ require_python() {
     }
 }
 
+apply_workload_fixture() {
```
hzxuzhonghu, line 114: what does this function do

hzxuzhonghu (commented):
mostly LG, some nits

cmd/router/main.go
mahil-2040: spiffe-helper can’t be an init container because it must keep running to renew/rotate SVID certificates. It already runs in parallel as a sidecar, so certs usually appear quickly; the wait is only a bounded startup-race guard for the brief moment before the initial files are written, and I moved that timeout into `pkg/mtls` with the shared helper.

manifests/charts/base/templates/agentcube-router.yaml
mahil-2040: This is needed so the app container can read the SVID private key written by `spiffe-helper` on the shared volume. `spiffe-helper` writes the key as 0640, so `fsGroup` lets Router/WorkloadManager read it via group permissions without making the key world-readable.

test/e2e/run_e2e.sh
mahil-2040: It makes sure the test YAMLs are created in the namespace we choose with WORKLOAD_NAMESPACE.

The YAML files have a fixed namespace inside them, so this function makes a temporary copy with the right namespace, applies it, and deletes the temp file.

hzxuzhonghu (commented):
Why do you skip test if it is mtls. It should not break the tests

pkg/mtls/loader.go
```diff
@@ -0,0 +1,195 @@
...
+			Roots:         caPool,
+			Intermediates: intermediates,
+			KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
+		}
+		if _, err := peerCert.Verify(opts); err != nil {
+			return fmt.Errorf("verify client certificate chain: %w", err)
+		}
+
+		if len(expectedIDs) > 0 {
+			for _, uri := range peerCert.URIs {
```
hzxuzhonghu: I didnot notice this, I donot think it should verify this. think about if another user want to access workload manager, even it has cert provisioned by spire. It cannot be allowed now

pkg/mtls/watcher.go
```diff
@@ -0,0 +1,237 @@
...
+	caPool := x509.NewCertPool()
+	if !caPool.AppendCertsFromPEM(caCert) {
+		return nil, fmt.Errorf("no valid CA certificates found in %s", cw.caFile)
+	}
+	return caPool, nil
+}
+
+func (cw *CertWatcher) watchLoop() {
+	var debounceC <-chan time.Time
+	scheduleReload := func() {
```
hzxuzhonghu: remove this temp function

manifests/charts/base/templates/agentcube-router.yaml
hzxuzhonghu: I see, this is because we run router and workload manager as user 1000

mahil-2040: Yes

mahil-2040:
> Why do you skip test if it is mtls. It should not break the tests

The MCP code interpreter tests launch a subprocess that runs the MCP server. Inside that subprocess, the `CodeInterpreterClient` (from `sdk-python`) connects directly to the `WorkloadManager `using `WORKLOAD_MANAGER_URL`. When `mTLS` is enabled, the `WorkloadManager` requires a valid **SPIRE-provisioned client certificate** at the TLS layer, but the **MCP subprocess doesn't have any**, so the TLS handshake fails before any application logic runs and e2e tests fails.

This is actually the same issue that already exists with the code interpreter e2e tests (`test_code_interpreter.py`) - they also connect directly to WM and are already being skipped when `mTLS` is enabled. The MCP tests were added later and hit the exact same problem, so I applied the same skip pattern for consistency. So this is a broader concern that affects both test suites, not just MCP.

I think the proper fix for both would be to route through the Router instead of directly to WM. The Router is the gateway - external clients authenticate to it, and it internally forwards to WM using its own SPIRE certs. This way `mTLS` becomes transparent to the client:

`Test subprocess → Router (no mTLS needed) → WorkloadManager (Router handles mTLS)`

However, this would require changes to how `CodeInterpreterClient` in the SDK discovers and connects to its backend, which I think is out of scope for this PR and also the project. So i think it should be raised as an issue and solved separately and for now skipping the tests is only thing i can think of.

Do you have a better approach in mind to handle this?

pkg/mtls/loader.go
mahil-2040: i am confused, can you please clarify if another user (other than router) with valid cert provisioned by spire should be allowed or not?

hzxuzhonghu (commented):
Comprehensive mTLS implementation for Router-to-WorkloadManager communication.

`pkg/mtls/config.go`
- Lines 20-29: Clean config struct. Certificate-source agnostic design is good - works with SPIRE, cert-manager, or self-signed.

`pkg/mtls/loader.go`
- Lines 30-55 (`LoadServerConfig`): Using `RequireAnyClientCert` + manual `VerifyPeerCertificate` for dynamic CA rotation is the correct approach. `RequireAndVerifyClientCert` would use a static CA pool.
- Line 44: TLS 1.3 minimum is good security posture.
- Lines 64-88 (`LoadClientConfig`): `InsecureSkipVerify=true` with manual verification is the standard pattern for SPIFFE ID validation. The comment explaining this is helpful.
- The `expectedServerID` requirement enforces identity verification. Good.

`cmd/router/main.go`
- Lines 40-43: Separate mTLS flags (`--mtls-cert/key/ca`) distinct from the listener TLS flags (`--tls-cert/key`) is the right design. The Router serves clients with one cert and authenticates to WM with another.
- Lines 112-121 (`validateMTLSFlags`): All-or-nothing validation prevents partial configuration.

`cmd/workload-manager/main.go`
- Lines 61-62: Adding `--tls-ca` flag enables the WM to verify client (Router) certificates.
- Lines 199-214 (`validateTLSFlags`): Correct - when CA is provided, all three files are required for mTLS.

`manifests/charts/base/templates/`
- Helm chart updates properly configure SPIRE sidecar integration with mTLS cert/key/CA paths.
- Lines in `workloadmanager.yaml`: Switching health probes from HTTP to TCP when SPIRE is enabled is correct - the probes cannot present a client cert.
- `fsGroup: 1000` in security context allows app containers to read the SPIRE-delivered key files. Good.

`manifests/charts/base/templates/spire/cluster-spiffe-ids.yaml`
- Removing the PicoD ClusterSPIFFEID is intentional since PicoD uses JWT auth, not mTLS. Good cleanup.

`manifests/charts/base/templates/spire/spiffe-helper-config.yaml`
- `key_file_mode = "0640"` allows fsGroup members to read the key. Necessary for the non-root app containers.

`manifests/charts/base/values.yaml`
- Bumping spiffe-helper from 0.8.0 to 0.10.0. Make sure this version is compatible with the SPIRE server version used.
- Updated comment clarifying cert delivery is for Router and WM only (not PicoD).

LGTM. Production-grade mTLS implementation.

pkg/mtls/loader.go
hzxuzhonghu: I think it should be allowed. In the future, we may allow others call workload manager directly.

hzxuzhonghu:
@mahil-2040 Is spire able to issue a certs for non pod workload? So for the failed tests, it would be able to fetch the certs and communicate with WM.

mahil-2040:
> @mahil-2040 Is spire able to issue a certs for non pod workload? So for the failed tests, it would be able to fetch the certs and communicate with WM.

@hzxuzhonghu Yes, SPIRE can issue certificates for non-pod workloads. However, the problem here is that our Python SDK (`CodeInterpreterClient`) has no TLS support (it uses a bare `requests.Session`).

To make this work, we would need to:

1. Update the SDK (`sdk-python/`): Add TLS cert/key/CA parameters to `create_session()`, `ControlPlaneClient`, and `CodeInterpreterClient`.
2. Update the MCP Server: Allow it to load these cert paths from env variables and pass them to the client.
3. Update E2E script (`run_e2e.sh`): Fetch certificates from the SPIRE agent, save them, and pass their paths to the test processes.

Doing this will completely solve the issue and allow us to remove all test skipping (both MCP and code interpreter tests).

Since updating the Python SDK and test harness is quite wide in scope, I think we should address this in a follow-up PR. But if you prefer, I can implement these changes directly in this PR. Let me know what you think!

hzxuzhonghu (approved):
/lgtm

PR #294, opened 2026-04-22
cleanup: remove unused GetCodeInterpreter method

Storm1289:
## Summary

As discussed during review, I verified that `CodeInterpreterReconciler.GetCodeInterpreter` has no production call sites — it was only referenced by its own tests.

This method originally wrapped `mgr.GetCache().Get(...)` with a hardcoded `context.Background()`, and the initial goal of this PR was to fix that context propagation issue. However, since the method is effectively dead code, the cleaner approach is to remove it entirely rather than evolve an unused API.

If a cache-backed CodeInterpreter lookup helper is needed in the future, it can be introduced alongside a real caller to properly define its interface (context usage, error handling, etc.).

## Changes

- Removed `GetCodeInterpreter` from `pkg/workloadmanager/codeinterpreter_controller.go`
- Removed associated tests:
  - `TestGetCodeInterpreter_Found`
  - `TestGetCodeInterpreter_NotFound`
  - `TestGetCodeInterpreter_NilClient`
  - `TestGetCodeInterpreter_ReturnsDeepCopy`
- Removed helper `newTestCodeInterpreter`
- Cleaned up unused imports (`context`, `apierrors`, `metav1`) from the test file

## Test Plan

- [x] `go build ./...`
- [x] Confirmed zero remaining references to `GetCodeInterpreter` in the repo

### DIFF
--- a/pkg/workloadmanager/codeinterpreter_controller.go
+++ b/pkg/workloadmanager/codeinterpreter_controller.go
@@ -334,27 +334,29 @@ func (r *CodeInterpreterReconciler) podTemplateEqual(a, b sandboxv1alpha1.PodTem
 	return reflect.DeepEqual(a.Spec, b.Spec)
 }
 
-// GetCodeInterpreter retrieves a CodeInterpreter from the cache by namespace and name.
-// The cache uses Kubernetes informer cache which is automatically maintained by controller-runtime
-// and stays synchronized with the Kubernetes API server through watch mechanism.
+// GetCodeInterpreter retrieves a CodeInterpreter by namespace and name via the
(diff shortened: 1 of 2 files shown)


hzxuzhonghu (commented):
This is a unused method, can you help cleanup

Storm1289:
Hi @hzxuzhonghu ,thanks for the review,
Confirmed that GetCodeInterpreter isn’t used in production (only in its tests). I’ve removed the method, related tests. We can reintroduce it later if there’s a real use case and updated the describe

hzxuzhonghu (commented):
/lgtm

PR #295, opened 2026-04-23
fix: abort sandbox create when request context is cancelled

Storm1289:
## Summary

`createSandbox` waits for the Sandbox to reach Running state via a `select` on a notification channel and a 2-minute timer. The select had no `ctx.Done()` case, so when the HTTP client disconnects (tab closed, curl Ctrl+C, upstream proxy timeout < 2m), the handler keeps blocking for the full 2 minutes — holding a goroutine, a watcher-map entry in `SandboxReconciler`, and the about-to-be-rolled-back K8s objects the entire time.

Fix the select to observe context cancellation and swap `time.After` for an explicit timer + `Stop()` so the timer doesn't leak when the select exits early.

## Changes

- Add `case <-ctx.Done():` to the wait-for-Running `select` in [`pkg/workloadmanager/handlers.go`](pkg/workloadmanager/handlers.go). Returns `ctx.Err()` wrapped so the existing `sandboxRollback` defer still tears down the placeholder, claim, and sandbox.
- Replace `time.After(2 * time.Minute)` with `time.NewTimer` + `defer timer.Stop()` to free the timer when the channel receives or `ctx` cancels.
- Add `TestServerCreateSandbox_ContextCancelled` that cancels the context before invoking `createSandbox` and asserts it returns with `context.Canceled` in well under 1 second, and that the rollback (`deleteSandbox`) still runs.

## Test plan

- [x] `go build ./...`
- [x] New test `TestServerCreateSandbox_ContextCancelled` aborts in <1s with `context.Canceled`, confirms rollback fired

### DIFF
--- a/pkg/workloadmanager/handlers.go
+++ b/pkg/workloadmanager/handlers.go
@@ -172,8 +172,8 @@ func (s *Server) createSandbox(ctx context.Context, dynamicClient dynamic.Interf
 	}()
 
 	var createdSandbox *sandboxv1alpha1.Sandbox
-	// Use an explicit timer so it is stopped as soon as the channel receives or
-	// ctx cancels, instead of leaking until time.After fires.
+	// Explicit timer + deferred Stop so the runtime timer is released on
+	// function return rather than lingering until the 2m deadline fires.
 	timer := time.NewTimer(2 * time.Minute) // consistent with router settings
 	defer timer.Stop()
 	select {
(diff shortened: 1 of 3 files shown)


hzxuzhonghu (commented):
duplicate with #273

Storm1289:
Hi @hzxuzhonghu,
Just saw the changes on #273 after your message confirmed it's the same fix with broader scope. Closing this PR. Thanks for the pointer.

PR #296, opened 2026-04-24
fix: TokenCache.Get drops stale entries and updates LRU on hit

Storm1289:
## Summary

`TokenCache.Get` ([pkg/workloadmanager/client_cache.go:232](pkg/workloadmanager/client_cache.go#L232)) held an `RLock` and neither removed expired entries nor moved hits to the front of the LRU list. Two effects:

1. **Stale entries leaked.** When a cached auth result expired, `Get` returned `found=false` but left the entry in `c.cache` + `c.lruList`. The slot was held until natural LRU eviction under size pressure pushed it out — so a token the caller will never look up again could keep a fresh, live entry out of the cache.
2. **Not actually LRU.** Because `Get` couldn't mutate (`RLock`), it never called `MoveToFront`. The cache behaved as insertion-ordered (FIFO), contradicting the type comment that calls it an "LRU cache for token validation results" (line 204).

The sibling `ClientCache.Get` in the same file handles both cases correctly — this fix aligns `TokenCache.Get` with that pattern.

## Changes

- `TokenCache.Get`: `RLock → Lock`, drop the expired entry from both `c.cache` and `c.lruList`, `MoveToFront` on a hit.
- `TestTokenCache_LRUBehavior`: updated to assert true LRU semantics. Previously it codified the buggy behavior (`// Get doesn't update LRU, only Set does`). Now a `Get` on `token0` keeps it warm and the least-recently-used `token1` is evicted.
- `TestTokenCache_Get_Expired`: extended to assert `Size() == 0` after an expired `Get`, verifying the stale entry is dropped.

## Impact

On busy clusters with many service-account tokens, the cache now actually honors LRU and prunes expired entries on access. Reduces pointless `TokenReview` API-server load and stops stale entries from evicting live ones.

## Test plan

- [x] `go build ./...`
- [x] `make lint` (golangci-lint v1.64.1)
- [x] `TestTokenCache_ConcurrentAccess` still passes after `RLock → Lock`

### DIFF
--- a/pkg/workloadmanager/client_cache.go
+++ b/pkg/workloadmanager/client_cache.go
@@ -197,8 +197,13 @@ type tokenCacheEntry struct {
 	token         string
 	authenticated bool
 	username      string
-	lastAccess    time.Time
-	element       *list.Element
+	// validatedAt is the wall-clock time the entry was last confirmed against
+	// the Kubernetes TokenReview API (i.e. on Set). TTL is measured from this
+	// time, not from the last Get — reads intentionally do NOT slide the
+	// expiration so a compromised token in active use still gets re-validated
+	// every ttl. This matches ClientCache's use of absolute tokenExpiry.
+	validatedAt time.Time
+	element     *list.Element
 }
 
 // TokenCache is a thread-safe LRU cache for token validation results
@@ -242,7 +247,7 @@ func (c *TokenCache) Get(token string) (found bool, authenticated bool, username
 	// occupy a slot until natural eviction pushes them out. Without this, a
 	// token the caller will never look up again can keep a fresh, live entry
 	// out of the cache under size pressure.
-	if time.Since(entry.lastAccess) > c.ttl {
+	if time.Since(entry.validatedAt) > c.ttl {
 		c.lruList.Remove(entry.element)
 		delete(c.cache, token)
 		return false, false, ""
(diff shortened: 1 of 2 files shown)


Storm1289:
I’ve reviewed all existing PRs and confirmed that this change does not overlap with any ongoing work or follow any existing blueprint. This addresses a separate issue identified during code review.

pkg/workloadmanager/client_cache.go
```diff
@@ -197,8 +197,13 @@ type tokenCacheEntry struct {
...
 	authenticated bool
 	username      string
-	lastAccess    time.Time
-	element       *list.Element
+	// validatedAt is the wall-clock time the entry was last confirmed against
+	// the Kubernetes TokenReview API (i.e. on Set). TTL is measured from this
+	// time, not from the last Get — reads intentionally do NOT slide the
+	// expiration so a compromised token in active use still gets re-validated
+	// every ttl. This matches ClientCache's use of absolute tokenExpiry.
+	validatedAt time.Time
```
hzxuzhonghu, line 205: prefer the old name

hzxuzhonghu (commented):
I donot see a real leak, as in Set method

```
	if len(c.cache) >= c.maxSize {
		c.evictOldest()
	}
```

Storm1289:
Hi @hzxuzhonghu , I am not able to work on this pr, closing it if need please assgin it to someone else :)

PR #298, opened 2026-04-26
test: add e2e session/sandbox lifecycle tests

### LINKED ISSUE #103: Umbrella issues for E2E
## Background

We already have [e2e_test.go](https://github.com/volcano-sh/agentcube/blob/main/test/e2e/e2e_test.go) covering previous ssh-based codeinterpreter:

Now have redesigned AgentCube, bring two scenarios:

- AgentRuntime and CodeInterpreter invocation via Agentcube Router
- Session creation and reuse via x-agentcube-session-id
- Session lifecycle and TTL semantics (at least smoke coverage)

This issue tracks adding a first-stage, minimal but concrete e2e suite for both AgentRuntime and CodeInterpreter.

####  A1. Basic agent invocation: session creation and reuse

Precondition: AgentRuntime CR (e.g. default/echo-agent) deployed.
POST /v1/namespaces/{ns}/agent-runtimes/{name}/invocations/echo without x-agentcube-session-id.

**Assert:**
- HTTP 200.
- Reasonable response body (e.g. echo output).
- Response header x-agentcube-session-id is present.

Repeat call with that session header and a different request body.

Assert:
- Still HTTP 200.
- Behavior consistent with same session (e.g. same session ID; optional state check if agent supports state).

####  A2. Missing/invalid AgentRuntime

- Call POST on a non-existent {name}.
- Assert:
  - HTTP 404 (or the documented error code).
  - Clear error payload (e.g. “AgentRuntime not found”).

#### A3. Idle session / TTL behavior (smoke test)

- Create a session as in A1 and capture SESSION_ID.
- Wait long enough for idle TTL + cleanup.
- Call with the same SESSION_ID.
-Assert:
  - Either error indicating session not found.

#### B1. Simple code execution with session auto-creation

Precondition: CodeInterpreter CR (e.g. default/python-interpreter) deployed.
POST /v1/namespaces/{ns}/code-interpreters/{name}/invocations/run without x-agentcube-session-id.
Body: simple code snippet (e.g. print(1+1)).

- Assert:
  - HTTP 200.
  - Output (e.g. "stdout":"2\n").
  - Response header x-agentcube-session-id is present.

#### B2. Stateful interpreter session (multi-step)

- Reuse the SESSION_ID from B1.
- Second call: code = "x = 10".
- Third call: code = "print(x)".
- Assert:
  - Third response prints 10 (or explicitly assert stateless behavior if we decide CodeInterpreter is stateless initially).
#### B3. File-based workflow via CodeInterpreter (Fibonacci JSON)

- Use CodeInterpreter API to:
- Upload a small Python script (fibonacci.py) into /workspace.
- Execute it to generate /workspace/output.json.
- Retrieve the JSON result.
- Assert:
    - retrieve the result to check.

Corner cases:

1. longer delay task execution
2. warmer pool/non warmpool

JagjeevanAK:
**What type of PR is this?**

/kind feature

Adds `test/e2e/session_lifecycle_test.go` with three net-new lifecycle tests from issue #103 that were not previously covered by `e2e_test.go`. All tests live in the existing `e2e` package and reuse the shared helpers already present in `e2e_test.go`.

Also adds `test/e2e/echo_agent_short_ttl.yaml` — an AgentRuntime CR with `sessionTimeout: 30s` required by the existing `TestAgentRuntimeSessionTTL` (A3) test.

Tests added:

| Function | Scenario | What it asserts |
|---|---|---|
| `TestAgentRuntimeSessionCreationAndReuse` | A1 | First call without `x-agentcube-session-id` auto-creates a session and returns the header; second call with that ID returns the **same** session ID and correct output |
| `TestCodeInterpreterSessionAutoCreation` | B1 | POST to Router without session ID → correct stdout, exit 0, and `x-agentcube-session-id` set in response header |
| `TestCodeInterpreterStatefulSession` | B2 | State written to the shared workspace in one call (via file) is readable in the next call of the same session |

Scenarios already covered by existing tests (no duplication):

| Scenario | Existing test |
|---|---|
| A2 – missing runtime → HTTP 404 | `TestAgentRuntimeErrorHandling` |
| A3 – idle session TTL smoke test | `TestAgentRuntimeSessionTTL` |
| B3 – file-based Fibonacci workflow | `TestCodeInterpreterFileOperations / "download file"` |

Fixes #103

- B2 uses a file written to the sandbox workspace to simulate state persistence across commands (each command runs in a fresh sub-process).
- `echo_agent_short_ttl.yaml` uses the same echo-server logic as `echo_agent.yaml` but with `sessionTimeout: 30s`; apply it before running `TestAgentRuntimeSessionTTL`.

```release-note
NONE
```


JagjeevanAK:
@hzxuzhonghu can you check this PR ?

test/e2e/session_lifecycle_test.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+	if resp.ExitCode != 0 {
+		t.Errorf("B1: expected exit code 0, got %d (stderr: %s)", resp.ExitCode, resp.Stderr)
+	}
+
+	// Assert: session ID header is present.
+	if sessionID == "" {
+		t.Error("B1: expected x-agentcube-session-id header in response, got empty string")
+	} else {
+		t.Logf("B1: session auto-created by Router, session_id=%s", sessionID)
```
acsoto: Can we clean up the auto-created session here? Otherwise this test leaves the CodeInterpreter session alive until TTL expires.

test/e2e/echo_agent_short_ttl.yaml
```diff
@@ -0,0 +1,104 @@
+apiVersion: runtime.agentcube.volcano.sh/v1alpha1
```
acsoto: Do we still need to add this fixture? `run_e2e.sh` already creates `echo-agent-short-ttl` by generating it from `echo_agent.yaml`, so this file does not seem to be used by the E2E setup.

test/e2e/session_lifecycle_test.go
JagjeevanAK: okay thank you

test/e2e/echo_agent_short_ttl.yaml
JagjeevanAK: done 👍

JagjeevanAK:
can you check the recent updates ?

CC: @acsoto

acsoto:
> can you check the recent updates ?
> 
> CC: @acsoto

Thank you. E2E Tests failed. Cloud you pls check?

JagjeevanAK:
Hey man Sorry for previous stuff now I think all test should pass easily

CC: @acsoto

acsoto:
LGTM

JagjeevanAK:
Any update on this ?

CC: @YaoZengzeng and @tjucoder

pkg/router/session_manager.go
```diff
@@ -185,6 +185,10 @@ func (m *manager) createSandbox(ctx context.Context, namespace string, name stri
 
 	// Construct Sandbox Info from response
 	sandbox := &types.SandboxInfo{
+		// Auto-created sessions come directly from WorkloadManager response, which
+		// currently does not include kind. Default to Sandbox so Router signs JWT
+		// when forwarding to sandbox runtimes (e.g. PicoD).
+		Kind:        types.SandboxKind,
```
hzxuzhonghu, line 191: This field is missing set, because we cannot know whether it is sandbox or sandboxclaim now. 

Do you see any problem without it?

test/e2e/session_lifecycle_test.go
```diff
@@ -0,0 +1,306 @@
...
+	t.Logf("B2: stateful session verified – printed x=%q", strings.TrimSpace(resp2.Stdout))
+}
+
+// ============================================================
+// Helpers used only in this file
+// ============================================================
+
+// invokeCodeInterpreterWithHeader is like invokeCodeInterpreter but also returns
+// the x-agentcube-session-id response header so B1 can assert on it.
+func (e *testEnv) invokeCodeInterpreterWithHeader(
```
hzxuzhonghu, line 256: prefer not make it a method

test/e2e/session_lifecycle_test.go
```diff
@@ -0,0 +1,306 @@
...
+// ============================================================
+// B1 – CodeInterpreter: session auto-creation via Router
+// ============================================================
+
+// TestCodeInterpreterSessionAutoCreation verifies that a POST to the Router
+// code-interpreter endpoint *without* an x-agentcube-session-id header:
+//  1. Returns HTTP 200.
+//  2. Produces the expected stdout (e.g. "2\n" for print(1+1)).
+//  3. Sets the x-agentcube-session-id response header.
+func TestCodeInterpreterSessionAutoCreation(t *testing.T) {
```
hzxuzhonghu, line 125: isn't this included in `TestAgentRuntimeSessionCreationAndReuse`

PR #299, opened 2026-04-27
fix: register sandbox rollback defer before workload create to plug placeholder leak

Sanchit2662:
## What type of PR is this?
/kind bug

The existing TestServerCreateSandbox test cases for failure paths (sandbox creation fails, sandbox claim creation fails, etc.) were not asserting that the store placeholder gets cleaned up during rollback. fakeStore.DeleteSandboxBySessionID was a silent no-op , so even if rollbackSandboxCreation stopped calling DeleteSandboxBySessionID, no test would catch
  it.
  
  This PR closes that gap:

  - Adds a deleteSessionCalls counter to fakeStore so DeleteSandboxBySessionID calls are tracked
  - Adds expectStoreDeleteCalls to the test case struct and asserts it is 1 for every error path that triggers rollback
  - Patches deleteSandboxClaim in the test setup to prevent a nil-client panic when the claim rollback path runs during the claim-failure case
  - Renames the two affected test cases to make the expected rollback behavior explicit in the name                                                                                      

The store placeholder fails case correctly remains at zero, since the defer is registered only after StoreSandbox succeeds.                                                            

### Does this PR introduce a user-facing change?

```release-note
NONE
```

### DIFF
--- a/pkg/workloadmanager/handlers.go
+++ b/pkg/workloadmanager/handlers.go
@@ -220,6 +220,7 @@ func (s *Server) createSandbox(ctx context.Context, dynamicClient dynamic.Interf
 	// preventing the runtime from retaining the timer until it fires.
 	timer := time.NewTimer(2 * time.Minute) // consistent with router settings
 
+
 	var createdSandbox *sandboxv1alpha1.Sandbox
 	select {
 	case result := <-resultChan:
--- a/pkg/workloadmanager/handlers.go
+++ b/pkg/workloadmanager/handlers.go
@@ -220,7 +220,6 @@ func (s *Server) createSandbox(ctx context.Context, dynamicClient dynamic.Interf
 	// preventing the runtime from retaining the timer until it fires.
 	timer := time.NewTimer(2 * time.Minute) // consistent with router settings
 
-
 	var createdSandbox *sandboxv1alpha1.Sandbox
 	select {
 	case result := <-resultChan:
--- a/pkg/workloadmanager/handlers_test.go
+++ b/pkg/workloadmanager/handlers_test.go
@@ -44,10 +44,11 @@ import (
 
 type fakeStore struct {
 	store.Store
-	storeErr    error
-	updateErr   error
-	storeCalls  int
-	updateCalls int
+	storeErr           error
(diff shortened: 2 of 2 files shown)


pkg/workloadmanager/handlers.go
```diff
@@ -149,6 +149,22 @@ func (s *Server) createSandbox(ctx context.Context, dynamicClient dynamic.Interf
 		return nil, err
 	}
 
+	// Register rollback IMMEDIATELY after the placeholder is committed, before
```
acsoto: Nit: can we trim this comment down a bit? The key point is that rollback must be registered right after the store placeholder exists, and that it is safe before the workload exists because deletes ignore NotFound. The longer GC/create-path explanation is already covered by the PR description.

Sanchit2662: Done!

hzxuzhonghu (commented):
Please change the pr description，then we can merge

Sanchit2662:
I have changed the pr description.

pkg/workloadmanager/handlers_test.go
```diff
@@ -242,13 +252,16 @@ func TestServerCreateSandbox(t *testing.T) {
 				claim = &extensionsv1alpha1.SandboxClaim{ObjectMeta: metav1.ObjectMeta{Name: sb.Name, Namespace: sb.Namespace}}
 			}
 
+
```
hzxuzhonghu: Two consecutive blank lines snuck in here. `gofmt` will flag this — can you trim to one?

hzxuzhonghu (commented):
This is on the right track — tracking `DeleteSandboxBySessionID` calls and verifying the rollback on store failures makes the test much more precise. One note below.

pkg/workloadmanager/handlers_test.go
Sanchit2662: Done!

Sanchit2662:
@hzxuzhonghu , all feedback addressed ready for another look whenever you have time

RainbowMango (approved):
/approve

PR #307, opened 2026-05-08
[BUG] wrap underlying errors with %w

krrishrastogi05:
/kind bug

This PR wraps two underlying errors with `%w` instead of formatting them with `%v`, preserving error context for callers that use `errors.Is` or `errors.As`.

`go test ./pkg/store` passes. `go test ./pkg/picod` was attempted, but fails on this Windows setup due to missing Unix commands like `sh`/`sleep` and symlink privilege restrictions.

```release-note
NONE


hzxuzhonghu (approved):
/lgtm

PR #308, opened 2026-05-08
[BUG] make PicoD execute tests portable

krrishrastogi05:
/kind cleanup

While testing a small PicoD change locally on Windows, I noticed that `go test ./pkg/picod` failed for reasons unrelated to the code under test. Some tests assumed Unix tools like `sh`, `sleep`, `pwd`, `true`, and `false` were available, and symlink tests failed when Windows denied symlink creation privileges.

Original behavior (Before changes):
image

This PR makes those tests more portable without changing PicoD runtime behavior. It should make the project a bit more welcoming for contributors who are developing on Windows, since they can run the PicoD test suite locally without needing a Unix-like shell setup or elevated symlink permissions.

The main changes are:
- use a small Go test helper process instead of host-specific shell commands
- keep the same coverage for stdout, stderr, exit codes, timeouts, working directory handling, and env vars
- skip symlink-specific assertions only when the OS refuses symlink creation

After the changes:

image

This is a test-only cleanup. The production PicoD command execution path is unchanged.

**Testing done**:
- `gofmt`
- `git diff --check`
- `go test ./pkg/picod`

```release-note
NONE

### DIFF
--- a/pkg/picod/test_helpers_test.go
+++ b/pkg/picod/test_helpers_test.go
@@ -1,0 +1,171 @@
+/*
+Copyright The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package picod
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"runtime"
+	"strconv"
+	"strings"
+	"syscall"
+	"testing"
+	"time"
+)
+
+const testHelperProcessEnv = "GO_WANT_PICOD_HELPER_PROCESS"
+
+func init() {
+	if os.Getenv(testHelperProcessEnv) == "1" {
+		picodHelperProcessMain()
+		os.Exit(0)
+	}
+}
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pkg/picod/test_helpers_test.go [hunk 1]
```diff
@@ -0,0 +1,137 @@
...
+			os.Exit(2)
+		}
+		os.Exit(code)
+	default:
+		fmt.Fprintf(os.Stderr, "unknown helper command: %s\n", args[0])
+		os.Exit(2)
+	}
+
+	os.Exit(0)
+}
```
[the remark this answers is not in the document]
krrishrastogi05: i have addressed these changes please do recheck

pkg/picod/test_helpers_test.go
```diff
@@ -0,0 +1,171 @@
...
+	"syscall"
+	"testing"
+	"time"
+)
+
+const testHelperProcessEnv = "GO_WANT_PICOD_HELPER_PROCESS"
+
+func init() {
+	if os.Getenv(testHelperProcessEnv) == "1" {
+		picodHelperProcessMain()
```
hzxuzhonghu, line 35: not sure i understand this

krrishrastogi05: Hey @hzxuzhonghu ,

This is for the child helper process used by the execute tests. The tests re-run the package’s own test binary as the command being executed, and GO_WANT_PICOD_HELPER_PROCESS=1 tells that child process to run the helper logic instead of the normal tests.

I put it in init() so it runs before the Go test runner writes anything to stdout/stderr. That keeps the captured output clean for assertions like exact resp.Stdout checks.

I’ve added a short comment above this block to make the intent clearer. If you’d prefer another helper-test pattern here, I’m happy to adjust.

hzxuzhonghu:
It is intended to not run on windows. There are kinds of gaps.

PR #311, opened 2026-05-09
fix(sdk-python): correct router port and bootstrap session race

Abhinav-kodes:
## Summary
Two fixes found while running the getting-started examples end-to-end.

### Fix 1 — Wrong port in example
`agent_runtime_usage.py` hardcoded port `18081` but the documented
`kubectl port-forward` command maps the router to `8081`. This caused
`Connection refused` for anyone following the getting-started guide.

### Fix 2 — Bootstrap session race on non-2xx agent responses
`bootstrap_session_id()` called `raise_for_status()` before reading
the session header. The AgentCube Router injects `x-agentcube-session-id`
even when the upstream agent returns a non-2xx response (e.g. `404`
for `GET /` on agents that only handle `POST`). This caused the client
to crash before it could read the valid session ID.

**Fix**: Read the header first. Only surface the HTTP error if the
header is absent.

## Testing
Verified end-to-end with `kind` cluster using the getting-started guide.

### DIFF
--- a/sdk-python/agentcube/clients/agent_runtime_data_plane.py
+++ b/sdk-python/agentcube/clients/agent_runtime_data_plane.py
@@ -56,14 +56,19 @@ class AgentRuntimeDataPlaneClient:
             self.base_url,
             timeout=(self.connect_timeout, self.timeout),
         )
-        resp.raise_for_status()
-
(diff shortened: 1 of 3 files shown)


hzxuzhonghu (approved):
/lgtm

PR #314, opened 2026-05-10
fix: simplify sandbox pod lookup

### LINKED ISSUE #277: Simplify sandbox pod match
After 
https://github.com/kubernetes-sigs/agent-sandbox/pull/272 merged, each sanbox will have an annotation pointed to the pod.

So we donot need to do `label selector` matching, not this depend on agent-sandbox v0.3.10

```
// GetSandboxPodIP gets the IP address of the pod corresponding to the Sandbox
func (c *K8sClient) GetSandboxPodIP(_ context.Context, namespace, sandboxName, podName string) (string, error) {
	// If podName is provided, try to get it directly from cache first
	if podName != "" {
		pod, err := c.podLister.Pods(namespace).Get(podName)
		if err == nil && pod != nil {
			return validateAndGetPodIP(pod)
		}
		klog.Infof("failed to get sandbox pod %s/%s: %v, try get pod by sandbox-name label", namespace, podName, err)
	}
	// Find pod through label selector (sandbox-name label we set)
	pods, err := c.podLister.Pods(namespace).List(labels.SelectorFromSet(map[string]string{SandboxNameLabelKey: sandboxName}))
	if err != nil {
		return "", fmt.Errorf("failed to list pods from cache: %w", err)
	}
	// Find the pod that belongs to this sandbox by checking ownerReferences
	for _, pod := range pods {
		for _, ownerRef := range pod.OwnerReferences {
			if ownerRef.Kind == "Sandbox" && ownerRef.Name == sandboxName {
				if ownerRef.Controller == nil || *ownerRef.Controller {
					return validateAndGetPodIP(pod)
				}
			}
		}
	}

	return "", fmt.Errorf("no pod found for sandbox %s", sandboxName)
}```

### PR DESCRIPTION
**Pull request by krrishrastogi05**
/kind bug

Updates `sigs.k8s.io/agent-sandbox` to `v0.3.10` and uses `SandboxPodNameAnnotation` as the source of truth for the backing pod name.

After kubernetes-sigs/agent-sandbox#272, Sandboxes are annotated with their pod name, so the workload manager no longer needs to list pods by label and then verify ownerReferences. This keeps sandbox pod lookup simpler and direct through the pod lister/cache.

Fixes #277

This removes the label selector + ownerReference fallback from `GetSandboxPodIP`. Unit tests cover direct pod-name lookup, missing annotations, invalid pod status, and the no-fallback behavior when a label-matched pod exists but the annotated pod is missing.

**Testing done**:
- `go test ./pkg/workloadmanager`
- `golangci-lint run ./pkg/workloadmanager`
- `git diff --check`

```release-note
NONE
```


hzxuzhonghu:
@krrishrastogi05 I think this rely on the agent-sandbox version, maybe the e2e failure related

krrishrastogi05:
@hzxuzhonghu 

After your comment I tried to rework the PR properly around `agent-sandbox v0.3.10`, not just change the lookup function. I tried different approaches, worked like for hours on this but I have been getting E2E tests failed again and again. I would like to have your advice on how I can fix this.

So far I have:
bumped `agent-sandbox` to `v0.3.10`
updated the Go / CI / Docker versions needed for that bump
changed the pod lookup to use the new pod-name annotation
removed the old label selector + ownerRef fallback
updated the related unit tests
also tried to handle the warm-pool case where a claim can point to an already existing/adopted sandbox with a different name

The CI exposed changes step by step, so I had to force-push a few times. First it was dependency/toolchain mismatch, then Docker/CI Go version mismatch, then warm-pool claim name resolution. 

Could you please point me in the right direction here?

go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/volcano-sh/agentcube
 
-go 1.24.4
```
hzxuzhonghu, line 3: I think this is unreleated change

pkg/workloadmanager/handlers.go
```diff
@@ -172,6 +168,19 @@ func (s *Server) handleSandboxCreate(c *gin.Context, kind string) {
 	respondJSON(c, http.StatusOK, response)
 }
 
+func (s *Server) watchSandboxReady(ctx context.Context, namespace, sandboxName string, enabled bool) (<-chan SandboxStatusUpdate, func()) {
+	if !enabled {
+		return nil, func() {}
+	}
```
hzxuzhonghu, line 174: can not understand it

FAUST-BENCHOU:
friendly ping @krrishrastogi05 .Are u still working on it?

PR #315, opened 2026-05-10
feat(cli): add env support in agent_metadata.yaml

### LINKED ISSUE #222: Add env support in agent_metadata.yaml
`WORKLOAD_MANAGER_URL` and `ROUTER_URL` in podTemplate, 
Image

but there's some need to inject extra env variable for agent, the only way for me to implement that is updating crd after execute `kubectl agentcube publish xxx`, and it will be overrideed after re-publish.

**Why is this needed**:
Provide more convenience for the scenarios that need extra env when executing agent

madmecodes:
## Summary

Adds first-class `env` support to `agent_metadata.yaml` so users can declare extra environment variables that get injected into sandbox pod containers during publish.

Right now the only way to get custom env vars (like `LLM_BASE_URL`, `LLM_API_KEY`) into a deployed agent is to manually patch the CRD after `kubectl agentcube publish`, and that gets overwritten on the next publish. This change lets you put them straight in the metadata file:

```yaml
agent_name: my-agent
entrypoint: python main.py
port: 8080
env:
  LLM_BASE_URL: https://api.example.com
  LLM_API_KEY: sk-xxx
```

### What changed

- **`metadata_service.py`** -- added `env: Optional[Dict[str, str]]` field to `AgentMetadata` model. Defaults to `None`, excluded from saved YAML when absent.
- **`publish_runtime.py`** -- both `_publish_cr_to_k8s` (AgentRuntime CR) and `_publish_k8s` (standard Deployment) now read `metadata.env` and merge it into the `env_vars` dict before passing to providers. Programmatic `env_vars` from options still override metadata values.
- **`tests/test_env_support.py`** -- 10 unit tests covering model validation, YAML round-trip, and env forwarding to both providers.

Fixes #222

## Test plan

- [x] `AgentMetadata` accepts `env` dict, defaults to `None`
- [x] YAML load/save round-trips `env` correctly
- [x] Saving metadata without `env` omits the field from YAML
- [x] `env` from metadata forwarded to AgentCube provider
- [x] `env` from metadata forwarded to K8s provider
- [x] No `env` passes `None` to providers (no regression)
- [x] All 10 tests pass locally

### DIFF
--- a/cmd/cli/agentcube/runtime/publish_runtime.py
+++ b/cmd/cli/agentcube/runtime/publish_runtime.py
@@ -150,6 +150,11 @@ class PublishRuntime:
                 "Please ensure 'readiness_probe_path' and 'readiness_probe_port' are set in agent_metadata.yaml."
             )
 
+        # Merge user-defined env from metadata with any env_vars from options
+        env_vars = dict(metadata.env) if metadata.env else {}
+        if options.get('env_vars'):
+            env_vars.update(options['env_vars'])
+
         # Step 3: Deploy AgentRuntime CR
         try:
             k8s_info = self.agentcube_provider.deploy_agent_runtime(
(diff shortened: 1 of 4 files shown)


hzxuzhonghu (approved):
/lgtm

Should we update document as well like docs/design/AgentRun-CLI-Design.md

hzxuzhonghu:
LLM_API_KEY should better reference a secret key in k8s spec

warjiang:
🚀  cool, need this feature ~

PR #320, opened 2026-05-11
fix(workloadmanager): prevent deletion of healthy sandbox on store cache update failure

Abhinav-kodes:
**What type of PR is this?**

/kind bug

Currently, in `createSandbox`, if a sandbox successfully starts, receives a Pod IP, and has its entrypoints verified, a subsequent failure in `s.storeClient.UpdateSandbox` will trigger the deferred rollback function. This results in the healthy Kubernetes pod being deleted purely due to a transient cache/store write failure.

Because the K8s cluster is the source of truth and the workload is already running and verified, it should not be destroyed if the secondary store fails to update.

This PR:
1. Moves the `needRollbackSandbox = false` clearance *above* the `UpdateSandbox` call to prevent destroying verified, running K8s workloads.
2. Adds a targeted cleanup step (`DeleteSandboxBySessionID`) using a new, detached context if `UpdateSandbox` fails. This ensures we don't leave a stale "placeholder" entry in the store, while safely leaving the pod alive.

* **Testing Change:** The existing test `TestServerCreateSandbox/update_store_fails_triggers_rollback` originally expected `1` delete call, which effectively enforced the bug. I updated this test to expect `0` delete calls to enforce the correct architectural behavior (a healthy pod should not be deleted on a store update failure).
* **Context Handling:** The store placeholder cleanup explicitly uses a new `context.WithTimeout(context.Background(), ...)` because the original request `ctx` may have already been canceled (e.g., client disconnected/timed out), which would cause the cleanup to fail instantly if reused.

```release-note
Fixed an issue where a successfully created sandbox could be erroneously deleted due to a transient internal database/cache error.
```

### DIFF
--- a/pkg/workloadmanager/handlers.go
+++ b/pkg/workloadmanager/handlers.go
@@ -270,14 +270,21 @@ func (s *Server) createSandbox(ctx context.Context, dynamicClient dynamic.Interf
 		EntryPoints: storeCacheInfo.EntryPoints,
 	}
 
+	needRollbackSandbox = false
 	if err := s.storeClient.UpdateSandbox(ctx, storeCacheInfo); err != nil {
+
+		cleanupCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		if delErr := s.storeClient.DeleteSandboxBySessionID(cleanupCtx, sandboxEntry.SessionID); delErr != nil {
+			klog.Errorf("sandbox %s/%s failed to clean up store placeholder for session %s: %v", sandbox.Namespace, sandbox.Name, sandboxEntry.SessionID, delErr)
+		}
+
 		if isContextError(err) {
 			return nil, err
 		}
 		return nil, api.NewInternalError(fmt.Errorf("update store cache failed: %w", err))
 	}
 
-	needRollbackSandbox = false
 	klog.V(2).Infof("init sandbox %s/%s successfully, kind: %s, sessionID: %s", createdSandbox.Namespace,
 		createdSandbox.Name, createdSandbox.Kind, sandboxEntry.SessionID)
 	return response, nil
--- a/pkg/workloadmanager/handlers.go
+++ b/pkg/workloadmanager/handlers.go
@@ -219,23 +219,9 @@ func (s *Server) createSandbox(ctx context.Context, dynamicClient dynamic.Interf
 		return nil, err
 	}
 
-	// Use NewTimer so we can stop it explicitly when another branch wins,
-	// preventing the runtime from retaining the timer until it fires.
-	timer := time.NewTimer(2 * time.Minute) // consistent with router settings
-
-	var createdSandbox *sandboxv1alpha1.Sandbox
-	select {
-	case result := <-resultChan:
-		timer.Stop()
-		createdSandbox = result.Sandbox
-		klog.V(2).Infof("sandbox %s/%s reported ready, verifying entrypoints", createdSandbox.Namespace, createdSandbox.Name)
-	case <-ctx.Done():
-		timer.Stop()
-		klog.Warningf("sandbox %s/%s wait canceled: %v", sandbox.Namespace, sandbox.Name, ctx.Err())
-		return nil, ctx.Err()
-	case <-timer.C:
-		klog.Warningf("sandbox %s/%s create timed out", sandbox.Namespace, sandbox.Name)
-		return nil, errSandboxCreationTimeout
+	createdSandbox, err := waitForSandboxReady(ctx, resultChan, sandbox.Namespace, sandbox.Name)
+	if err != nil {
+		return nil, err
 	}
 
 	// agent-sandbox create pod with same name as sandbox if no warmpool is used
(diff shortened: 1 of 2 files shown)


0xA8hinav:
@hzxuzhonghu , I have applied the changes suggested by copilot and ignored a could be "false positive", the pr is ready to be merged from my end

pkg/workloadmanager/handlers.go
```diff
@@ -267,19 +253,48 @@ func (s *Server) createSandbox(ctx context.Context, dynamicClient dynamic.Interf
 		EntryPoints: storeCacheInfo.EntryPoints,
 	}
 
+	needRollbackSandbox = false
 	if err := s.storeClient.UpdateSandbox(ctx, storeCacheInfo); err != nil {
+		client := s.storeClient
+		go func(sessionID, ns, name string) {
+			cleanupCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			defer cancel()
+			if delErr := client.DeleteSandboxBySessionID(cleanupCtx, sessionID); delErr != nil {
```
hzxuzhonghu: why donot we delete sandbox from k8s, isn't it a leak? WE donot use it later

pkg/workloadmanager/handlers.go
```diff
@@ -270,19 +256,48 @@ func (s *Server) createSandbox(ctx context.Context, dynamicClient dynamic.Interf
 		EntryPoints: storeCacheInfo.EntryPoints,
 	}
 
+	needRollbackSandbox = false
 	if err := s.storeClient.UpdateSandbox(ctx, storeCacheInfo); err != nil {
+		client := s.storeClient
+		go func(sessionID, ns, name string) {
+			cleanupCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
hzxuzhonghu: This `10*time.Second` hardcode will conflict with PR #321 which defines `storeCleanupTimeout = 30 * time.Second` for exactly this purpose. They should use the same constant so cleanup timeouts stay in sync across the codebase.

hzxuzhonghu (commented):
Good fix — protecting running pods from a transient store failure makes total sense since K8s is the source of truth. The `waitForSandboxReady` extraction is also a nice cleanup.

0xA8hinav:
@gemini-code-assist @Copilot , the architectural suggestions regarding the K8s rollback are incorrect in this context.

If `UpdateSandbox` experiences a *permanent* failure (exhausting all retries), the `createSandbox` function returns a 500 Internal Server Error to the client. This means the client **never receives the `SessionID`**. 

If we were to clear the `needRollbackSandbox` flag and leave the K8s pod running (as suggested), that pod would become an orphaned "zombie pod". The client cannot access it (no SessionID), and the system's Garbage Collector cannot find it (no store placeholder). 

Therefore, the correct architectural flow must be:
1. Wrap the store update in a retry loop to protect the pod from *transient* network drops.
2. If the retries succeed, clear the rollback flag and keep the healthy K8s pod.
3. If the retries *permanently* fail, we **must** execute the deferred rollback to delete the K8s pod and prevent a permanent resource leak.

My recent pushes establish exactly this logic. The test `update_store_fails_triggers_retries_and_full_rollback` correctly expects 1 K8s deletion because a hard store failure must trigger the rollback to prevent the zombie pod.

**Note:** I have applied Copilot's medium-severity suggestion regarding the backoff timer. The retry loop now uses `time.NewTimer` and a `select` block to properly respect `ctx.Done()` instead of blocking via `time.Sleep`.

pkg/workloadmanager/handlers_test.go
```diff
@@ -137,44 +143,49 @@ func TestServerCreateSandbox(t *testing.T) {
...
 		},
 		{
-			name:              "update store fails triggers rollback",
-			updateErr:         errors.New("update failed"),
-			sendResult:        true,
-			expectErr:         true,
-			expectCreateCalls: 1,
-			expectUpdateCalls: 1,
-			expectDeleteCalls: 1,
+			name:                   "update store fails triggers retries and full rollback",
```
acsoto: Can we add a companion case where `UpdateSandbox` fails once or twice and then succeeds? That is the transient failure path this PR is trying to protect, and the current fake store only covers the permanent-failure path where rollback still happens.

hzxuzhonghu (commented):
This PR correctly addresses the zombie-pod scenario by adding retry logic and preserving rollback on permanent failure.

`pkg/workloadmanager/handlers.go`
- Lines 259-279 (`updateSandboxWithRetry`): Retry logic is clean. The 500ms backoff between attempts is reasonable. Hardcoded retry count of 3 is fine for now; consider making it configurable if this proves insufficient in production.
- Lines 282-298 (`waitForSandboxReady`): Good extraction of the select block into a named function. Using `defer timer.Stop()` is cleaner than the previous manual stop on each branch.
- Lines 263-265: The timer-based backoff correctly checks `ctx.Done()` to avoid sleeping through a canceled context. Good.
- Lines 269-272: On permanent failure, the comment clearly explains why the K8s rollback is intentionally preserved. This is the core fix.

`pkg/workloadmanager/handlers_test.go`
- The `fakeStore` now supports per-call error sequences via `updateErrs []error`, which enables testing transient-then-success scenarios. Good.
- `deleteStoreCalls` using `atomic.Int32` is correct for potential concurrent access in deferred rollback.
- Test cases "update store fails permanently" (3 retries exhausted, rollback triggered) and "fails once/twice then succeeds" (no rollback) directly verify the fix.

Note: `waitForSandboxReady` is a free function here but PR #353 makes it a method receiver. These PRs will conflict on merge - coordinate which approach to keep.

LGTM.

0xA8hinav:
Thanks for flagging this @hzxuzhonghu
I'll rebase `feat/prometheus-metrics` on `main` once #320 merges and resolve the conflicts there — specifically the `waitForSandboxReady` free-function vs. method form and the `createSandbox` signature.

0xA8hinav:
cc @hzxuzhonghu

PR #321, opened 2026-05-11
fix(workloadmanager): use detached context for store delete and add missing log

Abhinav-kodes:
**What type of PR is this?**

/kind bug

handleDeleteSandbox uses c.Request.Context() for both the K8s deletion and
the subsequent store deletion. If the client disconnects after the K8s
resource is successfully deleted but before DeleteSandboxBySessionID runs,
the request context is already canceled. The store call fails instantly,
leaving a stale entry permanently pointing to a K8s resource that no longer
exists. Future GET or DELETE calls for that sessionID will return stale data
or fail with a misleading error.

This PR fixes the issue by using a detached context.WithTimeout for the store
delete, matching the pattern already established in rollbackSandboxCreation.

It also adds a missing klog.Errorf before the respondError on store delete
failure (Every other store error in this handler logs before responding, but
this path was silently swallowing the error, making store failures during
deletion invisible in production diagnostics).

The detached context uses a 30s timeout, consistent with rollbackSandboxCreation
which uses the same value for identical store cleanup operations.

The K8s deletion calls intentionally retain c.Request.Context(). A client
disconnect does not cancel an already-dispatched K8s API call server-side,
so the request context is appropriate there. Only the store write after
K8s deletion is at risk from a canceled context.

These two fixes are independent but co-located in the same function, so
they are included in a single PR to keep the diff minimal.

```release-note
Fixed a bug where a client disconnect during sandbox deletion could leave
a stale store entry after the Kubernetes resource was already deleted,
causing subsequent delete or lookup calls for that session to fail or
return incorrect state.
```

### DIFF
--- a/pkg/workloadmanager/handlers_test.go
+++ b/pkg/workloadmanager/handlers_test.go
@@ -475,3 +475,59 @@ func TestHandleSandboxCreate(t *testing.T) {
 		})
 	}
 }
+
+/*
+This test verifies that the deleteSandbox handler correctly handles scenarios where the client disconnects (Context Cancellation) before the deletion operation completes.
+
+Key Points:
+
+The handler creates a new context for the K8s deletion operation using context.WithTimeout(ctx, deletionTimeout).
+This derived context remains valid even if the parent context (c.Request.Context()) is canceled.
+The test simulates a client disconnect by canceling the request context immediately after calling the deleteSandbox function.
+It verifies that the store deletion (the final cleanup step) still occurs by checking that the store's DeleteSandboxBySessionID method was called with a valid, non-canceled context.
+*/
+func TestHandleDeleteSandbox_DetachedContext(t *testing.T) {
+	gin.SetMode(gin.TestMode)
+	fakeServer := newFakeServer()
+
+	fakeStoreInst := &fakeStore{}
+	fakeServer.storeClient = fakeStoreInst
+
+	patches := gomonkey.NewPatches()
+	defer patches.Reset()
+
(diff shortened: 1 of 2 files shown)


0xA8hinav:
@hzxuzhonghu , I have applied all the changes suggested by copilot and the pr is ready to be merged from my end

pkg/workloadmanager/handlers_test.go
```diff
@@ -475,3 +475,59 @@ func TestHandleSandboxCreate(t *testing.T) {
...
+		require.NoError(t, ctx.Err(), "Store deletion context MUST NOT be canceled despite client disconnect")
+		storeDeleteCalled = true
+		return nil
+	})
+
+	fakeServer.handleDeleteSandbox(c)
+
+	require.True(t, storeDeleteCalled, "DeleteSandboxBySessionID should be called even if the request context is canceled")
+	require.Equal(t, http.StatusOK, w.Code)
+}
```
hzxuzhonghu, line 533: File is missing a trailing newline — most editors and `gofmt`/`goimports` will flag this. Easy to miss in a PR diff.

pkg/workloadmanager/handlers_test.go
```diff
@@ -475,3 +475,59 @@ func TestHandleSandboxCreate(t *testing.T) {
 		})
 	}
 }
```
hzxuzhonghu, line 477: Go convention is to use `//` line comments for inline documentation, even for multi-line blocks. This `/* ... */` block comment is a bit unusual in Go test files — would be cleaner as a series of `//` lines above the function.

hzxuzhonghu (commented):
The detached context for store delete is the right call — good catch that a client disconnect post-K8s-deletion would otherwise orphan the store entry. The error log before the 500 response is also a welcome addition.

PR #325, opened 2026-05-12
fix(cli): include models package in wheel

mrinalchaturvedi27:
# Description

**What type of PR is this?**

/kind bug

## What I fixed

I fixed the Python CLI wheel packaging so `agentcube.models` is included after a normal wheel install. The CLI entry point can now import `agentcube.models.pack_models` instead of crashing with `ModuleNotFoundError`.

## How I found it

I started from the reported importers in `cmd/cli/agentcube/cli/main.py`, `cmd/cli/agentcube/runtime/pack_runtime.py`, and `cmd/cli/agentcube/runtime/__init__.py`. The source tree had `agentcube/models/pack_models.py`, but there was no `agentcube/models/__init__.py`, and `cmd/cli/pyproject.toml` listed packages manually without `agentcube.models`.

## What I changed

- I added `cmd/cli/agentcube/models/__init__.py`.
- I switched the CLI pyproject to `setuptools.packages.find`, matching the Python SDK style.
- I added a focused packaging discovery test for `agentcube.models`.
- I added PR and publish workflow smoke tests that install the built CLI wheel and run `kubectl-agentcube --help`.

## Why this is legit

This keeps the package layout aligned with the existing `agentcube*` Python package structure instead of maintaining a fragile hand-written package list. The added CI checks the actual installed-wheel path that broke for users, so source-tree imports will not hide this class of issue again.

## Tests

- Screenshot 2026-05-12 at 3 38 10 PM

- Screenshot 2026-05-12 at 3 38 31 PM

- Screenshot 2026-05-12 at 3 39 06 PM

-  Screenshot 2026-05-12 at 3 39 30 PM

- `.venv/bin/python -m ruff check . --config pyproject.toml` - Passed.
- `git diff --check` - Passed.

## Notes

I did not run Go or Kubernetes codegen because this change only touches the Python CLI packaging and CI workflow.

### DIFF
--- a/.github/workflows/python-cli-tests.yml
+++ b/.github/workflows/python-cli-tests.yml
@@ -1,0 +1,53 @@
+name: Python CLI Tests
+
+on:
+  pull_request:
+  merge_group:
+
+jobs:
+  python-cli-tests:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Filter paths
+        id: changes
+        uses: dorny/paths-filter@v3
+        with:
+          filters: |
+            cli:
+              - "cmd/cli/**"
+              - "Makefile"
+              - ".github/workflows/python-cli-publish.yml"
+              - ".github/workflows/python-cli-tests.yml"
+
+      - name: Setup Python
+        if: steps.changes.outputs.cli == 'true'
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install dependencies
+        if: steps.changes.outputs.cli == 'true'
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install build pytest setuptools
+          python -m pip install -e ./cmd/cli
+
+      - name: Run unit tests
+        if: steps.changes.outputs.cli == 'true'
+        run: python -m pytest cmd/cli/tests -q
+
+      - name: Build CLI package
+        if: steps.changes.outputs.cli == 'true'
+        run: make build-python-cli
+
+      - name: Verify wheel install
+        if: steps.changes.outputs.cli == 'true'
+        run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/python-cli-tests.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Filter paths
+        id: changes
+        uses: dorny/paths-filter@v3
+        with:
+          filters: |
+            cli:
+              - "cmd/cli/**"
+              - "Makefile"
+              - ".github/workflows/python-cli-tests.yml"
```
acsoto, line 24: Can we include `.github/workflows/python-cli-publish.yml` in this filter too? Changes to the release workflow are otherwise allowed to skip the CLI build/install smoke path, so a publish-only workflow edit would not be exercised until tag time.

mrinalchaturvedi27: Thank you for the review @acsoto. I have made the described fix. Please let me know if you want any other specific change :)

hzxuzhonghu (commented):
Good fix for the missing `models` package.

`cmd/cli/pyproject.toml`
- Lines 64-67: Switching from an explicit package list to `find_packages` with `include = ["agentcube*"]` is the right call - it prevents future packages from being silently excluded from the wheel.

`cmd/cli/agentcube/models/__init__.py`
- Needed for package discovery. Good.

`.github/workflows/python-cli-publish.yml`
- Lines 34-38: The smoke-test step (install wheel into a clean venv and run `--help`) is a useful gate. Note that it only verifies the CLI entrypoint loads, not that all subpackages are importable. Consider adding an `import agentcube.models` check to catch the specific regression this PR fixes.

`cmd/cli/tests/test_packaging.py`
- Line 21-32: Good regression test using `find_packages` to verify all expected packages are discovered.

LGTM.

PR #327, opened 2026-05-12
fix(cli): add publish endpoint option

mrinalchaturvedi27:
**What type of PR is this?**

/kind bug

This makes the documented `kubectl agentcube publish --endpoint` option work. The CLI passes it to `PublishRuntime` as `agent_endpoint`, and the AgentCube publish path saves it in metadata so later commands like `invoke` can read it.

Fixes #

I found this while working on PR #325. This adds a CLI test for the option pass-through and a runtime test to check that the endpoint is saved in metadata.

## What I fixed

I fixed the Python CLI publish command so `kubectl agentcube publish --endpoint` works as documented. The endpoint is now passed to the runtime and saved in metadata.

## How I found it

I found this while working on PR #325. I checked the README section for `publish`, then looked at `cmd/cli/agentcube/cli/main.py` and `cmd/cli/agentcube/runtime/publish_runtime.py`. The runtime already checked `options.get('agent_endpoint')`, but the CLI did not expose the option. A review also pointed out that the AgentCube publish path returned the endpoint but did not save it in metadata.

## What I changed

- Added `--endpoint` to the `publish` command options.
- Passed it to `PublishRuntime.publish` as `agent_endpoint`.
- Saved `agent_endpoint` in metadata in the AgentCube publish path.
- Added a CLI test for `publish --endpoint`.
- Added a runtime test for saving the endpoint in metadata.

## Why this is legit

This matches the README and keeps `publish` and `invoke` behavior consistent. It is a small Python CLI/runtime change and does not touch Kubernetes resources, Go code, generated files, or chart templates.

## Tests

- Screenshot 2026-05-12 at 6 09 01 PM

- Screenshot 2026-05-12 at 6 09 24 PM

- Screenshot 2026-05-12 at 6 09 51 PM

- Screenshot 2026-05-12 at 6 10 23 PM

### DIFF
--- a/cmd/cli/agentcube/cli/main.py
+++ b/cmd/cli/agentcube/cli/main.py
@@ -303,6 +303,11 @@ def publish(
         "--replicas",
         help="Number of replicas for K8s deployment (default: 1)",
     ),
+    endpoint: Optional[str] = typer.Option(
+        None,
+        "--endpoint",
+        help="Custom API endpoint for AgentCube or Kubernetes cluster",
+    ),
     namespace: Optional[str] = typer.Option(
         None,
         "--namespace",
@@ -339,9 +344,10 @@ def publish(
                 "description": description,
                 "region": region,
                 "cloud_provider": cloud_provider,
-                "provider": provider, # Pass provider down
+                "provider": provider,
                 "node_port": node_port,
                 "replicas": replicas,
+                "agent_endpoint": endpoint,
                 "namespace": namespace,
             }
 
--- a/cmd/cli/agentcube/runtime/publish_runtime.py
+++ b/cmd/cli/agentcube/runtime/publish_runtime.py
@@ -171,19 +171,19 @@ class PublishRuntime:
         except Exception as e:
             raise RuntimeError(f"Failed to deploy AgentRuntime CR to K8s: {str(e)}")
 
+        endpoint = options.get('agent_endpoint') or metadata.agent_endpoint
+        if not endpoint:
+            raise ValueError("Please enter the endpoint for the agent")
+
         # Step 4: Update metadata with K8s deployment information
         updates = {
             "agent_id": k8s_info["deployment_name"],
+            "agent_endpoint": endpoint,
             "k8s_deployment": {
                 **k8s_info,
                 "type": "AgentRuntime",
             }
         }
-
-        # Use provided endpoint or fall back to router_url from metadata
-        endpoint = options.get('agent_endpoint') or metadata.agent_endpoint
-        if not endpoint:
-            raise ValueError("Please enter the endpoint for the agent")
 
         self.metadata_service.update_metadata(workspace_path, updates)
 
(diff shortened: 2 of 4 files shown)


cmd/cli/agentcube/cli/main.py
```diff
@@ -339,9 +344,10 @@ def publish(
                 "description": description,
                 "region": region,
                 "cloud_provider": cloud_provider,
-                "provider": provider, # Pass provider down
+                "provider": provider,
                 "node_port": node_port,
                 "replicas": replicas,
+                "agent_endpoint": endpoint,
```
acsoto, line 350: Can we also persist this value in the AgentCube publish path? `PublishRuntime._publish_cr_to_k8s` returns `agent_endpoint`, but its metadata update only writes `agent_id` and `k8s_deployment`; a later `agentcube invoke` still reads `metadata.agent_endpoint`, so publishing with `--endpoint` can succeed and then invoke fails unless the metadata already had an endpoint.

mrinalchaturvedi27: Thank you so much for the review and catching this @acsoto . 
I updated the AgentCube publish path to persist the endpoint passed via `--endpoint` into metadata as `agent_endpoint`. I also added a runtime test to cover this path, so a later `invoke` can read the saved endpoint from metadata.

mrinalchaturvedi27:
Hi @acsoto @YaoZengzeng quick ping for this PR :)

cmd/cli/agentcube/cli/main.py
```diff
@@ -303,6 +303,11 @@ def publish(
         "--replicas",
         help="Number of replicas for K8s deployment (default: 1)",
     ),
+    endpoint: Optional[str] = typer.Option(
+        None,
+        "--endpoint",
+        help="Custom API endpoint for AgentCube or Kubernetes cluster",
```
acsoto, line 309: Nit: this reads like a Kubernetes API server endpoint, but the value is stored as `agent_endpoint` and later used as the base URL for invoking the published agent. Can we make the help text say that more directly?

cmd/cli/agentcube/runtime/publish_runtime.py
```diff
@@ -171,20 +171,20 @@ def _publish_cr_to_k8s(
         except Exception as e:
             raise RuntimeError(f"Failed to deploy AgentRuntime CR to K8s: {str(e)}")
 
+        endpoint = options.get('agent_endpoint') or metadata.agent_endpoint
```
acsoto, line 174: Can we resolve and validate this before calling `deploy_agent_runtime`? If neither `--endpoint` nor `metadata.agent_endpoint` is set, publish currently creates the AgentRuntime CR and then fails before updating metadata, leaving the cluster state ahead of the workspace state.

hzxuzhonghu (commented):
`cmd/cli/agentcube/cli/main.py`
- Lines 306-310: The `--endpoint` option is clean. It is passed as `agent_endpoint` in the options dict.

`cmd/cli/agentcube/runtime/publish_runtime.py`
- Lines 174-176: Moving the endpoint resolution and validation before the metadata update (Step 4) is a good improvement. In the original code, the endpoint check happened after `update_metadata`, meaning a missing endpoint would leave stale partial metadata. Now it fails early and cleanly.

`cmd/cli/tests/test_cli_publish.py`
- Good end-to-end test using a stub `PublishRuntime` to verify the `--endpoint` flag is forwarded correctly.

LGTM.

PR #328, opened 2026-05-13
test: add mkdirSafe symlink escape test for PicoD

HarshitPal25:
**What type of PR is this?**

/kind security

Adds `TestMkdirSafe` to verify the security-critical `mkdirSafe` function in the PicoD file operations package. This function guards against symlink-based directory traversal attacks within the sandbox workspace.

The test adds two scenarios:
1. Normal `mkdir` within the workspace succeeds.
2. Symlink pointing outside the workspace is blocked properly.

Fixes #

`mkdirSafe` is a security boundary function — if it fails to detect symlink escapes, an attacker could create directories outside the sandbox workspace, potentially compromising host filesystem integrity.

```release-note
NONE

### DIFF
--- a/pkg/picod/files_test.go
+++ b/pkg/picod/files_test.go
@@ -275,3 +275,45 @@ func TestSetWorkspace_WithTemporaryDirectory(t *testing.T) {
 
 	assert.Equal(t, tmpDir, server.workspaceDir)
 }
+
+func TestMkdirSafe(t *testing.T) {
+	// Create a temporary workspace directory
+	tmpDir, err := os.MkdirTemp("", "picod-mkdirsafe-*")
+	assert.NoError(t, err)
+	defer os.RemoveAll(tmpDir)
+
+	server := &Server{workspaceDir: tmpDir}
+
+	t.Run("normal mkdir within workspace succeeds", func(t *testing.T) {
+		dir := filepath.Join(tmpDir, "subdir", "nested")
+		err := server.mkdirSafe(dir)
+		assert.NoError(t, err)
+
+		info, statErr := os.Stat(dir)
+		assert.NoError(t, statErr)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by hzxuzhonghu — APPROVED** [review]
/lgtm

File: pkg/picod/files_test.go [hunk 1]
```diff
@@ -275,3 +275,45 @@ func TestSetWorkspace_WithTemporaryDirectory(t *testing.T) {
...
+		dir := filepath.Join(tmpDir, "subdir", "nested")
+		err := server.mkdirSafe(dir)
+		assert.NoError(t, err)
+
+		info, statErr := os.Stat(dir)
+		assert.NoError(t, statErr)
+		assert.True(t, info.IsDir())
+	})
+
+	t.Run("symlink pointing outside workspace is blocked", func(t *testing.T) {
```
hzxuzhonghu, line 297: Minor: `assert.NoError` on line 275 and 297 but using `assert` (not `require`) means the test continues even if setup fails. For the symlink creation step specifically, if `os.Symlink` fails the rest of the test will give a confusing error. `require.NoError` on the setup lines would give cleaner failure messages.

hzxuzhonghu (commented):
Good security test — the symlink escape scenario is exactly the kind of thing that needs explicit coverage. The test is clear and the assertion that nothing was created outside the workspace is solid.

One thing that might be worth adding: a test for a symlink chain (symlink pointing to another symlink that escapes). Depends how `mkdirSafe` resolves paths internally — if it uses `filepath.EvalSymlinks` it likely handles it already, but worth having the test confirm it.

HarshitPal25:
@hzxuzhonghu thanks for the compliment,
And for the test case for symlink chain do I just make changes here or create a new pr which includes the test case

PR #334, opened 2026-05-13
fix: evict expired entries in TokenCache.Get() to prevent memory leak

HarshitPal25:
## What type of PR is this?

/kind bug

## What this PR does / why we need it:

This PR fixes a memory leak in `TokenCache.Get()` where expired entries are never evicted from the cache.

Previously, when `TokenCache.Get()` encountered an expired entry, it returned `(false, false, "")` but left the stale entry in the cache map and LRU list. Under high-cardinality token usage (e.g., short-lived K8s service account tokens rotated frequently), dead entries accumulate until LRU eviction pressure pushes them out, starving valid tokens from being cached.

`ClientCache.Get()` in the same file (`client_cache.go:110-114`) correctly evicts expired entries inline, proving this was an oversight in `TokenCache`.

The root cause is that `TokenCache.Get()` used `RLock()` (read lock), which prevented it from mutating the cache to remove stale entries. `ClientCache.Get()` uses a full `Lock()` and removes expired entries correctly.

This PR:
- Upgrades `TokenCache.Get()` from `RLock` → `Lock` so it can mutate state on expiry.
- Evicts expired entries inline (removes from both the LRU list and the map), matching the pattern established by `ClientCache.Get()`.
- Promotes accessed entries in the LRU list on `Get()` for consistent eviction ordering.
- Adds a missing assertion in `TestTokenCache_Get_Expired` verifying that expired entries are actually removed (`cache.Size() == 0`), not just hidden by the return value.
- Updates `TestTokenCache_LRUBehavior` to reflect that `Get()` now promotes entries, so eviction order changes correctly.

## Which issue(s) this PR fixes:

None (discovered during code review )

## Special notes for your reviewer:

The lock upgrade from `RLock` → `Lock` in `Get()` trades a small amount of read concurrency for correctness. Under production workloads, `Get()` calls are short-lived (map lookup + time check), so the contention impact is negligible. This matches how `ClientCache.Get()` already operates with a full `Lock()`.

All existing tests pass with `go test -race`.

## Does this PR introduce a user-facing change?:

```release-note
NONE
```

### DIFF
--- a/pkg/workloadmanager/client_cache.go
+++ b/pkg/workloadmanager/client_cache.go
@@ -230,19 +230,24 @@ func NewTokenCache(maxSize int, ttl time.Duration) *TokenCache {
 // Returns found status, authenticated status, and username
 // If found is false, the token was not in cache or expired
 func (c *TokenCache) Get(token string) (found bool, authenticated bool, username string) {
-	c.mu.RLock()
-	defer c.mu.RUnlock()
+	c.mu.Lock()
+	defer c.mu.Unlock()
 
 	entry, exists := c.cache[token]
 	if !exists {
 		return false, false, ""
 	}
 
-	// Check if entry is expired
+	// Check if entry is expired; evict stale entries to prevent memory leak.
+	// This mirrors ClientCache.Get which correctly removes expired entries.
 	if time.Since(entry.lastAccess) > c.ttl {
+		c.lruList.Remove(entry.element)
+		delete(c.cache, token)
 		return false, false, ""
 	}
 
+	// Move to front on access for proper LRU ordering
+	c.lruList.MoveToFront(entry.element)
 	return true, entry.authenticated, entry.username
 }
 
--- a/pkg/workloadmanager/client_cache.go
+++ b/pkg/workloadmanager/client_cache.go
@@ -246,7 +246,8 @@ func (c *TokenCache) Get(token string) (found bool, authenticated bool, username
 		return false, false, ""
 	}
 
-	// Move to front on access for proper LRU ordering
+	// Move to front on access for proper LRU ordering and sliding TTL.
+	entry.lastAccess = time.Now()
 	c.lruList.MoveToFront(entry.element)
 	return true, entry.authenticated, entry.username
 }
--- a/pkg/workloadmanager/client_cache_test.go
+++ b/pkg/workloadmanager/client_cache_test.go
@@ -432,6 +432,8 @@ func TestTokenCache_Get_Expired(t *testing.T) {
 	found, authenticated, _ = cache.Get(token)
 	assert.False(t, found)
 	assert.False(t, authenticated)
+	// Expired entry should be evicted from cache, not just hidden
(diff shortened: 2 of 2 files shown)


HarshitPal25:
@hzxuzhonghu hello sir please checkout my work

pkg/workloadmanager/client_cache.go
```diff
@@ -230,19 +230,24 @@ func NewTokenCache(maxSize int, ttl time.Duration) *TokenCache {
...
+	// Check if entry is expired; evict stale entries to prevent memory leak.
+	// This mirrors ClientCache.Get which correctly removes expired entries.
 	if time.Since(entry.lastAccess) > c.ttl {
+		c.lruList.Remove(entry.element)
+		delete(c.cache, token)
 		return false, false, ""
 	}
 
+	// Move to front on access for proper LRU ordering
+	c.lruList.MoveToFront(entry.element)
```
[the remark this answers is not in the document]
hzxuzhonghu: yeah, you should also update the timestamp

pkg/workloadmanager/client_cache.go
```diff
@@ -230,19 +230,24 @@ func NewTokenCache(maxSize int, ttl time.Duration) *TokenCache {
 // Returns found status, authenticated status, and username
 // If found is false, the token was not in cache or expired
 func (c *TokenCache) Get(token string) (found bool, authenticated bool, username string) {
-	c.mu.RLock()
-	defer c.mu.RUnlock()
+	c.mu.Lock()
```
hzxuzhonghu, line 233: Upgrading from `RLock` to `Lock` is correct for correctness but worth calling out: every call to `Get()` on a hot path now takes the exclusive lock. Under high concurrency with lots of token checks this could become a bottleneck. A two-phase approach (check under `RLock`, upgrade to `Lock` only when eviction is needed) would keep reads fast, though it adds complexity. For now this is fine — just something to keep in mind if we profile and find contention here.

pkg/workloadmanager/client_cache.go
hzxuzhonghu, line 251: Nice. Without `MoveToFront` here, Get() was basically treating the list as FIFO — a recently-validated token could get evicted before a stale one. The test update confirms the corrected eviction order.

hzxuzhonghu (commented):
Solid fix — the stale-entry accumulation was a real leak. Two quick notes below.

HarshitPal25:
@hzxuzhonghu ohh I would resolve it by tomorrow

hzxuzhonghu:
maybe replace persoanl iilru implement with hashicorp golang-lru pkg

HarshitPal25:
Thanks for the suggestion. Do you recommend making the `hashicorp/golang-lru` replacement in this PR, or would you prefer I keep this PR focused on the timestamp/expired-entry fix and open a separate follow-up PR for the LRU package migration?

hzxuzhonghu:
Can do the replace in a separate pr
/lgtm
/approve

Issue #267, opened 2026-04-11
[Proposal] Position AgentCube as a Stateful, Isolated, Concurrent Rollout Execution Layer for Agentic RL and Verifiable Agentic Tasks


acsoto:
### Background

Many emerging agent workloads are no longer single-turn prompt/response tasks. They are **multi-step, stateful, sandboxed executions** that require:

- isolated execution environments,
- persistent per-run state and workspace,
- concurrent fan-out of multiple rollouts,
- and automatic evaluation through tests, scripts, or environment checks.

This is especially relevant for:

- verifiable agentic tasks such as SWE-bench-style software engineering tasks,
- best-of-N agent evaluation,
- trajectory collection,
- and full RL / post-training pipelines where rollout execution is a core stage.

AgentCube already appears well aligned with this direction through its sandbox-oriented, stateful execution model. This suggests an important positioning:

> **AgentCube can serve as a stateful, isolated, concurrent rollout execution layer for agentic RL and other verifiable agentic tasks.**

### Proposal

The proposal is to make this role explicit.

AgentCube should be positioned as the **execution substrate** responsible for:

- launching and managing one sandbox per rollout,
- preserving per-rollout workspace and session state,
- supporting concurrent rollout fan-out,
- collecting rollout traces, artifacts, and execution outcomes,
- and enabling automated evaluators to score results.

In this model, AgentCube is **not the RL trainer itself**, but it plays a critical role in the RL loop by serving as the rollout layer between:

- task input,
- agent execution,
- environment interaction,
- and evaluator / reward generation.

This also naturally supports non-training scenarios such as best-of-N evaluation and verifiable agent benchmarking.

### Example Workflow

A minimal example is a SWE-bench-style software engineering task:

1. A task includes a repository snapshot, issue description, evaluation script, and execution limits.
2. The same task is expanded into **N rollouts**.
3. AgentCube launches **one isolated sandbox per rollout**.
4. The agent performs multi-step execution inside each sandbox.
5. Intermediate files, logs, and workspace state remain local to that rollout.
6. An evaluator script runs automatically and returns pass/fail, score, and execution metrics.
7. The system compares rollout outcomes such as success rate, best-of-N result, runtime, and artifact quality.

This already demonstrates clear value even before full policy optimization is added.

### Why This Matters

This positioning makes AgentCube useful for two closely related layers of the stack:

**1. Verifiable agent execution**

- sandboxed multi-step tasks
- concurrent rollout evaluation
[...]

PR #337, opened 2026-05-14
Harden PicoD server: graceful shutdown, symlink guard, doc fix

Abhinav-kodes:
**What type of PR is this?**

/kind security
/kind enhancement

This PR hardens the PicoD server with three targeted improvements:

1. **Graceful shutdown** — `Run()` now accepts a `context.Context` and shuts down the HTTP server cleanly on `SIGINT`/`SIGTERM`, allowing in-flight requests to complete before exit. This is foundational for the stateful execution direction (#267).

2. **Symlink traversal guard in upload handlers** — `handleMultipartUpload` and `handleJSONBase64Upload` used `os.MkdirAll` to create parent directories, which follows symlinks in existing path components. An attacker with workspace write access could place a symlink pointing outside the jail, causing directory creation beyond the workspace boundary. Both call sites now use the existing `mkdirSafe` guard, consistent with `ExecuteHandler`.

3. **Doc fix** — The `ExecuteRequest.Timeout` comment stated the default was `"30s"` but the actual default is `60s`.

Fixes #

- The `setWorkspace` call site intentionally keeps `os.MkdirAll` — it creates the workspace root itself at startup from admin config, and `mkdirSafe` uses `workspaceDir` as the jail root, so using it there would be circular.
- The graceful shutdown goroutine in `Run()` mirrors the pattern used by the workloadmanager's `Shutdown()` method.

```release-note
NONE

### DIFF
--- a/pkg/picod/files.go
+++ b/pkg/picod/files.go
@@ -94,7 +94,7 @@ func (s *Server) handleMultipartUpload(c *gin.Context) {
 
 	// Create directory
 	dir := filepath.Dir(safePath)
-	if err := os.MkdirAll(dir, 0755); err != nil {
+	if err := s.mkdirSafe(dir); err != nil {
 		c.JSON(http.StatusInternalServerError, gin.H{
 			"error": fmt.Sprintf("Failed to create directory: %v", err),
 			"code":  http.StatusInternalServerError,
@@ -186,7 +186,7 @@ func (s *Server) handleJSONBase64Upload(c *gin.Context) {
 
 	// Create directory
 	dir := filepath.Dir(safePath)
-	if err := os.MkdirAll(dir, 0755); err != nil {
+	if err := s.mkdirSafe(dir); err != nil {
 		c.JSON(http.StatusInternalServerError, gin.H{
 			"error": fmt.Sprintf("Failed to create directory: %v", err),
 			"code":  http.StatusInternalServerError,
--- a/pkg/picod/files.go
+++ b/pkg/picod/files.go
@@ -18,6 +18,7 @@ package picod
 
 import (
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"io"
 	"mime"
@@ -31,6 +32,10 @@ import (
 	"github.com/gin-gonic/gin"
 	"k8s.io/klog/v2"
 )
+
+// ErrAccessDenied is returned by mkdirSafe when a path escapes the workspace
+// jail via a symlink. Callers should use errors.Is to detect it.
+var ErrAccessDenied = errors.New("access denied: path escapes workspace jail via symlink")
 
 const (
 	maxFileMode = 0777 // Maximum allowed file permission mode
@@ -96,7 +101,7 @@ func (s *Server) handleMultipartUpload(c *gin.Context) {
 	dir := filepath.Dir(safePath)
 	if err := s.mkdirSafe(dir); err != nil {
 		code := http.StatusInternalServerError
-		if strings.Contains(err.Error(), "access denied") {
+		if errors.Is(err, ErrAccessDenied) {
 			code = http.StatusForbidden
 		}
 		c.JSON(code, gin.H{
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: pkg/picod/server.go [hunk 1]
```diff
@@ -106,7 +107,24 @@ func (s *Server) Run() error {
 		ReadHeaderTimeout: 10 * time.Second, // Prevent Slowloris attacks
 	}
 
-	return server.ListenAndServe()
+	idleConnsClosed := make(chan struct{})
+	go func() {
+		<-ctx.Done()
+		// Allow enough time for in-flight commands to finish (default timeout is 60s).
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), 90*time.Second)
```
acsoto: Can we make this shutdown timeout configurable, or cap `ExecuteRequest.Timeout` so it cannot exceed the shutdown grace period? The current `90s` still allows a request with `timeout: "10m"` to be cut off during graceful shutdown, so the new guarantee only holds for default-length commands.

pkg/picod/files.go
```diff
@@ -94,7 +94,7 @@ func (s *Server) handleMultipartUpload(c *gin.Context) {
 
 	// Create directory
 	dir := filepath.Dir(safePath)
-	if err := os.MkdirAll(dir, 0755); err != nil {
+	if err := s.mkdirSafe(dir); err != nil {
```
acsoto, line 102: Can we add regression coverage for both JSON and multipart uploads where an existing workspace path component is a symlink outside the workspace? This is the core security fix, and endpoint-level tests would catch either handler drifting back to `os.MkdirAll`.

pkg/picod/server.go
Abhinav-kodes: Done. I've introduced the `-shutdown-timeout` flag, capped the execution timeout accordingly, and added `TestExecuteHandler_TimeoutCapping`.

pkg/picod/files.go
Abhinav-kodes: I've added `TestPicoD_SymlinkUploadGuard` which tests both JSON and multipart upload endpoints with symlinks pointing outside the workspace jail.

hzxuzhonghu (commented):
Solid hardening PR covering graceful shutdown, timeout capping, and symlink guards.

`cmd/picod/main.go`
- Lines 34-44: Signal handling with `signal.NotifyContext` is idiomatic Go. Good.
- Line 34: `shutdownTimeout` flag defaults to 90s which gives enough headroom for in-flight commands.

`pkg/picod/server.go`
- Lines 48-50: Default shutdown timeout in `NewServer` is good defensive coding.
- Lines 114-134: Graceful shutdown implementation is correct. The `idleConnsClosed` channel ensures the main goroutine waits for shutdown to complete. The `server.ListenAndServe` error check correctly filters `http.ErrServerClosed`.
- Lines 121-123: The redundant `shutdownTimeout <= 0` check (already defaulted in `NewServer`) is harmless but unnecessary.

`pkg/picod/execute.go`
- Lines 159-180 (`parseAndCapTimeout`): Capping the command timeout to the shutdown grace period prevents orphaned commands that would outlive the server. This is an important safety measure for sandbox environments.

`pkg/picod/files.go`
- Lines 97, 189: Replacing `os.MkdirAll` with `s.mkdirSafe` prevents symlink-based directory traversal during file uploads. Critical for sandbox security.

`pkg/picod/picod_test.go`
- Lines 422-498 (`TestPicoD_SymlinkUploadGuard`): Both JSON and multipart upload paths are tested. Verifying that the file was NOT created outside the workspace is the right assertion.
- Lines 67-78: `setupTestServer` now returns `(*httptest.Server, string)` dropping the unused `*Server` return. Clean.

LGTM.

0xA8hinav:
cc @hzxuzhonghu

acsoto:
e2e failed

0xA8hinav:
@acsoto PTAL

PR #339, opened 2026-05-14
feat(picod): add gzip middleware for response compression

Abhinav-kodes:
**What type of PR is this?**

/kind enhancement
/kind feature

This PR implements Gzip compression for the PicoD daemon, fulfilling one of the planned network enhancements in the [`picod-proposal.md`](../docs/design/picod-proposal.md) design doc (Future Enhancements, item-2).

By injecting the `gin-contrib/gzip` middleware, the PicoD server performs on-the-fly compression for HTTP responses. This reduces network bandwidth and improves latency for large data transfers, such as retrieving bulky `stdout`/`stderr` payloads from `/api/execute` or downloading large text/JSON files via `/api/files/*path`.

The middleware uses standard HTTP content negotiation:
- Clients advertising `Accept-Encoding: gzip` receive compressed payloads.
- Clients that do not support compression seamlessly receive plain responses (no change in behavior).

Fixes #

- Pinned `gin-contrib/gzip` to `v1.0.1` to avoid a transitive upgrade to Go 1.25.0+, keeping compatibility with the project's `go1.24.4` toolchain directive.
- Uses `gzip.BestSpeed` (level 1) rather than the default level 6, prioritising minimal CPU overhead over maximum compression ratio — appropriate for a latency-sensitive daemon serving constrained sandbox networks.
- `/health` is excluded from compression via `gzip.WithExcludedPaths` to avoid unnecessary processing on a lightweight liveness probe.
- Two focused tests have been added to `server_test.go`:
  - `TestServer_GzipMiddleware_CompressesResponse` — verifies `Content-Encoding: gzip` is set when the client sends `Accept-Encoding: gzip`.
  - `TestServer_GzipMiddleware_ExcludesHealthEndpoint` — verifies `/health` is **not** compressed even when the client requests it.

```release-note
Added Gzip compression support to the PicoD daemon to reduce bandwidth usage for file transfers and command execution payloads over constrained networks.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ toolchain go1.24.9
 require (
 	github.com/agiledragon/gomonkey/v2 v2.13.0
 	github.com/alicebob/miniredis/v2 v2.35.0
+	github.com/gin-contrib/gzip v1.0.1
 	github.com/gin-gonic/gin v1.10.0
 	github.com/golang-jwt/jwt/v5 v5.2.2
 	github.com/google/uuid v1.6.0
--- a/go.sum
+++ b/go.sum
@@ -37,6 +37,8 @@ github.com/fxamacker/cbor/v2 v2.9.0 h1:NpKPmjDBgUfBms6tr6JZkTHtfFGcMKsw3eGcmD/sa
 github.com/fxamacker/cbor/v2 v2.9.0/go.mod h1:vM4b+DJCtHn+zz7h3FFp/hDAI9WNWCsZj23V5ytsSxQ=
 github.com/gabriel-vasile/mimetype v1.4.3 h1:in2uUcidCuFcDKtdcBxlR0rJ1+fsokWf+uqxgUFjbI0=
 github.com/gabriel-vasile/mimetype v1.4.3/go.mod h1:d8uq/6HKRL6CGdk+aubisF/M5GcPfT7nKyLpA0lbSSk=
+github.com/gin-contrib/gzip v1.0.1 h1:HQ8ENHODeLY7a4g1Au/46Z92bdGFl74OhxcZble9WJE=
+github.com/gin-contrib/gzip v1.0.1/go.mod h1:njt428fdUNRvjuJf16tZMYZ2Yl+WQB53X5wmhDwXvC4=
 github.com/gin-contrib/sse v0.1.0 h1:Y/yl/+YNO8GZSjAhjMsSuLt29uWRFHdHYUb5lYOV9qE=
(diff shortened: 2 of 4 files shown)


acsoto:
Thanks. Conflicts exists and I guess this package introduced is not necessary?

0xA8hinav:
Thanks for the review, @acsoto!

### Conflicts

Resolved — I've rebased the branch onto the latest `upstream/main`, which also preserves the `MaxBodySize` DoS-protection middleware added in #326.

### On necessity

Gzip compression is explicitly listed as a planned enhancement in the design document:

> `picod-proposal.md` → **Future Enhancements**, item-2:
> “Compression — Gzip or similar compression for efficient file transfers.”

The motivation is concrete: PicoD's primary use case involves transferring ML model artifacts, training datasets, and bulky stdout/stderr payloads between the SDK and sandboxes over constrained in-cluster networks.

For these workloads, gzip at `BestSpeed` (level 1) typically achieves 5–10× size reduction on JSON/text payloads with negligible CPU cost (~1–2ms per response).

The dependency (`gin-contrib/gzip`) is an official, actively maintained Gin middleware with no transitive upgrades required. It's completely transparent to existing clients — those without `Accept-Encoding: gzip` automatically fall back to uncompressed responses.

Happy to drop it if the project would prefer to defer this, but given it's already in the design doc and the footprint is minimal, I think this is a good time to land it.

hzxuzhonghu (commented):
`pkg/picod/server.go`
- Line 98: `gzip.BestSpeed` is the right choice for a latency-sensitive runtime proxy. Excluding `/health` avoids burning CPU on frequent probe requests. Note that this middleware is order-dependent with the body-size limiter from PR #326 - if both merge, verify that gzip runs after the body-size check so Content-Length validation is not defeated by a compressed request body (it should be fine since gzip middleware only compresses responses, not requests).

`pkg/picod/server_test.go`
- Lines 374-450: Both test cases (compression on /api/execute, exclusion on /health) are well structured. Using `DisableCompression: true` on the transport to inspect raw headers is the correct approach.

LGTM.

0xA8hinav:
Thanks for the review, @hzxuzhonghu
On the middleware ordering, the body-size guard (lines 85–96) is registered before the gzip middleware (line 98), and since gin-contrib/gzip only compresses responses (never reads/modifies the request body), the Content-Length pre-check and MaxBytesReader both execute independently of gzip. No bypass risk.

The PR is ready to be merged from my side without conflicts

PR #343, opened 2026-05-17
fix(agentd): use custom idle timeout annotation for garbage collection

### LINKED ISSUE #342: agentd: custom idle timeout annotation is ignored by garbage collector
/kind bug

**What happened**:
When configuring a custom `sessionTimeout` or `maxSessionDuration` on a sandbox runtime (e.g. an `AgentRuntime` or `CodeInterpreter`), the `workloadmanager` successfully generates the Sandbox and sets the `runtime.agentcube.io/idle-timeout` annotation to the requested custom duration (e.g., `"5m"` or `"30m"`). 

However, the `agentd` garbage collection daemon completely ignores this custom annotation. In `pkg/agentd/agentd.go` inside the `Reconcile` loop, the Reconciler calculates sandbox expiration using the hardcoded `SessionExpirationTimeout` constant which is strictly set to `15 * time.Minute`:
```go
expirationTime := lastActivity.Add(SessionExpirationTimeout)
```
Consequently, all custom idle timeouts are completely bypassed, and every idle sandbox is garbage-collected exactly at 15 minutes.

The `agentd` garbage collection daemon should check the `runtime.agentcube.io/idle-timeout` annotation on the Sandbox metadata. If the annotation exists and is a valid duration, it should parse and use that duration to compute the expiration time. It should fall back to the default `SessionExpirationTimeout` (15 minutes) only if the annotation is missing or invalid.

1. Create a `CodeInterpreter` or `AgentRuntime` resource with a custom session timeout (e.g., `5m`).
2. Invoke the sandbox to trigger creation. Note that the sandbox is created with the `runtime.agentcube.io/idle-timeout: "5m"` annotation.
3. Stop sending any requests to the sandbox so that it becomes idle.
4. Observe that `agentd` does not clean up the sandbox at 5 minutes, but instead cleans it up at exactly 15 minutes.

This is a functional bug that prevents proper session lifecycle management for environments that need shorter timeouts (to save resources) or longer timeouts (for interactive user tasks).

**Environment**:

- agentcube version: v0.1.0 (main branch)
- Kubernetes version: Any supported K8s cluster (e.g. kind/v1.31)

hrmn3s:
**What type of PR is this?**

/kind bug

This PR fixes a bug where `agentd` ignores the custom `runtime.agentcube.io/idle-timeout` annotation set on a Sandbox. Previously, the garbage collection reconciliation loop used a hardcoded `15 * time.Minute` value for every sandbox. 

With this change, `agentd` successfully parses the Sandbox's custom idle timeout from the metadata annotations and correctly honors shorter or longer custom session durations. It will fall back to the default `15 * time.Minute` if the annotation is missing or improperly formatted.

It also introduces new unit test coverage inside `agentd_test.go` to explicitly verify custom short timeouts, custom long timeouts, and invalid timeout fallback paths.

Fixes #342

* All changes covered by unit tests.

```release-note
Fixes an issue where custom Sandbox idle timeouts (`runtime.agentcube.io/idle-timeout`) were ignored by the garbage collector and always fell back to 15 minutes.
```

### DIFF
--- a/pkg/agentd/agentd.go
+++ b/pkg/agentd/agentd.go
@@ -29,10 +29,6 @@ import (
 	"github.com/volcano-sh/agentcube/pkg/workloadmanager"
 )
 
-var (
-	SessionExpirationTimeout = 15 * time.Minute
-)
-
 // Reconciler reconciles a Sandbox object
 type Reconciler struct {
 	client.Client
@@ -57,7 +53,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			return ctrl.Result{RequeueAfter: 30 * time.Second}, err
 		}
 
-		expirationTime := lastActivity.Add(SessionExpirationTimeout)
+		expirationTime := lastActivity.Add(workloadmanager.DefaultSandboxIdleTimeout)
+		
+		// Use custom idle timeout if defined.
+		// Ignore invalid, zero, or negative values and fall back to the default timeout.
+		if timeoutStr, exists := sandbox.Annotations[workloadmanager.IdleTimeoutAnnotationKey]; exists && timeoutStr != "" {
+			if customTimeout, err := time.ParseDuration(timeoutStr); err == nil && customTimeout > 0 {
+				expirationTime = lastActivity.Add(customTimeout)
+			}
+		}
 		// Delete sandbox if expired
 		if time.Now().After(expirationTime) {
 			if err := r.Delete(ctx, sandbox); err != nil {
--- a/pkg/agentd/agentd_test.go
+++ b/pkg/agentd/agentd_test.go
@@ -160,7 +160,7 @@ func TestReconciler_Reconcile_LifecycleOrchestration(t *testing.T) {
 					Name:      "expired-sandbox",
 					Namespace: "default",
 					Annotations: map[string]string{
-						workloadmanager.LastActivityAnnotationKey: now.Add(-SessionExpirationTimeout).Format(time.RFC3339),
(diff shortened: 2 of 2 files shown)


hrmn3s:
Hi @acsoto, @YaoZengzeng

I have updated this PR to completely streamline its scope and address the automated review feedback.
I would really appreciate it if you could take a look and share your thoughts. Thank you for your guidance and time!

pkg/agentd/agentd.go
```diff
@@ -58,6 +58,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		}
 
 		expirationTime := lastActivity.Add(SessionExpirationTimeout)
```
acsoto: Should this fallback use `workloadmanager.DefaultSandboxIdleTimeout` instead of the separate `SessionExpirationTimeout` constant? The annotation producer and agentd fallback are meant to represent the same default, so keeping two constants can drift.

hzxuzhonghu (commented):
Clean implementation with proper fallback behavior.

`pkg/agentd/agentd.go`
- Line 62-67: Good defensive coding - the `err == nil && customTimeout > 0` guard correctly rejects invalid, zero, and negative durations. Consider logging a warning when the annotation is present but unparseable, so operators can spot misconfigurations without reading the source.

`pkg/agentd/agentd_test.go`
- Test coverage is thorough: covers short timeout (deleted), long timeout (requeued), invalid string, negative, and zero values. Good edge case coverage.

LGTM.

pkg/agentd/agentd.go
hrmn3s: I've just pushed a commit that will fix this.
Thanks!

hzxuzhonghu (approved):
/lgtm

PR #136, opened 2026-01-13
Remove NeedInitialization in workload manager

acsoto:
**Fixes #134** 

This pull request primarily removes the `NeedInitialization` field from the `CodeInterpreterSpec` and related logic across the codebase. The goal is to simplify the handling of sandbox initialization for code interpreters, making initialization behavior implicit rather than configurable via a field. Additionally, there are minor dependency and deepcopy improvements.

**Removal of Code Interpreter Initialization Logic:**

* Removed the `NeedInitialization` field from the `CodeInterpreterSpec` struct and related CRD, API, and documentation, eliminating explicit configuration for initialization. [[1]](diffhunk://#diff-2866351444e50270a0a4a86083e7f067f302456c54bccc5e2def04d01e5d9403L76-L80) [[2]](diffhunk://#diff-5d0d5882cd1f229975322b147aa2d1f558670722e258cb57b93ac7dcd805ddd9L399-L403) [[3]](diffhunk://#diff-5677a0bc53b5f99c3e55551a3498ca2aff7ee0bcb1993c9e43765c1b710867c1L249-L253) [[4]](diffhunk://#diff-5677a0bc53b5f99c3e55551a3498ca2aff7ee0bcb1993c9e43765c1b710867c1R30)
* Deleted the `codeInterpreterInitialization` function and its invocation in the sandbox creation workflow, as initialization is now handled implicitly. [[1]](diffhunk://#diff-692b42265fe143f697bc303a7488dbf89b719b40d0943988d8f28d850dd79583L64-L87) [[2]](diffhunk://#diff-692b42265fe143f697bc303a7488dbf89b719b40d0943988d8f28d850dd79583L241-L248)
* Removed usage and assignment of `NeedInitialization` in external info structures and builders, further simplifying sandbox creation logic. [[1]](diffhunk://#diff-80a0f63fb0ef1a60e35489cc99d0229d2cbe52298907d2db5c15dc184dd6c0c8L79) [[2]](diffhunk://#diff-6cd9fcb8284286df1eb8a1fa514bce1a5c3ae38682d9372c347cd5a31f2d9affL285) [[3]](diffhunk://#diff-6cd9fcb8284286df1eb8a1fa514bce1a5c3ae38682d9372c347cd5a31f2d9affL323-R328)
[...]

PR #352, opened 2026-05-19
feat(picod): implement two-stage secure initialization to isolate sandbox sessions

Abhinav-kodes:
feat(picod): implement two-stage secure initialization to isolate sandbox sessions

**What type of PR is this?**

/kind feature
/kind security

---

This PR implements **Two-Stage Secure Initialization** for PicoD to resolve a critical security vulnerability regarding cross-sandbox token replays, while incorporating defense-in-depth, encrypted key storage, and performance optimizations.

Previously, under the Plain Authentication design, all PicoD sandboxes verified JWTs using the same shared public key injected via the `PICOD_AUTH_PUBLIC_KEY` environment variable. This meant a valid JWT issued for one sandbox could theoretically be replayed against another sandbox, breaking tenant isolation.

This PR introduces a cryptographically isolated two-stage handshake:

1. **Bootstrap Stage**: The Workload Manager generates a static RSA-2048 Bootstrap Key Pair on startup, stores it in the `agentcube-bootstrap-identity` Kubernetes Secret/ConfigMap (renamed from `picod-router-identity` / `picod-router-public-key`), and injects `PICOD_BOOTSTRAP_PUBLIC_KEY` along with a unique `PICOD_SESSION_ID` into each PicoD container environment. PicoD loads the bootstrap key at startup and waits in an uninitialized state.

2. **Session Stage (`POST /init`)**: The Workload Manager generates a per-sandbox **ECDSA (P-256)** Session Key Pair. It calls the new `POST /init` endpoint on PicoD with a short-lived bootstrap JWT (RS256, signed by the Bootstrap Private Key, scoped by `sub` = `PICOD_SESSION_ID`) containing the `session_public_key` claim. On success, PicoD stores the ECDSA public key and transitions to initialized. All subsequent user requests must be signed with the ECDSA Session Private Key (ES256).

---

## Files Changed

| File | Change |
|---|---|
| `cmd/picod/main.go` | Pass `context.Background()` to `NewServer` to enable lifecycle-aware goroutines |
| `cmd/workload-manager/main.go` | Reorder init to pass `server.GetBootstrapPublicKeyPEM` into `CodeInterpreterReconciler` after server construction |
| `pkg/picod/auth.go` | Refactored `AuthManager`: split into `bootstrapPublicKey` (RSA) + `sessionPublicKey` (ECDSA); added `VerifyBootstrapJWT`, `SetSessionPublicKey`, JTI replay-protection via `sync.Map`, background JTI eviction goroutine |
| `pkg/picod/server.go` | Added `POST /init` route with per-endpoint rate limiter (2 req/s, burst 5); `InitHandler` verifies bootstrap JWT and sets session key; API routes return `503 DAEMON_NOT_INITIALIZED` until session is established |
| `pkg/router/jwt.go` | Renamed secret/configmap to `agentcube-bootstrap-identity`; added LRU key cache (`keyCacheMaxSize=1000`) for per-session ECDSA private keys; added `GenerateTokenWithKey` and `GetCachedKey` helpers |
| `pkg/router/handlers.go` | `generateSandboxJWT` fetches per-session ECDSA private key from store (with `ErrKeyNotCached` fallback); returns `503 SESSION_PRIVATE_KEY_NOT_FOUND` if key is missing |
| `pkg/router/server.go` | Set `IdleConnTimeout: 85s` and `MaxIdleConnsPerHost: 100` on HTTP transport; pass `storeClient` to `TryStoreOrLoadJWTKeySecret` |
| `pkg/store/interface.go` | Added `StoreSessionPrivateKey`, `GetSessionPrivateKey`, `SetEncryptionKey` to `Store` interface |
| `pkg/store/crypto.go` *(new)* | AES-256-GCM `Crypto` helper for at-rest encryption of session private keys |
| `pkg/store/keys.go` *(new)* | Centralized `sessionPrivKeyPrefix` constant shared by Redis and Valkey backends |
| `pkg/store/store_redis.go` | Implemented `StoreSessionPrivateKey` / `GetSessionPrivateKey` (AES-GCM + base64); updated `DeleteSandboxBySessionID` to use non-transactional `Pipeline()` for cluster-safe deletion |
| `pkg/store/store_valkey.go` | Same as Redis backend; `DeleteSandboxBySessionID` uses bare `DoMulti` for cluster-safe deletion |
| `pkg/common/types/sandbox.go` | Added `SessionPrivateKey` (excluded from JSON, persisted separately) and `AuthMode` fields to `SandboxInfo` |
| `docs/design/PicoD-Plain-Authentication-Design.md` | Added migration warning; renamed ConfigMap/Secret references; updated pod spec YAML and sequence diagrams |
| `docs/design/auth-proposal.md` | Updated Router→PicoD description to reflect RSA/ECDSA dual-key scheme and new env var names |
| `docs/agentcube/blog/release-v0.1.0/index.md` | Updated JWT security chain section to describe two-stage trust chain and ES256 token signing |
| `pkg/picod/auth_test.go`, `picod_test.go`, `execute_test.go`, `server_test.go` | Full test suite updated: RSA→ECDSA test helpers, new `TestInitHandler` covering success/conflict/invalid-token cases, `TestAuthMiddleware_ValidToken` |
| `pkg/router/session_manager_test.go` | Added stub implementations for new store interface methods |
| `pkg/workloadmanager/server.go` | Added `PicoInitTimeout`, bootstrap auth manager wiring, `GetBootstrapPublicKeyPEM`, and store encryption key setup |
| `pkg/workloadmanager/workload_builder.go` | Removed the old public-key cache flow and switched sandbox builds to use `AuthMode` defaults from the new model |

---

## Key Security Hardening & Optimizations Included

- **ECDSA Migration**: Session tokens switched from RS256 (2048-bit RSA) to ES256 (ECDSA P-256), reducing CPU overhead during high-throughput sandbox initialization.
- **Defense-in-Depth (`PICOD_SESSION_ID`)**: PicoD validates the `sub` claim in the bootstrap JWT against `PICOD_SESSION_ID` when set (on-demand pods). Warm pool pods that are pre-created before a session is assigned skip this check; replay protection is still enforced via the JTI guard.
- **JTI Replay Protection**: `AuthManager` tracks seen JTIs via `sync.Map` with a background 1-minute cleanup goroutine; bootstrap JTIs older than 3 minutes are evicted.
- **One-Shot Initialization**: `SetSessionPublicKey` returns `ErrAlreadyInitialized` on second call; `InitHandler` responds `409 Conflict` to prevent session key hijacking via re-initialization.
- **At-Rest Encryption**: Session ECDSA private keys stored in Redis/Valkey are encrypted with AES-256-GCM (`pkg/store/crypto.go`) before persistence.
- **Cluster-Safe Key Deletion**: `DeleteSandboxBySessionID` uses a non-transactional `Pipeline()` (Redis) / bare `DoMulti` (Valkey) to delete sandbox metadata and the private key safely in clustered deployments.
- **Rate Limiting on `/init`**: A token-bucket rate limiter (2 req/s, burst 5) guards the `/init` endpoint against spam during the startup window.
- **HTTP Transport Hardening**: `IdleConnTimeout` set to 85s (aligned with AWS ALB 60s timeout) and `MaxIdleConnsPerHost=100` to prevent stale-connection race conditions.
- **PicoInitTimeout**: Workload Manager now uses a dedicated timeout for the `/init` handshake so sandbox initialization does not block indefinitely.
- **Auth Mode Plumbing**: `SandboxInfo` and `CreateSandboxResponse` now carry `AuthMode`, allowing the Router to distinguish `picod` sandboxes from `none`-auth sandboxes.
- **Backwards Compatibility**: `PICOD_AUTH_PUBLIC_KEY` is still accepted as a fallback; deployments should migrate to `PICOD_BOOTSTRAP_PUBLIC_KEY`.

---

## Which issue(s) this PR fixes

Fixes #

---

## Special notes for your reviewer

This implementation realizes the architecture outlined in `docs/design/agentcube-proposal.md` under section 5.2 (Picod Workflow). The relevant design documents (`docs/design/PicoD-Plain-Authentication-Design.md` and `auth-proposal.md`) have been fully updated within this PR to accurately document the new `PICOD_BOOTSTRAP_PUBLIC_KEY`, `PICOD_SESSION_ID`, and the two-stage ECDSA handshake.

> **⚠️ Operator Migration Required**
> The Kubernetes ConfigMap `picod-router-public-key` has been renamed to `agentcube-bootstrap-identity`. The Kubernetes Secret `picod-router-identity` has been renamed to `agentcube-bootstrap-identity`. Secret data keys have been renamed from `private.pem`/`public.pem` → `bootstrap-private.pem`/`bootstrap-public.pem`. Operators upgrading from the old schema must migrate these resources before deploying.

The initialization order in `cmd/workload-manager/main.go` was intentionally changed: `CodeInterpreterReconciler` is now constructed *after* `NewServer` so it can receive `server.GetBootstrapPublicKeyPEM` as a live function reference rather than a nil pointer.

---

## Does this PR introduce a user-facing change?

```release-note
Implemented two-stage secure initialization for PicoD. Sandboxes are now cryptographically isolated using dynamic ECDSA (P-256) session keys to prevent cross-sandbox token replay vulnerabilities. Session private keys are now encrypted at rest in Redis/Valkey using AES-256-GCM. The PICOD_AUTH_PUBLIC_KEY env var is deprecated in favor of PICOD_BOOTSTRAP_PUBLIC_KEY.
```


acsoto:
Actually this reintroduces the old two-stage PicoD initialization flow from the outdated proposal. The current implementation was intentionally simplified in #115 to use startup-time key injection, and #136 removed the remaining Workload Manager `NeedInitialization` flow. This `/init` call brings back the extra post-ready dependency and failure mode that those PRs removed.

0xA8hinav:
Hi @acsoto
After looking back at #115 and #136, this flow is actually pretty different from the old initialization path that was removed. Those earlier changes removed PicoD’s `/init` + `NeedInitialization` flow because it was being used to push runtime config like workspace paths and user data, and PicoD could block on it indefinitely.

In this PR, `/init` is only used for a one-time cryptographic key exchange after `waitForSandboxEntryPointsReady` succeeds. There’s no polling loop, no persistent initialization flag/state, and the request is bounded with a 5s timeout. If it fails, sandbox creation rolls back through the normal `prepareSandbox` error path.

The main reason for adding it is to close a security gap in the current shared-key model: today all PicoD pods trust the same `PICOD_AUTH_PUBLIC_KEY`, which means a JWT issued for one sandbox could be replayed against another. The per-session keypair exchange here prevents that.

hzxuzhonghu (commented):
Important security improvement - two-stage initialization isolates bootstrap from session auth.

`pkg/picod/auth.go`
- Lines 36-38 (`ErrAlreadyInitialized`): Good sentinel error to prevent re-initialization attacks.
- Lines 41-42: Renaming `PICOD_AUTH_PUBLIC_KEY` to `PICOD_BOOTSTRAP_PUBLIC_KEY` is a breaking change for existing deployments. Add a migration note or support the old env var name as a fallback.
- Lines 59-76 (`parseRSAPublicKeyFromPEM`): Extracted helper is cleaner than duplicating PEM parsing. Good.
- Lines 79-92 (`LoadBootstrapPublicKey`): Same logic as before, just renamed. Clean.
- Lines 95-113 (`SetSessionPublicKey`): The `initialized` flag with mutex protection prevents race conditions. Once set, the session key cannot be overwritten. This is the core security invariant.
- Lines 116-145 (`VerifyBootstrapJWT`): Good - validates issuer (`agentcube-workload-manager`), requires expiration, and extracts the `session_public_key` claim. The 1-minute leeway is reasonable for clock skew.
- Lines 149-162 (`AuthMiddleware`): Returning 503 when not initialized is correct - it tells the client to retry. Make sure health/liveness probes are not routed through this middleware, otherwise the pod will be killed before init completes.

`pkg/common/types/sandbox.go`
- Line 41-44 (`SessionPrivateKey`): Using `json:"-"` to exclude the private key from serialization is critical. The comment explains the transient-only nature well. Good.

`pkg/picod/auth_test.go`
- Tests properly updated to use two-stage init flow (LoadBootstrapPublicKey + SetSessionPublicKey). Good coverage.

`pkg/picod/execute_test.go`
- Line 66: Test setup now includes `SetSessionPublicKey` after server creation. Consistent with the new flow.

LGTM. The env var rename needs migration documentation.

0xA8hinav:
cc @hzxuzhonghu

PR #355, opened 2026-05-20
test: cover workload builder runtime paths

Maximus-08:
**What type of PR is this?**
/kind enhancement

Adds unit coverage for workload builder paths, including AgentRuntime and CodeInterpreter sandbox construction, CodeInterpreter warm-pool behavior, PicoD auth env var injection, and not-found/error cases.
The tests also preserve and restore the cached public key during test execution to avoid leaking package-global state between tests.

Part of #103 

This is test-only coverage for existing workload builder behavior. No production behavior changes.
Validated with:
- `make vet`
- `make lint`
- `go test -race -v -coverprofile=coverage.out -coverpkg=./pkg/... ./pkg/...`
- `make docker-build`
- `make e2e`
I ran `make fmt` too it cleaned up the code and changed the indents and white spaces in a few files, not including those here. 

No

```release-note
NONE
```

### DIFF
--- a/pkg/workloadmanager/workload_builder.go
+++ b/pkg/workloadmanager/workload_builder.go
@@ -184,7 +184,7 @@ func buildSandboxObject(params *buildSandboxParams) *sandboxv1alpha1.Sandbox {
 			Labels: map[string]string{
 				SessionIdLabelKey:    params.sessionID,
 				WorkloadNameLabelKey: params.workloadName,
-				"managed-by":        "agentcube-workload-manager",
+				"managed-by":         "agentcube-workload-manager",
 			},
 			Annotations: map[string]string{
 				IdleTimeoutAnnotationKey: params.idleTimeout.String(),
@@ -386,8 +386,7 @@ func buildSandboxByCodeInterpreter(namespace string, codeInterpreterName string,
 				Namespace: namespace,
 				Name:      sandboxName,
 				Labels: map[string]string{
-					SessionIdLabelKey:    sessionID,
-					WorkloadNameLabelKey: codeInterpreterName,
+					SessionIdLabelKey: sessionID,
 				},
 			},
 		}
--- a/pkg/workloadmanager/workload_builder.go
+++ b/pkg/workloadmanager/workload_builder.go
@@ -386,7 +386,8 @@ func buildSandboxByCodeInterpreter(namespace string, codeInterpreterName string,
 				Namespace: namespace,
 				Name:      sandboxName,
 				Labels: map[string]string{
-					SessionIdLabelKey: sessionID,
+					SessionIdLabelKey:    sessionID,
+					WorkloadNameLabelKey: codeInterpreterName,
 				},
 			},
 		}
--- a/pkg/workloadmanager/workload_builder_test.go
+++ b/pkg/workloadmanager/workload_builder_test.go
@@ -248,6 +248,9 @@ func toUnstructured(t *testing.T, obj interface{}, kind string) *unstructured.Un
 	return u
 }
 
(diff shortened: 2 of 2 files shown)


Maximus-08:
cc @hzxuzhonghu @FAUST-BENCHOU

acsoto (approved):
LGTM

pkg/workloadmanager/workload_builder.go
```diff
@@ -386,7 +386,8 @@ func buildSandboxByCodeInterpreter(namespace string, codeInterpreterName string,
 				Namespace: namespace,
 				Name:      sandboxName,
 				Labels: map[string]string{
-					SessionIdLabelKey: sessionID,
+					SessionIdLabelKey:    sessionID,
+					WorkloadNameLabelKey: codeInterpreterName,
```
FAUST-BENCHOU: why need this if only add ut tests

Maximus-08: > medium
> 
> The test is missing an assertion for `WorkloadNameLabelKey` on the returned `sandbox` object. Additionally, looking at the implementation in `workload_builder.go:384`, it appears that the warm pool path for CodeInterpreters is indeed missing this label on the Sandbox object, which contradicts the PR's goal of ensuring this label is present across all CodeInterpreter sandbox paths.

So I added to address Gemini review which suggested standardizing label assertions and flagged the above suggestion. I refactored the tests to use a shared assertSandboxMetadata helper (which asserts WorkloadNameLabelKey), and added the label to simpleSandbox in workload_builder.go as suggested.

Should I revert the change to workload_builder.go and update the test assertions so we don't enforce this label on the warm-pool dummy struct ?

FAUST-BENCHOU (commented):
/lgtm

hzxuzhonghu (commented):
Good test coverage for the workload builder paths.

`pkg/workloadmanager/workload_builder_test.go`
- Lines 220-256: Helper functions (`fakeInformer`, `toUnstructured`, `setCachedPublicKeyForTest`, `assertSandboxMetadata`) are well factored. Note that `setCachedPublicKeyForTest` directly mutates the package-level `cachedPublicKey` behind a mutex - this is fine for serial test execution, but if tests ever run with `t.Parallel()`, this will race. Consider noting that in a comment.

- Lines 260-300 (`TestBuildCodeInterpreterEnvVars`): Table-driven, covers authMode none/picod/nil template env. Good.

- Lines 302-315 (`TestBuildSandboxByAgentRuntime_NotFound`): Verifies the correct sentinel error. Good.

- Lines 317-380 (`TestBuildSandboxByAgentRuntime_Success`): Validates sandbox metadata, pod labels, TTL, entry kind, idle timeout, and ports. Thorough.

- Lines 382-410 (`TestBuildSandboxByCodeInterpreter_PicodAuthFailsWithoutKey`): Validates that picod auth mode fails when no public key is cached. Good edge case.

- Lines 412-465 (`TestBuildSandboxByCodeInterpreter_SuccessNoWarmPool`): Confirms claim is nil for non-warm-pool path. Important assertion.

- Lines 467-530 (`TestBuildSandboxByCodeInterpreter_SuccessWithWarmPool`): Validates owner references and claim template ref. Good.

- Line 187: Whitespace alignment fix for the `managed-by` label is a nice cleanup.

LGTM.

Maximus-08:
Added the comment documenting the mutation and possible race @hzxuzhonghu

Maximus-08:
Sorry but the lgtm label was removed because I committed the comments

FAUST-BENCHOU (commented):
/lgtm

hzxuzhonghu (approved):

PR #358, opened 2026-05-22
perf(store): optimize redis session last activity check

Maximus-08:
**What type of PR is this?**
/kind cleanup

This PR optimizes Redis session tracking inside `UpdateSessionLastActivity` for performance and atomicity.
Previously, a full `GET` fetched the entire `SandboxInfo` JSON payload solely to check existence, introducing severe serialization and network overhead under concurrent routing

- Replaced sequential round-trips with an atomic Redis Lua script performing `EXISTS` and `ZADD` in a single network round-trip. This prevents race conditions.
-  Replaced ambiguous error strings with the standard `ErrNotFound` sentinel when a session is missing.
- Added coverage verifying missing-session error propagation.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/store/store_redis.go
+++ b/pkg/store/store_redis.go
@@ -297,6 +297,11 @@ func (rs *redisStore) Close() error {
 	return rs.cli.Close()
 }
 
+// updateSessionLastActivityScript atomically checks if the session key exists
+// and updates its last activity time in the sorted-set index.
+// It uses EXISTS to verify presence rather than GET (which retrieved the
+// actual value in older implementations). This saves data transfer overhead,
+// if in future the value is required GET can be used instead.
 var updateSessionLastActivityScript = redisv9.NewScript(`
 if redis.call("EXISTS", KEYS[1]) == 1 then
 	redis.call("ZADD", KEYS[2], ARGV[2], ARGV[1])
--- a/pkg/store/store_redis.go
+++ b/pkg/store/store_redis.go
@@ -297,6 +297,15 @@ func (rs *redisStore) Close() error {
 	return rs.cli.Close()
 }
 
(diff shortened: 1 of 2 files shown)


Maximus-08:
cc @hzxuzhonghu

acsoto (approved):
LGTM

hzxuzhonghu (commented):
Good optimization - atomically combines the EXISTS check and ZADD into a single Lua script, cutting a round trip.

`pkg/store/store_redis.go`
- Line 300-307: The Lua script is correct. One subtle difference from the original: the old code used `Get` (which also retrieves the value), while the new code uses `EXISTS` (which only checks presence). Functionally equivalent here since the value is not used, but worth noting in case future callers depend on the mapping being readable, not just present.

`pkg/store/store_redis_test.go`
- Line 347-355: Good - adds the missing non-existent session test case.

LGTM.

pkg/store/store_redis.go
```diff
@@ -297,6 +297,20 @@ func (rs *redisStore) Close() error {
 	return rs.cli.Close()
 }
 
+// updateSessionLastActivityScript atomically checks if the session key exists
+// and updates its last activity time in the sorted-set index.
+// It uses EXISTS to verify presence rather than GET (which retrieved the
+// actual value in older implementations). This saves data transfer overhead,
+// if in future the value is required GET can be used instead.
+var updateSessionLastActivityScript = redisv9.NewScript(`
```
hzxuzhonghu, line 305: can you also update the valkey store in a following pr?

hzxuzhonghu (approved):

hzxuzhonghu:
@acsoto You can label /lgtm , this label takes effect

acsoto:
ok
/lgtm

PR #360, opened 2026-05-26
feat: remove default 8h max TTL for AgentRuntime sandboxes

### LINKED ISSUE #303: Remove AgentRuntime default max ttl

As agent itself maybe long running, like if user want to host openclaw within agentcube `agentRuntime`, it cannot be deleted after 8 hour.

HarshitPal25:
## What this PR does

Fixes #303

### Problem

`AgentRuntime.spec.maxSessionDuration` had a kubebuilder default of `8h`, which caused the workload-manager to always set `Sandbox.Spec.Lifecycle.ShutdownTime` to `now + 8h`. This meant every AgentRuntime sandbox was hard-deleted after 8 hours, even if it was still actively serving a long-running agent (e.g. hosting OpenClaw or any persistent agent service).

### Solution

Make `maxSessionDuration` truly opt-in for `AgentRuntime`:

- **`agent_type.go`**: Removed `+kubebuilder:default="8h"` from `AgentRuntimeSpec.MaxSessionDuration`; added `+optional` and a clear comment that omitting it means no hard cap.
- **`workload_builder.go`**: `buildSandboxObject` no longer falls back to `DefaultSandboxTTL` when `ttl == 0`. `ShutdownTime` is only set when `MaxSessionDuration` is explicitly configured by the user.
- **`sandbox_helper.go`**: `buildSandboxPlaceHolder` and `buildSandboxInfo` no longer fall back to `DefaultSandboxTTL` when `ShutdownTime` is absent. `ExpiresAt` is left as zero time (meaning no hard expiry).
- **`store_redis.go`**: Skips adding to the expiry sorted-set when `ExpiresAt` is zero, so the GC loop never forcibly expires these sandboxes.
- **`store_valkey.go`**: Same — zero `ExpiresAt` bypasses the expiry index, only tracking the session in the activity index for idle-timeout cleanup.
- **Tests**: Updated to reflect the new behaviour: no `ShutdownTime` → zero `ExpiresAt` → sandbox runs indefinitely, cleaned up only by `sessionTimeout` idle logic.

`CodeInterpreter` is **unchanged** — it retains the 8h default at the `buildSandboxByCodeInterpreter` call-site.

### Behaviour after this change

| Resource | `maxSessionDuration` set? | Hard delete? |
|---|---|---|
| AgentRuntime | No (default) | Never (idle timeout only) |
| AgentRuntime | Yes (e.g. `24h`) | After 24h |
| CodeInterpreter | No (default) | After 8h (unchanged) |
| CodeInterpreter | Yes | After configured duration |

### DIFF
--- a/manifests/charts/base/crds/runtime.agentcube.volcano.sh_agentruntimes.yaml
+++ b/manifests/charts/base/crds/runtime.agentcube.volcano.sh_agentruntimes.yaml
@@ -44,10 +44,11 @@ spec:
             description: Spec defines the desired state of the AgentRuntime.
             properties:
               maxSessionDuration:
-                default: 8h
                 description: |-
                   MaxSessionDuration describes the maximum duration for a session.
                   After this duration, the session will be terminated no matter active or inactive.
+                  When omitted, sessions have no maximum duration and will only be cleaned up
+                  by idle timeout (SessionTimeout).
                 type: string
               podTemplate:
                 description: PodTemplate describes the template that will be used
@@ -8504,7 +8505,6 @@ spec:
                   type: object
                 type: array
             required:
-            - maxSessionDuration
             - podTemplate
             - sessionTimeout
             - targetPort
--- a/pkg/apis/runtime/v1alpha1/agent_type.go
+++ b/pkg/apis/runtime/v1alpha1/agent_type.go
@@ -53,8 +53,9 @@ type AgentRuntimeSpec struct {
 
 	// MaxSessionDuration describes the maximum duration for a session.
 	// After this duration, the session will be terminated no matter active or inactive.
-	// +kubebuilder:validation:Required
-	// +kubebuilder:default="8h"
+	// When omitted, sessions have no maximum duration and will only be cleaned up
+	// by idle timeout (SessionTimeout).
+	// +optional
 	MaxSessionDuration *metav1.Duration `json:"maxSessionDuration,omitempty" protobuf:"bytes,3,opt,name=maxSessionDuration"`
 }
 
--- a/pkg/store/store_redis.go
+++ b/pkg/store/store_redis.go
@@ -155,16 +155,16 @@ func (rs *redisStore) StoreSandbox(ctx context.Context, sandboxRedis *types.Sand
 		return fmt.Errorf("StoreSandbox: marshal sandbox failed: %w", err)
 	}
 
-	if sandboxRedis.ExpiresAt.IsZero() {
-		return fmt.Errorf("StoreSandbox: sandbox expired at is zero")
-	}
-
 	pipe := rs.cli.Pipeline()
 	pipe.SetNX(ctx, sessionKey, b, 0)
-	pipe.ZAdd(ctx, rs.expiryIndexKey, redisv9.Z{
-		Score:  float64(sandboxRedis.ExpiresAt.Unix()),
-		Member: sandboxRedis.SessionID,
-	})
+	// Only add to expiry index when ExpiresAt is set.
+	// Zero ExpiresAt means the sandbox has no maximum lifetime.
+	if !sandboxRedis.ExpiresAt.IsZero() {
+		pipe.ZAdd(ctx, rs.expiryIndexKey, redisv9.Z{
+			Score:  float64(sandboxRedis.ExpiresAt.Unix()),
+			Member: sandboxRedis.SessionID,
+		})
+	}
 	pipe.ZAdd(ctx, rs.lastActivityIndexKey, redisv9.Z{
 		Score:  float64(time.Now().Unix()),
 		Member: sandboxRedis.SessionID,
--- a/pkg/store/store_valkey.go
+++ b/pkg/store/store_valkey.go
@@ -174,14 +174,26 @@ func (vs *valkeyStore) StoreSandbox(ctx context.Context, sandboxStore *types.San
 		return errors.New("StoreSandbox: sandbox is nil")
 	}
 
-	if sandboxStore.ExpiresAt.IsZero() {
-		return fmt.Errorf("StoreSandbox: sandbox expires time is zero")
-	}
-
 	sessionKey := vs.sessionKey(sandboxStore.SessionID)
 	b, err := json.Marshal(sandboxStore)
 	if err != nil {
 		return fmt.Errorf("StoreSandbox: marshal sandbox: %w", err)
+	}
+
+	if sandboxStore.ExpiresAt.IsZero() {
+		// Zero ExpiresAt means no maximum lifetime — skip expiry index.
+		// The sandbox will still be tracked in the activity index for idle timeout.
+		commands := make(valkey.Commands, 0, 3)
+		commands = append(commands, vs.cli.B().Setnx().Key(sessionKey).Value(string(b)).Build())
+		commands = append(commands, vs.cli.B().Zadd().Key(vs.lastActivityIndexKey).ScoreMember().
+			ScoreMember(float64(time.Now().Unix()), sandboxStore.SessionID).Build())
+
+		for i, resp := range vs.cli.DoMulti(ctx, commands...) {
+			if err = resp.Error(); err != nil {
+				return fmt.Errorf("StoreSandbox: DoMulti failed: %w, command index: %v", err, i)
+			}
+		}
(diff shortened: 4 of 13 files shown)


pkg/apis/runtime/v1alpha1/agent_type.go
```diff
@@ -53,8 +53,9 @@ type AgentRuntimeSpec struct {
 
 	// MaxSessionDuration describes the maximum duration for a session.
 	// After this duration, the session will be terminated no matter active or inactive.
-	// +kubebuilder:validation:Required
-	// +kubebuilder:default="8h"
+	// When omitted, sessions have no maximum duration and will only be cleaned up
+	// by idle timeout (SessionTimeout).
+	// +optional
```
acsoto, line 58: This needs the generated CRD artifact update as well. Right now the installed AgentRuntime CRD still has `default: 8h` and still lists `maxSessionDuration` as required, so the API server will keep defaulting omitted values to 8h and this change will not actually fix #303 in a deployed cluster.

pkg/store/store_redis.go
```diff
@@ -155,16 +155,16 @@ func (rs *redisStore) StoreSandbox(ctx context.Context, sandboxRedis *types.Sand
...
-
 	pipe := rs.cli.Pipeline()
 	pipe.SetNX(ctx, sessionKey, b, 0)
-	pipe.ZAdd(ctx, rs.expiryIndexKey, redisv9.Z{
-		Score:  float64(sandboxRedis.ExpiresAt.Unix()),
-		Member: sandboxRedis.SessionID,
-	})
+	// Only add to expiry index when ExpiresAt is set.
+	// Zero ExpiresAt means the sandbox has no maximum lifetime.
+	if !sandboxRedis.ExpiresAt.IsZero() {
```
acsoto, line 162: Can we add store tests for the zero `ExpiresAt` branch? The important behavior is that the session is still stored and tracked in the last-activity index, but is not added to the expiry index.

pkg/workloadmanager/workload_builder.go
```diff
@@ -198,12 +193,18 @@ func buildSandboxObject(params *buildSandboxParams) *sandboxv1alpha1.Sandbox {
...
 		},
 	}
+
+	// Only set ShutdownTime when MaxSessionDuration is explicitly configured.
+	// When omitted (ttl == 0), the sandbox runs indefinitely and is only
+	// cleaned up by idle timeout (SessionTimeout).
+	if params.ttl > 0 {
+		shutdownTime := metav1.NewTime(time.Now().Add(params.ttl))
+		sandbox.Spec.Lifecycle.ShutdownTime = &shutdownTime
+	}
```
acsoto, line 206: This makes `ttl == 0` mean no hard expiry for every caller of `buildSandboxObject`. That is correct for AgentRuntime, but CodeInterpreter is supposed to keep the 8h max-session default from the issue discussion and PR description. Can we set `DefaultSandboxTTL` explicitly in the CodeInterpreter paths when `Spec.MaxSessionDuration` is nil, including the warm-pool placeholder path?

pkg/store/store_redis.go
HarshitPal25: yes ofc we can i was fixing the failure test cases till now i would add store test for the zero expireat now

pkg/workloadmanager/workload_builder.go
FAUST-BENCHOU, line 182: revert all this kind of fmt and rebase to the newest upstream first.We have solved all fmt problem in https://github.com/volcano-sh/agentcube/commit/f17fa1731016e86b319e8dd9f4a1972b06e947ec i think

pkg/workloadmanager/sandbox_helper_test.go
```diff
@@ -40,7 +40,7 @@ func TestBuildSandboxPlaceHolder_TableDriven(t *testing.T) {
 		validate     func(t *testing.T, result *types.SandboxInfo)
 	}{
 		{
-			name: "no ShutdownTime falls back to DefaultSandboxTTL",
+			name: "no ShutdownTime means no expiry (sandbox runs indefinitely)",
```
FAUST-BENCHOU, line 43: why not remove all DefaultSandboxTTL in ut test like in workload_builder_test.go L44 since we are not using it anymore

manifests/charts/base/crds/runtime.agentcube.volcano.sh_agentruntimes.yaml
```diff
@@ -8504,7 +8505,6 @@ spec:
                   type: object
                 type: array
             required:
-            - maxSessionDuration
```
FAUST-BENCHOU, line 8507: also upd example/agent-runtime/agent-runtime.yaml

pkg/workloadmanager/workload_builder.go
FAUST-BENCHOU: +1 It's better to align with CodeInterpreter.

FAUST-BENCHOU (commented):
recommand to rebase first

test/e2e/echo_agent.yaml
FAUST-BENCHOU, line 1: never used?

pkg/store/store_valkey.go
```diff
@@ -174,16 +174,28 @@ func (vs *valkeyStore) StoreSandbox(ctx context.Context, sandboxStore *types.San
...
+		// The sandbox will still be tracked in the activity index for idle timeout.
+		commands := make(valkey.Commands, 0, 3)
+		commands = append(commands, vs.cli.B().Setnx().Key(sessionKey).Value(string(b)).Build())
+		commands = append(commands, vs.cli.B().Zadd().Key(vs.lastActivityIndexKey).ScoreMember().
+			ScoreMember(float64(time.Now().Unix()), sandboxStore.SessionID).Build())
+
+		for i, resp := range vs.cli.DoMulti(ctx, commands...) {
+			if err = resp.Error(); err != nil {
+				return fmt.Errorf("StoreSandbox: DoMulti failed: %w, command index: %v", err, i)
+			}
```
hzxuzhonghu, line 194: I am seeing this is a little different from redis, is that because valkey sdk does not align with redis?

pkg/workloadmanager/workload_builder.go
hzxuzhonghu: @HarshitPal25 ptal

hzxuzhonghu (commented):
Another point from me, we can later expose sandbox lifecycle managerment api through sdk. Otherwise for agentruntime, user may not be able to delete the sandbox

pkg/store/store_valkey.go
HarshitPal25: the structural difference is purely due to the Valkey SDK's API design, not a logic difference.

Redis (go-redis/v9) has a Pipeline() abstraction that lets you queue a variable number of commands and skip some conditionally before calling Exec():

```
pipe := rs.cli.Pipeline()
pipe.SetNX(...)
if !sandboxRedis.ExpiresAt.IsZero() {
    pipe.ZAdd(...) // conditionally added
}
pipe.ZAdd(...) // lastActivity always added
pipe.Exec(ctx)
```

Valkey (valkey-go) uses DoMulti() which takes a fixed slice of commands built upfront. Since Go slices can't have "holes", the cleanest pattern with this SDK is to branch the two cases (zero vs non-zero ExpiresAt) into separate command slices with an early return, rather than trying to conditionally append inside a single block:

```
if sandboxStore.ExpiresAt.IsZero() {
    commands := [SETNX, ZADD lastActivity]
    vs.cli.DoMulti(ctx, commands...)
    return nil
}
commands := [SETNX, ZADD expiryIndex, ZADD lastActivity]
vs.cli.DoMulti(ctx, commands...)
```
The end result is identical behaviour — zero ExpiresAt skips the expiry index in both stores. The branching is an SDK-driven implementation detail, not a semantic difference.

HarshitPal25:
@hzxuzhonghu Agreed! This is a great point. Since maxSessionDuration is now opt-in, AgentRuntime sandboxes can run indefinitely — users will need a way to explicitly delete them. Exposing a sandbox lifecycle management API (like a delete/terminate session call) through the SDK would address this nicely. I'll open a separate follow-up issue to track it so it doesn't hold up this PR.

PR #367, opened 2026-05-28
feat: end-to-end Keycloak authentication, RBAC, and Python SDK integration

mahil-2040:
**What type of PR is this?**

/kind feature

This PR introduces end-to-end external user authentication and authorization using Keycloak as the identity provider. It expands the initial scope to include the complete flow in a single PR: Keycloak deployment, Router integration, Resource-Level Access Control (RLAC), Python SDK support, and the formal design proposal.
### 1. Design Proposal
- Adds `docs/design/keycloak-proposal.md` detailing the architecture, request flow, and implementation decisions for the external authentication layer.
### 2. Keycloak Deployment
- Declarative realm import with `agentcube` realm, role hierarchy (`sandbox:invoke` → `sandbox:manage` → `admin`), and OAuth2 clients.
- Production-hardened Deployment with security contexts, health probes, and Helm validation guards.
### 3. Router OIDC Integration & RBAC
- OIDC token validator using `go-oidc` with automatic JWKS key caching.
- Authentication middleware validates Bearer JWT tokens from Keycloak.
- RBAC middleware enforces `sandbox:invoke` realm role on all `/v1/*` endpoints.
- Feature-gated behind `--enable-external-auth` (default: off, zero impact).
### 4. Downstream Identity Forwarding & RLAC
- Router embeds user identity (`sub`) in internal PicoD JWT claims.
- Router sends `X-AgentCube-User-Id` header to WorkloadManager.
- **WorkloadManager (RLAC):** Tags newly created sandbox CRDs with the `agentcube.io/owner: <user_id>` label.
- **Router (RLAC):** Verifies that the caller's JWT subject matches the sandbox's `OwnerID` before proxying requests.
### 5. Python SDK Auth Integration
- Introduces pluggable auth provider pattern (`AuthProvider` protocol).
- Adds `ServiceAccountAuth` for `client_credentials` flow with automatic token refresh.
- Adds `TokenAuth` for static tokens.
- Updates clients (Control Plane and Data Plane) to seamlessly handle token injection.
### 6. Testing
- Unit tests for OIDC validator, auth middleware, RBAC, identity forwarding, and SDK auth.
- E2E tests deploying real Keycloak in Kind cluster to validate unauthenticated rejection, invalid tokens, valid invocations, and RLAC ownership blocks.

Fixes Part of #243 

- External auth is **opt-in** via `keycloak.enabled=true`. Existing deployments are unaffected.
- Token Exchange (RFC 8693) is not used - the Router forwards the user identity directly since mTLS provides transport-level security between components.
- New Go dependency: `github.com/coreos/go-oidc/v3` (used by Kubernetes itself).

yes

```release-note
Add Keycloak OIDC authentication support, Resource-Level Access Control (RLAC), and Python SDK auth integration. When enabled via `keycloak.enabled=true`, the Router requires a valid Bearer token with the `sandbox:invoke` realm role, and users can only access sandboxes they created. The Python SDK now supports seamless `client_credentials` authentication with automatic token refresh.
```


manifests/charts/addons/keycloak/templates/realm-config.yaml
```diff
@@ -0,0 +1,106 @@
...
+            }
+          }
+        ]
+      },
+
+      "defaultRoles": ["sandbox:invoke"],
+
+      "clients": [
+        {
+          "clientId": "agentcube-sdk",
```
acsoto: The confidential client secrets should be explicit. If Keycloak generates them during realm import, SDK/router/admin clients have no stable secret to configure against after Helm install.

manifests/charts/base/templates/keycloak/keycloak.yaml
```diff
@@ -0,0 +1,130 @@
...
+            - name: KC_BOOTSTRAP_ADMIN_USERNAME
+              value: {{ .Values.keycloak.adminUser | quote }}
+            - name: KC_BOOTSTRAP_ADMIN_PASSWORD
+              {{- if .Values.keycloak.existingSecret }}
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .Values.keycloak.existingSecret }}
+                  key: {{ .Values.keycloak.existingSecretKey | default "admin-password" }}
+              {{- else }}
+              value: {{ .Values.keycloak.adminPassword | quote }}
```
acsoto: This still renders the bootstrap admin password into the Deployment when no Secret is set. We should require a Secret when Keycloak is enabled instead of keeping the plaintext fallback.

manifests/charts/base/templates/keycloak/keycloak.yaml
```diff
@@ -0,0 +1,130 @@
...
+            - name: KC_DB_USERNAME
+              value: {{ .Values.keycloak.database.username | quote }}
+            - name: KC_DB_PASSWORD
+              {{- if .Values.keycloak.database.existingSecret }}
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .Values.keycloak.database.existingSecret }}
+                  key: {{ .Values.keycloak.database.existingSecretKey | default "db-password" }}
+              {{- else }}
+              value: {{ .Values.keycloak.database.password | quote }}
```
acsoto: Same for the database password. The production path should require `database.existingSecret` instead of rendering the DB credential as a literal env value.

manifests/charts/base/templates/keycloak/keycloak.yaml
```diff
@@ -0,0 +1,130 @@
...
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      containers:
+        - name: keycloak
+          image: "{{ .Values.keycloak.image.repository }}:{{ .Values.keycloak.image.tag }}"
+          imagePullPolicy: {{ .Values.keycloak.image.pullPolicy }}
+          {{- if .Values.keycloak.devMode }}
+          args: ["start-dev", "--import-realm"]
+          {{- else }}
+          args: ["start", "--import-realm"]
```
acsoto: `devMode=false` should fail template rendering when required production settings are missing. Otherwise Helm succeeds but Keycloak starts with an incomplete production config.

hzxuzhonghu (commented):
before adding a bunch of yamls, can you first add a docs exmplain about the proposal and the workflow

mahil-2040:
> before adding a bunch of yamls, can you first add a docs exmplain about the proposal and the workflow

@hzxuzhonghu @acsoto 
Yeah sure, it's a good suggestion, so now should i delete this PR, and make a fresh PR in which i will first make a doc explaining about the proposal and the workflow, and when it is approved then only i will push the `.yaml` code, or should i continue with this only this time, and follow this approach of making a doc from next feature?

acsoto:
Either is ok

mahil-2040:
> Either is ok

Okay i will continue in this one only for this time.

hzxuzhonghu:
@mahil-2040 I donot see your design

mahil-2040:
> @mahil-2040 I donot see your design

@hzxuzhonghu @acsoto 
I have now added the complete end-to-end design proposal for the Keycloak authentication as requested.

My plan is to implement the entire Keycloak authentication flow within this PR. Please review the desing doc and tell me if any changes are required.

Once the proposed approach is reviewed and approved, I'll move forward with the implementation in this same PR.

Please let me know your feedback on the design when you have a chance to review it. Thanks!

docs/design/keycloak-proposal.md
```diff
@@ -0,0 +1,299 @@
...
+    SDK->>KC: POST /realms/agentcube/protocol/openid-connect/token<br/>(client_credentials: agentcube-service + client_secret)
+    KC-->>SDK: Access Token (JWT)
+
+    Note over SDK, PicoD: 2. Authenticated Invocation
+    SDK->>Router: POST /v1/namespaces/.../invocations/...<br/>Authorization: Bearer <keycloak_jwt>
+
+    Note over Router: 3. Edge Authentication
+    Router->>Router: Validate JWT signature (cached JWKS)
+    Router->>Router: Check expiry, issuer, audience
+    Router->>Router: Extract realm_access.roles
```
hzxuzhonghu: How do you extract this role? Does the attribute auto signed into the jwt claim?

docs/design/keycloak-proposal.md
```diff
@@ -0,0 +1,299 @@
...
+
+    Note over Router: 4. RBAC Check
+    Router->>Router: Require "sandbox:invoke" role
```
hzxuzhonghu: THis is a litttle bit hard coding. This means we cannot use this authz midware to adapt to other idp provider

docs/design/keycloak-proposal.md
```diff
@@ -0,0 +1,299 @@
...
+
+    Note over Router, WM: 5. Session Creation (if new)
+    Router->>Router: Sign identity JWT (sub, iss, aud, exp)
```
hzxuzhonghu, line 46: Not quite understand, can you explain

docs/design/keycloak-proposal.md
```diff
@@ -0,0 +1,299 @@
...
+    Router->>WM: POST /v1/agent-runtime<br/>Authorization: Bearer <k8s_sa_token><br/>X-AgentCube-User-Identity: <router_signed_jwt>
+    WM->>WM: Verify identity JWT (Router public key)<br/>Record ownership (annotation + hashed label)
+    WM-->>Router: Sandbox info (sessionId, endpoints, ownerId)
+
+    Note over Router: 6. RLAC Check (if existing session)
```
hzxuzhonghu, line 51: what is rlac short for

docs/design/keycloak-proposal.md
```diff
@@ -0,0 +1,299 @@
...
+    PicoD-->>Router: Response
+    Router-->>SDK: Response
+```
+
+### Key decisions
+
+**JWKS-based offline validation.** The Router fetches Keycloak's public signing keys via OIDC discovery at startup. The `go-oidc` library caches and auto-rotates these keys - no per-request call to Keycloak.
+
+**Forwarding user identity via signed tokens, not plain headers.** The Router embeds the user's `sub` claim into existing channels:
+- For PicoD: added as a `user_sub` claim in the Router-signed internal JWT
```
hzxuzhonghu: why add? You need to document it

docs/design/keycloak-proposal.md
```diff
@@ -0,0 +1,299 @@
...
+- For WorkloadManager: sent as a short-lived Router-signed identity JWT in the `X-AgentCube-User-Identity` header
```
hzxuzhonghu: what is this value? IIUC, `X-AgentCube-User-Identity: <router_signed_jwt>` you mean tranparently pass the token, then workload manager also need to parse this jwt?

docs/design/keycloak-proposal.md
```diff
@@ -0,0 +1,299 @@
...
+
+Downstream services trust the identity because it is cryptographically signed by the Router's private key (the same key used for PicoD auth). WM verifies this JWT using the Router's public key from the `picod-router-public-key` ConfigMap — no dependency on mTLS for identity trust.
```
hzxuzhonghu: not sure i understand this sentence

docs/design/keycloak-proposal.md
```diff
@@ -0,0 +1,299 @@
...
+
+## Detailed Design
+
+### 1. Keycloak Helm Deployment
+
+Keycloak runs as a single-replica Deployment inside the AgentCube namespace, gated behind `{{- if .Values.keycloak.enabled }}`. The Helm chart **bootstraps** the initial realm, clients, and roles — it is not a declarative management tool. After first startup, realm changes must be made via the Keycloak Admin API or Admin Console.
```
hzxuzhonghu: I think it is not reasonable, cannot i run multiple replicas to be HA?

docs/design/keycloak-proposal.md
```diff
@@ -0,0 +1,299 @@
...
+
+The Deployment runs `quay.io/keycloak/keycloak:26.0.8` and supports two modes:
```
hzxuzhonghu: In a proposal, i think no explicit version needed

docs/design/keycloak-proposal.md
```diff
@@ -0,0 +1,299 @@
...
+        └── sandbox:invoke  (invoke agent runtimes and code interpreters)
+```
+
+`sandbox:invoke` is assigned to the default realm role, so every new user gets it automatically.
+
+**OAuth2 Clients:**
+
+| Client ID | Type | Purpose |
+|-----------|------|---------|
+| `agentcube-service` | Confidential (`client_credentials`) | Server-side / automation authentication |
```
hzxuzhonghu: represent which component or user?

docs/design/keycloak-proposal.md
```diff
@@ -0,0 +1,299 @@
...
+| `agentcube-sdk` | Public (`authorization_code` + PKCE) | Interactive SDK and CLI authentication |
```
hzxuzhonghu: This represent the end user?

hzxuzhonghu (commented):
I would also suggest separating keycloack as a separate addon, not place it in agentcube chart

docs/design/keycloak-proposal.md
mahil-2040: Yes, Keycloak automatically injects the roles into the JWT by default.

When a user or service account authenticates, Keycloak looks at the `scopeMappings` we defined in our Helm chart, grabs the `sandbox:invoke` role, and bakes it into a `realm_access.roles` JSON array inside the token payload.

Because Keycloak cryptographically signs the entire token, the Router can trust it. Our Go code just uses the `go-oidc` library to verify the signature, and then simply parses the JSON to check if `sandbox:invoke` is inside that array.

docs/design/keycloak-proposal.md
mahil-2040: The idea here is to safely pass the user's identity from the Router down to the WorkloadManager (WM) so WM knows who 'owns' the sandbox it is creating.

Instead of passing the heavy Keycloak token to WM (which would force WM to also connect to Keycloak to verify it), the Router acts as the middleman:

1. The Router verifies the Keycloak token.
2. It extracts the User ID (`sub`).
3. The Router creates a tiny, new, short-lived JWT containing just that User ID, and signs it using its own internal private key.

When WM receives this new token, it only has to verify the Router's signature (using the Router's public key, which it already has). This keeps our architecture simple: the Router handles all external Keycloak auth, and internal services only ever have to trust the Router.

docs/design/keycloak-proposal.md
mahil-2040: RLAC stands for **Resource-Level Access Control**.

In our system, we use both RBAC and RLAC together:

We need RLAC so that if User A and User B both have the `sandbox:invoke` role, User A is strictly blocked from connecting to a sandbox that was created by User B.

docs/design/keycloak-proposal.md
mahil-2040: I have updated the proposal to document the reason explicitly. We need to pass the `user_sub` down to PicoD (the agent runtime) so that it can trace actions back to a specific human user for auditing and logging. This also gives the AI agent context about who it is interacting with, and allows PicoD to forward that identity if it makes downstream calls to other internal services that require authorization.

docs/design/keycloak-proposal.md
mahil-2040: You are right , hardcoding `realm_access.roles` ties our authorization middleware to Keycloak, which is not provider-agnostic. Other OIDC providers use completely different claim paths for roles (e.g., Okta uses `groups`, Auth0 uses a custom URL namespace claim).

To fix this and make the middleware truly adapt to any IDP:

1. I have added a new required CLI flag to the Router: `--oidc-roles-claim`.
2. In Go, instead of unmarshaling into a hardcoded `RealmAccess` struct, we parse the JWT payload dynamically and extract the roles array using the JSON path provided by this flag.
3. In the Helm chart, I have added a validation rule: if `router.oidc.issuerUrl` is set, `router.oidc.rolesClaim` is strictly required (it has no default value). This fail-fast approach ensures Okta/Auth0 users don't accidentally deploy with a broken Keycloak default.

I have updated the design proposal and sequence diagrams to reflect this dynamic extraction. Thank you for catching this!

docs/design/keycloak-proposal.md
mahil-2040: No, we do not transparently pass the external OIDC token.

Instead, the Router creates a brand new, short-lived internal JWT containing just the user's `sub`, signs it with the Router's own private key, and sends that to WorkloadManager.

Yes, WorkloadManager has to parse this internal JWT, but it only verifies the Router's signature (using the Router's public key from a ConfigMap it already mounts). The benefit of this is that WorkloadManager remains completely decoupled from the external IDP. It never has to fetch external JWKS keys or know about Keycloak/Okta , the Router handles 100% of the external OIDC complexity, and internal services only have to trust the Router.

docs/design/keycloak-proposal.md
mahil-2040: This sentence means is that WorkloadManager and PicoD need a way to trust that the user identity they are receiving is actually real, and not a spoofed fake header sent by a malicious user.

They can trust it because the Router cryptographically signs the JWT with its own private key. WorkloadManager simply checks that signature using the Router's public key.

I have updated the wording in the proposal so this is easily understandable!

docs/design/keycloak-proposal.md
mahil-2040: You are right, a single replica isn't suitable for production HA.

I hardcoded it to 1 just to keep the local dev environment simple, but you are right. I have updated the addon chart to expose a replicas field in `values.yaml` (defaulting to 1).

If a user wants production HA, they can now deploy the addon chart with `--set replicas=3` and provide the external database configuration. I've also updated the proposal document to reflect this.

docs/design/keycloak-proposal.md
mahil-2040: This client does not represent an internal AgentCube component (we use `agentcube-router` and `agentcube-admin` for internal service identities).

`agentcube-service` is for external users building backend applications or automation scripts. For example, if a user writes a background Python script or their own backend server that needs to call the AgentCube API programmatically, they cannot use an interactive browser login. They would use this agentcube-service client ID along with a Client Secret to authenticate automatically (Machine-to-Machine / M2M flow).

I have updated the table in the proposal to explicitly say 'External backend applications and automated scripts using the Python SDK (Machine-to-Machine / M2M)' so it is clear who this is for.

docs/design/keycloak-proposal.md
mahil-2040: Yes.

The `agentcube-sdk` client represents human end users. For example, when a user runs an interactive script or CLI command, the SDK uses this client ID. It pops open a browser window for the user to type in their Keycloak username and password.

(This is why it is configured as a 'Public' client with PKCE, because it runs on the user's local machine and can't safely store a client secret).

I have updated the proposal table to explicitly say 'Human end users interacting via the CLI or browser' so this is completely clear!

mahil-2040:
> I would also suggest separating keycloack as a separate addon, not place it in agentcube chart

Yeah, This is a great suggestion, it makes sense to keep the core agentcube chart lightweight and identity-provider agnostic!

I have made the following changes to implement this:

1. Moved all Keycloak templates out of the base chart into a new standalone addon chart (`manifests/charts/addons/keycloak`).
2. Removed all keycloak configurations from the base `values.yaml`.

docs/design/keycloak-proposal.md
hzxuzhonghu: Will you introduce a private key in this proposal

mahil-2040: No, we will not introduce a new private key in this proposal.

We will simply reuse the existing RSA private key that the Router already generates and manages for PicoD authentication. The Router's `JWTManager` (in `pkg/router/jwt.go`) already stores this keypair in the `picod-router-identity` Secret.

We will just use that exact same existing private key to sign the `X-AgentCube-User-Identity` JWT sent to WorkloadManager. WorkloadManager will just need to read the public key from that existing Secret to verify the signature.

mahil-2040:
@hzxuzhonghu @acsoto Does the design look good? Should I start implementing now?

docs/design/keycloak-proposal.md
hzxuzhonghu: Still see sandbox:invoke"

mahil-2040: fixed it by introducing the `--oidc-required-role` flag

mahil-2040:
@acsoto @hzxuzhonghu Please review this!!

cmd/router/main.go
```diff
@@ -44,6 +44,13 @@ func main() {
 		mtlsCert = flag.String("mtls-cert", "", "Path to mTLS client certificate for upstream WorkloadManager connections")
 		mtlsKey  = flag.String("mtls-key", "", "Path to mTLS client key for upstream WorkloadManager connections")
 		mtlsCA   = flag.String("mtls-ca", "", "Path to mTLS CA bundle for verifying upstream WorkloadManager identity")
+
+		// OIDC configuration for external user authentication.
+		// External auth is automatically enabled when --oidc-issuer-url is provided.
+		oidcIssuerURL    = flag.String("oidc-issuer-url", "", "OIDC provider issuer URL, enables external auth when set")
+		oidcAudience     = flag.String("oidc-audience", "agentcube-api", "Expected audience (aud) claim in the access token")
+		oidcRolesClaim   = flag.String("oidc-roles-claim", "", "JSON path to roles array in the JWT (e.g., realm_access.roles)")
+		oidcRequiredRole = flag.String("oidc-required-role", "", "Role required to access the API (e.g., sandbox:invoke)")
```
hzxuzhonghu: Prefer update to 

oidc-issuer-url -> jwt-issuer-url
oidc-roles-claim -> jwt-role-claim
oidc-required-role -> jwt-required-role, this may need to be deleted, once we support authz, we can put iy in the api

docs/agentcube/docs/tutorials/external-auth-keycloak.md
```diff
@@ -0,0 +1,441 @@
...
+
+This task shows you how to add external authentication and authorization to
+AgentCube using [Keycloak](https://www.keycloak.org/) as the identity provider.
+By the end, every request to the Router API will require a valid JWT token
+issued by Keycloak, and access will be controlled by realm roles (RBAC) and
+resource ownership (RLAC).
+
+## Before you begin
+
+1. Follow the [Getting Started](../getting-started.md) guide to install
```
hzxuzhonghu, line 11: Following the getstarted, it will not install spire as you showed in L33-L36

docs/agentcube/docs/tutorials/external-auth-keycloak.md
```diff
@@ -0,0 +1,441 @@
...
+| `keycloak` | Service (ClusterIP) | Exposes Keycloak on port 8080 inside the cluster. |
+| `keycloak-realm-config` | Secret | Contains the realm JSON with roles, clients, and scope mappings. |
+| `keycloak-credentials` | Secret | Stores admin password and client secrets. |
+
+The Keycloak addon also auto-configures the following inside the `agentcube`
+realm:
+
+| Item | Details |
+|---|---|
+| **Roles** | `sandbox:invoke` (default for all users), `sandbox:manage` (inherits invoke), `admin` (inherits manage) |
```
hzxuzhonghu: elaborate a little bit more, what is the permission of each role

manifests/charts/addons/keycloak/values.yaml
```diff
@@ -0,0 +1,69 @@
...
+  # They define which keys inside your external Secret hold the actual passwords.
+  existingSecretServiceKey: "client-service-secret"
+  existingSecretRouterKey: "client-router-secret"
+  existingSecretAdminKey: "client-admin-secret"
+  service:
+    secret: ""
+  router:
+    secret: ""
+  admin:
+    secret: ""
```
hzxuzhonghu: please explain what's these secret are, and how to setup them

docs/agentcube/docs/tutorials/external-auth-keycloak.md
```diff
@@ -0,0 +1,441 @@
...
+```
+
+## Step 5 - Make an authenticated request
+
+Now use the token to make an authenticated request through the Router:
+
+```bash
+curl -s -w "\nHTTP Status: %{http_code}\n" \
+  -H "Authorization: Bearer $KEYCLOAK_TOKEN" \
+  http://localhost:8081/v1/namespaces/default/agent-runtimes/sample-agent/invocations/
```
hzxuzhonghu, line 316: I doubt this is not actually run at all, the invocation is not complete

docs/design/keycloak-proposal.md
```diff
@@ -0,0 +1,330 @@
...
+        └── sandbox:invoke  (invoke agent runtimes and code interpreters)
+```
+
+`sandbox:invoke` is assigned to the default realm role, so every new user gets it automatically.
+
+**OAuth2 Clients:**
+
+| Client ID | Type | Purpose |
+|-----------|------|---------|
+| `agentcube-service` | Confidential (`client_credentials`) | External backend applications and automated scripts using the Python SDK (Machine-to-Machine / M2M) |
```
hzxuzhonghu: confusing, 
does it mean caller to agentcube-service?

Or agentcube-service is the client?

docs/design/keycloak-proposal.md
```diff
@@ -0,0 +1,330 @@
...
+| `agentcube-sdk` | Public (`authorization_code` + PKCE) | Human end users interacting via the CLI or browser (Interactive) |
+| `agentcube-router` | Confidential (`client_credentials`) | Internal Router service identity |
+| `agentcube-admin` | Confidential (`client_credentials`) | Administrative operations |
```
hzxuzhonghu: all these client-id need to be explained

manifests/charts/addons/keycloak/values.yaml
```diff
@@ -0,0 +1,69 @@
...
+# Set to false for production — requires database.vendor and proxy settings below.
+devMode: true
+image:
+  repository: quay.io/keycloak/keycloak
+  tag: "26.0.8"
+  pullPolicy: IfNotPresent
+adminUser: ""           # Required.
+adminPassword: ""       # Dev only. Production must use existingSecret.
+existingSecret: ""       # Name of a pre-existing Secret containing admin credentials
+existingSecretKey: "admin-password"  # Key within the Secret holding the password
```
hzxuzhonghu: `existingSecret` is a bad name? Can you imagine how a user know the meaning of it from this name?

hzxuzhonghu (commented):
I have not read the implement, please address these comments first and help me understand some concepts. Thank you

docs/agentcube/docs/tutorials/external-auth-keycloak.md
mahil-2040: Apologies for the confusion! I had a previous installation in the `agentcube-system` namespace with `spire.enabled=true`, so when I captured that kubectl get pods output, SPIRE pods were already running. I mistakenly used that output in the tutorial thinking it reflected the base install but it doesn't. i have fixed this, it will just show the router and workloadmanager pods

manifests/charts/addons/keycloak/values.yaml
mahil-2040: Agree, the top-level existingSecret was ambiguous because the chart has multiple Secret types.

I moved the admin credentials under an admin block so the purpose is clear.

Now `admin.existingSecret` clearly means the existing Kubernetes Secret that contains the Keycloak bootstrap admin password. The other Secret fields are already scoped by purpose, like `database.existingSecret`, `clients.existingSecret`, and `tls.existingSecret`.

manifests/charts/addons/keycloak/values.yaml
mahil-2040: These are OAuth2 client secrets for the confidential Keycloak clients created by the addon chart. They are not Kubernetes service account tokens. Keycloak uses them when a caller requests an access token via the `client_credentials` grant.

We have three confidential clients:
1. `agentcube-service` for SDK/server-side token requests, 
2. `agentcube-router` reserved for Router service-to-service flows, and 
3. `agentcube-admin` for administrative tokens.

There are two setup options:

1. For dev/test, set the values directly:
`clients.service.secret`, `clients.router.secret`, and `clients.admin.secret`. The chart creates the `keycloak-credentials` Kubernetes Secret.

2. For production, create a Kubernetes Secret yourself and set `clients.existingSecret`. The `existingSecretServiceKey`, `existingSecretRouterKey`, and `existingSecretAdminKey` values tell the chart which keys inside that Secret contain each client secret.

docs/design/keycloak-proposal.md
mahil-2040: I expanded the OAuth2 clients table to explain each client ID: who uses it, which OAuth flow it uses, and why it exists. I also clarified that the confidential clients are intentionally separated by trust boundary, so external service credentials, Router internal credentials, and admin credentials can be scoped and rotated independently.

docs/agentcube/docs/tutorials/external-auth-keycloak.md
mahil-2040: Done!

docs/design/keycloak-proposal.md
mahil-2040: `agentcube-service` is an OAuth2 client ID registered in Keycloak. External backend apps and SDK users authenticate as this client to get a token, which they then send to the AgentCube Router. The old table mixed the client identity with who uses it. I've split it into separate "Who uses it" and "Purpose" columns in the updated table to make this clearer.

cmd/router/main.go
mahil-2040: renamed the flags.

docs/agentcube/docs/tutorials/external-auth-keycloak.md
mahil-2040: I did run the command, but since the sample agent in the getting started guide is just a Python HTTP server (`python3 -m http.server`), the response body was a verbose HTML directory listing of the container's root filesystem. I originally omitted that HTML from the expected output to keep the documentation clean, focusing only on the `HTTP Status: 200`.

To avoid any confusion and make the output match the command exactly, I have updated the command in the tutorial to redirect the response body to `/dev/null` using `-o /dev/null`.

This way, it prints only the `HTTP Status: 200` without the noisy HTML body

manifests/charts/addons/keycloak/values.yaml
hzxuzhonghu: In the user guide, we should have a step by step how to generate them. I still not get the secret format, x509 or other format?

docs/design/keycloak-proposal.md
hzxuzhonghu: got it by ` OAuth2 client ID`

hzxuzhonghu: Let's use professional words

docs/agentcube/docs/tutorials/external-auth-keycloak.md
```diff
@@ -0,0 +1,438 @@
...
+`sandbox:invoke`, and `aud` containing `agentcube-api`:
+
+```
+{
+  "exp": 1780695969,
+  "iat": 1780695669,
+  "jti": "31ef712f-f62a-4a4e-8714-7ccf710ab476",
+  "iss": "http://keycloak.agentcube.svc.cluster.local:8080/realms/agentcube",
+  "aud": "agentcube-api",
+  "sub": "bb0f4c04-e1d2-4bdb-b7bb-56d5c1cefe50",
```
hzxuzhonghu, line 275: How can we specify the user identity when signing jwt token

docs/design/keycloak-proposal.md
```diff
@@ -0,0 +1,330 @@
...
+        └── sandbox:invoke  (invoke agent runtimes and code interpreters)
+```
+
+`sandbox:invoke` is assigned to the default realm role, so every new user gets it automatically.
+
+**OAuth2 Clients:**
+
+| Client ID | Type | Who uses it | Purpose |
+|-----------|------|-------------|---------|
+| `agentcube-service` | Confidential (`client_credentials`) | External backend services, CI jobs, automation scripts, and SDK users that can safely store a client secret | Main machine-to-machine client for calling AgentCube APIs. Tokens from this client receive `sandbox:invoke` and are used by `ServiceAccountAuth` in the Python SDK. |
```
hzxuzhonghu: I think this is actually  agentcube clients that access agentcube router, so `agentcube-service` is not suitable, any other app client name make sense.

docs/design/keycloak-proposal.md
```diff
@@ -0,0 +1,330 @@
...
+| `agentcube-sdk` | Public (`authorization_code` + PKCE) | Human users authenticating from a CLI, browser, or other installed app where a client secret cannot be safely stored | Interactive login client. It uses PKCE(Proof Key for Code Exchange) instead of a client secret and is intended for future browser/device/CLI login flows. |
```
hzxuzhonghu: Explain the workflow?

hzxuzhonghu: Also confusing, since it seems has nothing to do with sdk

docs/design/keycloak-proposal.md
```diff
@@ -0,0 +1,330 @@
...
+
+Confidential client secrets can be provided securely via an existing Kubernetes Secret (`keycloak.clients.existingSecret`) to prevent leaking them in Helm values. They are injected into the Keycloak pod as environment variables and securely substituted into the realm JSON during import. The `agentcube-sdk` client is a **public client** (no secret) that uses authorization code with PKCE for interactive flows, following RFC 8252 (OAuth 2.0 for Native Apps). The confidential clients are used only where a secret can be stored securely, and each client ID is separated by trust boundary so service, internal Router, and admin credentials can be scoped and rotated independently.
+
+Both clients include a **hardcoded audience protocol mapper** (`oidc-audience-mapper`) that injects `agentcube-api` into the access token's `aud` claim. The Router validates `aud = "agentcube-api"`. This follows OAuth2 convention — the audience identifies the resource server (the Router API), not the client that requested the token.
```
hzxuzhonghu: Both clients means what. I can guess, but not clear

docs/design/keycloak-proposal.md
```diff
@@ -0,0 +1,330 @@
...
+```go
+type OIDCConfig struct {
+    IssuerURL  string // e.g. "http://keycloak.agentcube-system.svc:8080/realms/agentcube"
+    Audience   string // expected "aud" claim, e.g. "agentcube-api"
+    RolesClaim string // dot-separated path to roles array, e.g. "realm_access.roles"
+}
+
+type Claims struct {
+    Subject string   // from standard "sub" claim
+    Email   string   // from standard "email" claim
```
hzxuzhonghu, line 129: How can user add email claim when using keycloak? Can you point where it is documented

manifests/charts/addons/keycloak/templates/realm-config.yaml
hzxuzhonghu, line 1: Other than a static config, can we still use keycloak api to dynamically register other roles?

hzxuzhonghu (commented):
Reviewed the Keycloak auth / RLAC changes. Overall the OIDC validation, RBAC middleware, and SDK auth look solid. Main concern: the sandbox CRD is never actually tagged with the owner label/annotation, even though the PR description and the builder code suggest it should be. RLAC still works because the owner is persisted in the store, but the kubectl-level owner labels won't be present. Left a few inline notes.

pkg/workloadmanager/handlers.go
```diff
@@ -120,6 +120,11 @@ func (s *Server) handleSandboxCreate(c *gin.Context, kind string) {
 		return
 	}
 
+	// Set ownership from the Router-signed identity JWT
+	if ownerID := extractOwnerID(c.Request); ownerID != "" {
+		sandboxEntry.OwnerID = ownerID
```
hzxuzhonghu, line 141: The owner is set on `sandboxEntry` here, but the `sandbox` CR object was already built above (lines 108/110) before we know the owner. So `buildSandboxParams.ownerID` stays empty and the `agentcube.io/owner` annotation / `agentcube.io/owner-hash` label are never applied to the actual CRD. RLAC still works via the store's `OwnerID`, but the CR labeling described in the PR doesn't happen. Consider extracting `ownerID` before building, then passing it into the builder.

pkg/workloadmanager/workload_builder.go
```diff
@@ -203,6 +205,13 @@ func buildSandboxObject(params *buildSandboxParams) *sandboxv1alpha1.Sandbox {
 			Replicas: ptr.To[int32](1),
 		},
 	}
+
+	// Ownership metadata for RLAC
+	if params.ownerID != "" {
+		sandbox.ObjectMeta.Annotations["agentcube.io/owner"] = params.ownerID
```
hzxuzhonghu, line 211: This block is effectively dead code today: `params.ownerID` is never assigned anywhere, so this `if` is always false and the owner annotation/label is never written. The owner only ends up on the `sandboxEntry` (store), not on the CR. Wiring `ownerID` into `buildSandboxParams`/`buildSandboxClaimParams` would make this work as intended.

pkg/workloadmanager/identity.go
```diff
@@ -0,0 +1,105 @@
...
+	publicKey := GetCachedPublicKey()
+	if publicKey == "" {
+		klog.V(2).Info("Identity JWT present but public key not cached, skipping owner extraction")
+		return ""
+	}
+
+	sub, err := verifyIdentityJWT(publicKey, rawToken)
+	if err != nil {
+		klog.V(2).Infof("Identity JWT verification failed: %v", err)
+		return ""
```
hzxuzhonghu: `extractOwnerID` fails open here: if the public key isn't cached yet or verification fails, it returns an empty owner and the sandbox is created with no owner. On the Router side, RLAC fails closed (no owner = access denied), so such a sandbox becomes permanently unreachable by its creator. Worth deciding on one consistent behavior, or at least logging at a higher level so this is easy to spot.

pkg/router/oidc.go
```diff
@@ -0,0 +1,253 @@
...
+	// Validate subject
+	if subject == "" {
+		return nil, fmt.Errorf("token missing required sub claim")
+	}
+
+	// Validate expiration
+	if expiry == 0 {
+		return nil, fmt.Errorf("token missing required exp claim")
+	}
+	if time.Now().After(time.Unix(int64(expiry), 0)) {
```
hzxuzhonghu: nit: `exp`/`nbf` are compared against `time.Now()` with no clock-skew leeway. A small allowance (e.g. 30s) avoids spurious rejections when the IdP and Router clocks drift slightly.

sdk-python/agentcube/auth.py
```diff
@@ -0,0 +1,94 @@
...
+            data["scope"] = self._scope
+
+        logger.debug(f"Fetching token from {self._token_url}")
+        resp = requests.post(self._token_url, data=data, headers=self._headers, timeout=self._timeout)
+        resp.raise_for_status()
+
+        body = resp.json()
+        self._token = body["access_token"]
+        expires_in = int(body.get("expires_in", 3600))
+        self._expires_at = time.monotonic() + expires_in - _REFRESH_BUFFER_SECONDS
```
hzxuzhonghu: nit: if the IdP returns `expires_in <= 30`, `_expires_at` lands in the past and every `get_token()` call will re-fetch a token. Clamping the buffer (e.g. `max(expires_in - buffer, expires_in / 2)`) would avoid hammering the token endpoint for short-lived tokens.

docs/agentcube/docs/tutorials/external-auth-keycloak.md
mahil-2040: The `sub` claim is not specified by the caller, Keycloak sets it automatically. In this example we use the `client_credentials` grant, so the `sub` is the UUID that Keycloak assigns to the service account of the `agentcube-service` client. For real users (e.g. via `authorization_code` grant), `sub` would be the user's UUID from Keycloak. The caller cannot forge or choose the `sub`, it is signed into the JWT by Keycloak.

docs/design/keycloak-proposal.md
mahil-2040: Okay, i will keep that in mind.

manifests/charts/addons/keycloak/values.yaml
mahil-2040: Updated the tutorial to clarify the secret format and generation. The client secrets are plain-text OAuth2 shared secrets (similar to API keys). Added a step showing how to generate strong secrets for production using `openssl rand -base64 32`.

docs/design/keycloak-proposal.md
mahil-2040: You are right, `agentcube-service` is misleading since it's not an internal AgentCube service but an OAuth2 client for external applications accessing the Router API.

I have renamed it to `agentcube-app`.

if you prefer something else please tell me i will update the code according to that.

docs/design/keycloak-proposal.md
mahil-2040: You're right, the name `agentcube-sdk` is also misleading as the Python SDK actually uses the current `agentcube-service` client, not this one.

This is a public OAuth2 client for future interactive human login flows (CLI, browser) where a client secret cannot be safely stored. The `authorization_code` + PKCE workflow is:

1. User runs a CLI command or opens a browser app
2. App redirects user to the Keycloak login page
3. User enters username/password in Keycloak
4. Keycloak redirects back with a one-time authorization code
5. App exchanges the code for an access token, using a PKCE code verifier. random challenge generated at step (2) instead of a client secret.

This flow is not implemented yet. it's a placeholder for future interactive authentication. I have renamed it to `agentcube-cli`. If you have some better name then please let me know.

docs/design/keycloak-proposal.md
mahil-2040: The email claim is a standard OIDC claim, no special setup is needed. Keycloak includes it automatically via the built-in email client scope, which is a default scope for all clients. The user just needs to have an email address set in their Keycloak profile.

For service accounts (`client_credentials` grant), there is no real user, so email will be empty. The Router handles this gracefully, it only forwards the email to sandboxes when present (`oidc.go:115`, `handlers.go:291`). It is not a required claim.

See Keycloak docs: [Server Administration - Client Scopes](https://www.keycloak.org/docs/latest/server_admin/#_client_scopes)

manifests/charts/addons/keycloak/templates/realm-config.yaml
mahil-2040: Yes. The static realm config is only used for initial bootstrap on first startup. After that, roles, clients, users, scopes, etc. can be managed dynamically through: 

1. Keycloak Admin REST API. : e.g., create a new role:

```
POST /admin/realms/agentcube/roles
{"name": "custom:role"}
```
2. Keycloak Admin Console, web UI at `/admin` for point-and-click management.

The Router does not need any changes when new roles are added, it only checks that the configured `requiredRole` exists in the token's role claim. Any additional roles can be used for application-level authorization downstream.

YOu can check here: [Keycloak Admin REST API - Roles](https://www.keycloak.org/docs-api/latest/rest-api/index.html#_roles)

pkg/workloadmanager/identity.go
mahil-2040: I've updated the log level from `V(2)` to `klog.Warning` for both failure paths (key not cached + verification failed), so this is now easy to spot in production.

pkg/workloadmanager/handlers.go
mahil-2040: I moved the `extractOwnerID` call to the beginning of the handler and passed the extracted `ownerID` into `buildSandboxByAgentRuntime` and `buildSandboxByCodeInterpreter` before building the sandbox objects. This ensures that the owner metadata is successfully populated on both the Sandbox CRD and the store entry.

sdk-python/agentcube/auth.py
mahil-2040: Fixed. Clamped the effective cache lifetime to `max(expires_in - buffer, expires_in // 2)`, so for short-lived tokens the refresh window is at least half the token lifetime instead of going negative and hammering the token endpoint.

pkg/router/oidc.go
mahil-2040: Fixed. Added a `clockSkewLeeway = 30s` constant and applied it to both `exp` and `nbf` comparisons. `exp` is extended by 30s, `nbf` is shifted back by 30s. This handles minor clock drift between the IdP and Router without accepting truly expired tokens.

hzxuzhonghu (approved):
/lgtm

PR #372, opened 2026-06-02
manifests: load REDIS_PASSWORD from Secret via secretKeyRef

### LINKED ISSUE #368: Redis password stored in plain text
https://github.com/volcano-sh/agentcube/blob/c7adda884c8936586ff5dedab05bd1578fa9e9e9/manifests/charts/base/templates/workloadmanager.yaml#L63

**What happened:**

I found that in the Agentcube project, the Redis password is stored in plain text within the Pod YAML configuration in the workloadmanager.yaml file, which poses a security risk.

**What you expected to happen:**

The recommended approach is to store sensitive information such as Redis passwords in Kubernetes Secrets and reference them through environment variables or volume mounts, rather than embedding them directly in YAML configuration files. This follows established security practices and ensures proper protection of credentials.

kubeboiii:
**What type of PR is this?**

Security fix for Helm chart credential handling.

The Helm chart embedded the Redis password directly in Deployment env `value` fields for workloadmanager and agentcube-router. Anyone with read access to Deployments in the namespace could read the credential from `kubectl get deployment -o yaml`.

This PR stores the password in a Kubernetes Secret and references it with `secretKeyRef` for `REDIS_PASSWORD` in both components. When `redis.secretName` is unset and `redis.password` is set, the chart creates a Secret named `{release}-redis`. When `redis.secretName` is set, the chart skips Secret creation and uses the provided Secret name and key. When neither is set, no Secret or `REDIS_PASSWORD` env is rendered (supports e2e and dev installs with `REDIS_PASSWORD_REQUIRED=false`).

No changes to application code. Components continue to read `REDIS_PASSWORD` from the environment.

Fixes #368

- Chart-only change under `manifests/charts/base/` plus `docs/getting-started.md`.
- Backward compatible: `helm install --set redis.password=...` still works; the value is written to a Secret instead of the Deployment manifest.
- New values: `redis.secretName`, `redis.secretKey` (default key: `password`). Use either `redis.password` or `redis.secretName`, not both.
- E2E installs that use `--set redis.password=""` with `REDIS_PASSWORD_REQUIRED=false` are unchanged in behavior.
- Using `--set redis.password` still places the value in Helm release metadata. Production installs should prefer a pre-created Secret via `redis.secretName` (documented in getting-started).

Yes. Helm installs now create or reference a Redis Secret for credentials instead of inlining the password in Deployment specs.

```release-note
Redis credentials for workloadmanager and agentcube-router are loaded from a Kubernetes Secret via secretKeyRef instead of plain-text Deployment env values. Chart-managed Secret `{release}-redis` is created when `redis.secretName` is not set and `redis.password` is provided; use `redis.secretName` to supply your own Secret in production.
```


manifests/charts/base/values.yaml
```diff
@@ -9,7 +9,13 @@ fullnameOverride: ""
 # These must be provided by the user during installation
 redis:
   addr: ""
+  # Populates the chart-managed Secret (not plain text in Deployment manifests).
+  # For production, prefer creating a Secret yourself and set existingSecret instead.
   password: ""
+  # Name of an existing Secret containing the Redis password (skips chart Secret creation).
+  existingSecret: ""
```
hzxuzhonghu: why call existingSecret?

manifests/charts/base/values.yaml
```diff
@@ -9,7 +9,13 @@ fullnameOverride: ""
...
+  # Key within existingSecret or chart-managed Secret (default: password).
+  existingSecretPasswordKey: "password"
```
hzxuzhonghu: ditto

manifests/charts/base/templates/_helpers.tpl
hzxuzhonghu, line 1: This is not a must

manifests/charts/base/templates/_helpers.tpl
```diff
@@ -0,0 +1,17 @@
...
+{{- .Values.redis.existingSecret }}
+{{- else }}
+{{- printf "%s-redis" .Release.Name }}
+{{- end }}
+{{- end }}
+
+{{/*
+Key within the Redis Secret that holds the password.
+*/}}
+{{- define "agentcube.redis.passwordKey" -}}
```
hzxuzhonghu: I think redefining another var makes it harder to understand

kubeboiii:
Thanks @hzxuzhonghu for the review — addressed in the latest commits:

- Renamed `redis.existingSecret` / `redis.existingSecretPasswordKey` to `redis.secretName` and `redis.secretKey`.
- Removed `templates/_helpers.tpl` and inlined the Secret name/key logic in `redis-secret.yaml`, `workloadmanager.yaml`, and `agentcube-router.yaml`.
- Behavior for #368 is unchanged: no plain-text `REDIS_PASSWORD` in Deployment manifests; credentials come from a Secret via `secretKeyRef`.

Also added a short note in `values.yaml` and getting-started (mutual exclusivity of `password` vs `secretName`, upgrade rolling restart).

manifests/charts/base/values.yaml
kubeboiii: Renamed to `redis.secretName` — clearer than `existingSecret`. When set, the chart does not create `{release}-redis` and both Deployments use that Secret via `secretKeyRef`.

manifests/charts/base/values.yaml
kubeboiii: Renamed to `redis.secretKey` (default `password`). Used directly in templates with `| default "password"` instead of a helper.

manifests/charts/base/templates/_helpers.tpl
kubeboiii: Removed `_helpers.tpl` and inlined the logic in the three templates as suggested.

manifests/charts/base/templates/_helpers.tpl
kubeboiii: Dropped the `passwordKey` helper; templates now use `.Values.redis.secretKey | default "password"` directly.

hzxuzhonghu (approved):
I think we can also update the e2e with password

/lgtm

PR #378, opened 2026-06-05
Fix flaky router private key PEM test

googs1025:
Fixes a flaky router test seen in coverage:
https://github.com/volcano-sh/agentcube/actions/runs/26986069745/job/79635856049

The parsed RSA key is equivalent, but CRT internals can differ by leading zero bytes. Use rsa.PrivateKey.Equal for the assertion.


hzxuzhonghu (approved):
/lgtm

PR #384, opened 2026-06-13
ci: add formatting and Helm lint checks

googs1025:
## What this PR does
- Adds a `gofmt` job to the existing lint workflow, running `make fmt-check` for Go formatting verification.
- Adds a `helm-lint` job to validate the base Helm chart with `make helm-lint`.
- Renders the base Helm chart with `make helm-template` to catch template errors without flooding CI logs.
- Uses path filters so these checks only run for relevant Go, chart, Makefile, or workflow changes.

This follows up on review feedback from #383: some developer lint commands should also run in CI.

## Verification
- `make fmt-check`
- `make helm-lint`
- `make helm-template >/tmp/agentcube-rendered.yaml`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/lint.yml")'`
- `git diff --check`


.github/workflows/lint.yml
```diff
@@ -40,3 +40,59 @@ jobs:
       - name: Run golangci-lint
         if: steps.changes.outputs.lint == 'true'
         run: make lint
+
+  gofmt:
+    name: gofmt
+    runs-on: ubuntu-24.04
```
LiZhenCheng9527: suggest ubuntu-latest

LiZhenCheng9527 (commented):
Can it be written into the Makefile?

.github/workflows/lint.yml
googs1025: done

googs1025:
> Can it be written into the Makefile?

yes, already add in https://github.com/volcano-sh/agentcube/pull/383

googs1025:
@hzxuzhonghu /PTAL

Issue #386, opened 2026-06-15
[v0.2.0] Call for proposals


RainbowMango:
This is the **umbrella tracking issue** for the AgentCube **v0.2.0** release.

AgentCube has reached its first set of working components — **Workload Manager** (control plane), **Router** (data plane), **PicoD** (runtime), the `AgentRuntime` and `CodeInterpreter` CRDs, a Redis/ValKey-backed session store, mTLS between internal components, plus a Python SDK, CLI, and integrations (LangChain, Dify, MCP). The goal of **v0.2.0** is to harden these foundations, close gaps in the existing features, and improve the developer/operator experience.

> **This issue is open to the whole community.** Anyone is welcome to propose new **features**, **enhancements**, **cleanups**, **docs**, or **test** improvements. Comment below with your proposal , and we'll triage it into the checklist and create a dedicated sub-issue.

---

## 📋 Candidate Work Items

> Checkboxes are checked as the work merges. Items are **candidates** until triaged — feel free to vote 👍 or propose changes in the comments.

### Feautres 
- [ ] TBD

### Security & Auth
- [x] Automate Alpine base image updates with Dependabot. (@ranxi2001, #422)
  * Some Dockerfiles in this repository use Alpine as the runtime base image, for example docker/Dockerfile and docker/Dockerfile.router. These base image versions can become outdated over time, which may increase security maintenance burden and leave known vulnerabilities unpatched.
  * We should introduce an automated mechanism to keep Docker base images up to date, especially Alpine images.
  * A possible solution is to configure Dependabot for Docker image updates, similar to [Karmada’s approach](https://github.com/karmada-io/karmada/blob/d01d3a8fd305eb9eab48621c06e9171aed9d6748/.github/dependabot.yml#L18-L25).

### Cleanup/Testing/CI
- [x] Pin GitHub Actions Ubuntu runners. (@ranxi2001, #423)
- [x] Update Golang to v1.26.4 (@ranxi2001, #391)
- [x] Remove unused `agentd` component. (@ranxi2001, #403)
- [x] Enable CI workflow in fork repos. (@ranxi2001, #414)
- [x] Introduce proposal template. (@ranxi2001, #415)
- [x] Fix failing chart workflow. (@ranxi2001, #416)
- [x] Performance improvement for multi-arch images.(@ranxi2001, #420)

---

## How to Propose Something

Just comment on this issue and tell us, in your own words:

[...]

PR #387, opened 2026-06-17
fix: adapt code interpreter warm pool to agent-sandbox v0.4.6

ranxi2001:
**What type of PR is this?**

/kind feature

This PR updates AgentCube's CodeInterpreter warm pool integration from `agent-sandbox v0.1.1` to the current stable Go module release, `v0.4.6`. In v0.4.6, a `SandboxClaim` adopts a pre-warmed `Sandbox` and exposes that serving Sandbox through status, so AgentCube must resolve the claim to the adopted Sandbox instead of assuming the claim or template name identifies the runtime Pod.

The change updates the dependency and Kubernetes generator stack, uses the public Sandbox Pod annotation, waits for the adopted Sandbox to become ready, and probes its Pod endpoint. AgentCube still stores the claim name for delete and GC while using runtime data from the adopted Sandbox. CodeInterpreter-created templates explicitly use `networkPolicyManagement: Unmanaged` to preserve the existing Router and WorkloadManager traffic path. Warm pool e2e discovery now supports both the old direct ownership shape and `SandboxWarmPool -> Sandbox -> Pod`.

Refs #386

- Scope: this PR targets stable `v0.4.6` and its `v1alpha1` APIs. The `v0.5.x` / `v1beta1` migration is separate because it replaces `TemplateRef` with required `WarmPoolRef` and `Replicas` with `OperatingMode`.
- Generated code: the CRD and client-go diffs align with Kubernetes `v0.35.4`; there is no intentional AgentRuntime API change.
- Validation: focused workload-manager unit and race tests, non-e2e Go tests, build/lint/codegen checks, k3s direct and warm-pool scenarios, and Python integrations passed at `c2633c5`.
- AI assistance: Codex helped inspect the compatibility changes, tests, and PR text; I reviewed the code and validation results.

```release-note
NONE
```


zhzhuang-zju:
Hi @ranxi2001, thanks for your work.

First, this PR looks more like a feature than a bug fix. Please confirm and update the PR label if needed.

> I did not target v0.5.0rc1 because Go resolves that tag to a pseudo-version rather than the stable latest release.

Could you elaborate on this a bit more? Specifically, at which step would this become a problem? `agent-sandbox` also introduced substantial changes in `v0.5.0rc1`. So once `agent-sandbox v0.5.0` is officially released, we may need to do another substantial round of adaptation work.

pkg/workloadmanager/handlers.go
```diff
@@ -195,6 +197,71 @@ func (s *Server) createK8sResources(ctx context.Context, dynamicClient dynamic.I
 	return nil
 }
 
+func (s *Server) waitForDirectSandboxReady(ctx context.Context, sandbox *sandboxv1alpha1.Sandbox, resultChan <-chan SandboxStatusUpdate) (*sandboxv1alpha1.Sandbox, error) {
```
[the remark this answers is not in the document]
ranxi2001: Thanks, this was a valid point. I added an explicit `resultChan == nil` guard so a direct sandbox watcher wiring issue now returns `errSandboxReadyWatcherNotRegistered` immediately instead of being reported as a generic 2-minute sandbox creation timeout.

I also covered the related edge cases in `TestWaitForDirectSandboxReadyWatcherFailures`:
  - nil watcher channel
  - closed watcher channel
  - watcher event with a nil sandbox

This keeps the direct sandbox path failure mode immediate and diagnosable.

pkg/workloadmanager/handlers_test.go
```diff
@@ -100,6 +103,21 @@ func makeEntry() *sandboxEntry {
...
+
+func (f *recordingStore) UpdateSandbox(ctx context.Context, sandbox *types.SandboxInfo) error {
+	f.fakeStore.UpdateSandbox(ctx, sandbox)
+	if f.updateErr != nil {
+		return f.updateErr
+	}
+	copied := *sandbox
+	f.lastUpdated = &copied
+	return nil
+}
```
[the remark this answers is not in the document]
ranxi2001: Thanks, this was a valid test-fake issue. I updated `recordingStore.UpdateSandbox` to propagate the embedded store error and to deep-copy `EntryPoints` before saving `lastUpdated`.

This prevents the test from hiding an `UpdateSandbox` failure or asserting against a later-mutated slice.

zhzhuang-zju:
thanks
/assign

RainbowMango:
@ranxi2001 could you please resolve the conflict?

ranxi2001:
Reviewer note: why the E2E change is larger than a version bump.

The previous job installed `agent-sandbox v0.1.1` and enabled mTLS, which caused the direct WorkloadManager CodeInterpreter tests to be skipped. The updated workflow keeps the existing security-path coverage and adds a focused feature path:

```mermaid
flowchart LR
    A["e2e-test<br/>mTLS enabled"] --> B["AgentRuntime<br/>Router to WorkloadManager mTLS"]
    C["codeinterpreter-e2e-test<br/>mTLS disabled"] --> D["CodeInterpreter session"]
    D --> E["SandboxClaim"]
    E --> F["adopted Sandbox"]
    F --> G["pre-warmed Pod<br/>same UID"]
    G --> H["session delete"]
    H --> I["Claim, Sandbox, and Pod removed<br/>warm pool refilled"]
```

Both jobs install and verify `agent-sandbox-controller:v0.4.6`. The additional assertions at `df8eb9b` protect control identity versus runtime identity, prove reuse of the same pre-warmed Pod, and verify session cleanup and refill. This test update does not change auth or RBAC behavior. Both paths were validated on `ubuntu-24.04`.

pkg/workloadmanager/handlers.go
```diff
@@ -245,55 +347,47 @@ func (s *Server) createSandbox(ctx context.Context, dynamicClient dynamic.Interf
...
+	// If a warm pool is used, the adopted Pod name is stored in
+	// sandboxv1alpha1.SandboxPodNameAnnotation.
+	sandboxNameForPod := createdSandbox.Name
+	sandboxPodName := createdSandbox.Name
+	if podName, exists := createdSandbox.Annotations[sandboxv1alpha1.SandboxPodNameAnnotation]; exists {
 		sandboxPodName = podName
 	}
 
-	podIP, err := s.k8sClient.GetSandboxPodIP(ctx, sandbox.Namespace, sandbox.Name, sandboxPodName)
+	podIP, err := s.k8sClient.GetSandboxPodIP(ctx, createdSandbox.Namespace, sandboxNameForPod, sandboxPodName)
```
acsoto, line 428: This reads the claim and Sandbox from the live API, then switches to a separate Pod informer cache that may still lag the Ready status update. A valid warm claim can therefore fail once and be rolled back; please live-GET the annotated Pod or retry Pod discovery until the context expires.

pkg/workloadmanager/handlers.go
```diff
@@ -221,6 +243,86 @@ func (s *Server) createK8sResources(ctx context.Context, dynamicClient dynamic.I
...
+}
+
+func (s *Server) waitForClaimSandboxReady(ctx context.Context, dynamicClient dynamic.Interface, sandboxClaim *extensionsv1alpha1.SandboxClaim) (*sandboxv1alpha1.Sandbox, error) {
+	timer := time.NewTimer(2 * time.Minute) // consistent with router settings
+	defer timer.Stop()
+	ticker := time.NewTicker(1 * time.Second)
+	defer ticker.Stop()
+
+	for {
+		claim, err := getSandboxClaim(ctx, dynamicClient, sandboxClaim.Namespace, sandboxClaim.Name)
```
acsoto: This retries every GET error until the two-minute timer, including permanent Forbidden or conversion errors, and then reports them as a readiness timeout. Please retry only NotFound/transient API errors and return permanent failures immediately.

pkg/workloadmanager/handlers.go
ranxi2001: Thanks, reproduced this with a live API Pod at Running/IP while the informer still held a Pending/no-IP copy. Explicit Pod names now use a live API GET before IP validation, while the label-cache fallback remains for lookups without a Pod name. I also added a regression test for the stale-cache case.

pkg/workloadmanager/handlers.go
ranxi2001: Thanks, reproduced this with a Forbidden Claim GET being masked by the readiness timeout. Claim and adopted Sandbox reads now retry only NotFound and explicit transient API/transport errors; Forbidden, Unauthorized, and conversion failures return immediately as sanitized internal errors. Focused tests cover both read stages and the retry classification.

RainbowMango (approved):
/approve

PR #398, opened 2026-06-22
workloadmanager: honor default code interpreter auth mode

### LINKED ISSUE #397: CodeInterpreter authMode default is skipped in non-warm sandbox creation
`authMode` is documented/defaulted as `picod`, but the non-warm-pool CodeInterpreter path only enables auth when the field is exactly `"picod"`.

That means an empty authMode can skip the public key check and skip `PICOD_AUTH_PUBLIC_KEY` injection. The warm-pool path handles this differently: anything except `"none"` gets the key.

Steps:
1. Create/build a CodeInterpreter object with authMode omitted/empty.
2. Use the normal non-warm-pool sandbox path.
3. Check generated pod env.

Expected:
Empty authMode behaves like `picod`, and `PICOD_AUTH_PUBLIC_KEY` is injected.

Actual:
The direct sandbox path treats empty authMode as not picod and does not inject the key.

Fix should probably normalize authMode in one helper and use it in both warm-pool and direct sandbox paths.

avinxshKD:
/kind bug

Keeps CodeInterpreter auth handling consistent between warm-pool and non-warm sandbox creation.

The non-warm path only treated explicit `picod` as PicoD auth. This makes the default/empty value behave the same way, matching the warm-pool path: anything except `none` gets the PicoD public key handling.

Fixes #397

Added coverage for empty/default authMode in env injection and missing-key handling.

```release-note
NONE

### DIFF
--- a/pkg/workloadmanager/workload_builder.go
+++ b/pkg/workloadmanager/workload_builder.go
@@ -309,11 +309,11 @@ func buildSandboxByAgentRuntime(namespace string, name string, ownerID string, i
 }
 
 // buildCodeInterpreterEnvVars copies the template env vars and injects the
-// public key when authMode is picod.
+// public key unless authMode is none.
 func buildCodeInterpreterEnvVars(templateEnv []corev1.EnvVar, authMode runtimev1alpha1.AuthModeType) []corev1.EnvVar {
 	envVars := make([]corev1.EnvVar, len(templateEnv))
 	copy(envVars, templateEnv)
-	if authMode == runtimev1alpha1.AuthModePicoD {
+	if authMode != runtimev1alpha1.AuthModeNone {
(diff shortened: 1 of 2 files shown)


avinxshKD:
Hey @RainbowMango @hezhangjian Pls take a look,

Found this while checking the CodeInterpreter auth paths.

Warm-pool already treats empty authMode as default PicoD auth, but the non-warm builder only handled explicit `picod`. 
This keeps both paths consistent.

cc @acsoto @YaoZengzeng

avinxshKD:
@acsoto thankyou for reviewing, pls take a look at this too #389 when you get chance.

avinxshKD:
@RainbowMango PTAL. its been marked as lgtm. can we merge this??

PR #400, opened 2026-06-22
Feat/picod prometheus metrics

vanshika2720:
/kind feature

**What type of PR is this?**

/kind feature

PicoD currently exposes only a `/health` endpoint and does not provide any Prometheus-compatible metrics. This makes it difficult to monitor sandbox activity, understand request patterns, or observe command execution behavior.

This PR adds a Prometheus `/metrics` endpoint to PicoD using the existing `prometheus/client_golang` dependency already present in the module graph. The implementation introduces a dedicated Prometheus registry and exports metrics focused on request handling and command execution.

Metrics added:

* `picod_active_executions` – Gauge tracking in-flight execute requests.
* `picod_execute_requests_total` – Counter tracking execute request outcomes (`success`, `error`, `timeout`, `invalid`).
* `picod_http_requests_total` – Counter tracking processed HTTP requests by method, path, and status code.
* `picod_http_request_duration_seconds` – Histogram tracking request latency by method and path.

The `/metrics` endpoint is exposed similarly to `/health` and is excluded from gzip compression. The implementation also includes middleware-based request instrumentation and unit tests validating metric registration and collection behavior.

Part of #386 

* Uses a dedicated Prometheus registry instead of the global registry to avoid exposing default Go runtime metrics.
* No new dependencies are introduced.
* Metrics focus on PicoD's stateless execution model and avoid introducing artificial session concepts.
* Added programmatic metric validation tests using `Registry.Gather()` rather than relying on text exposition output.

```release-note
PicoD now exposes a Prometheus-compatible `/metrics` endpoint that provides request, latency, and command execution metrics for observability and monitoring.
```

### DIFF
--- a/pkg/picod/metrics.go
+++ b/pkg/picod/metrics.go
@@ -17,6 +17,7 @@ limitations under the License.
 package picod
 
 import (
+	"net/http"
 	"strconv"
 	"time"
 
@@ -24,6 +25,29 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
+
+const methodUnknown = "unknown"
+
+// requestDurationBuckets covers PicoD's full latency range: sub-millisecond
+// for fast file and health-style paths through the 60-second default command
+// timeout used by ExecuteHandler.
+var requestDurationBuckets = []float64{
+	0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5,
+	1, 2.5, 5, 10, 20, 30, 45, 60,
+}
+
+// normalizeMethod maps non-standard HTTP methods to "unknown" to prevent
+// unbounded label cardinality in Prometheus time series.
+func normalizeMethod(method string) string {
+	switch method {
+	case http.MethodGet, http.MethodPost, http.MethodPut, http.MethodDelete,
+		http.MethodPatch, http.MethodHead, http.MethodOptions,
+		http.MethodConnect, http.MethodTrace:
+		return method
+	default:
+		return methodUnknown
+	}
+}
 
 // Metrics holds the Prometheus collectors for PicoD.
 type Metrics struct {
@@ -63,7 +87,7 @@ func NewMetrics() *Metrics {
 		prometheus.HistogramOpts{
 			Name:    "picod_http_request_duration_seconds",
 			Help:    "Latency of HTTP requests, partitioned by method and path.",
-			Buckets: prometheus.DefBuckets,
+			Buckets: requestDurationBuckets,
 		},
 		[]string{"method", "path"},
 	)
@@ -108,7 +132,8 @@ func (m *Metrics) Middleware() gin.HandlerFunc {
 		duration := time.Since(start).Seconds()
 
 		status := strconv.Itoa(c.Writer.Status())
-		m.HTTPRequestsTotal.WithLabelValues(c.Request.Method, path, status).Inc()
-		m.HTTPRequestDuration.WithLabelValues(c.Request.Method, path).Observe(duration)
+		method := normalizeMethod(c.Request.Method)
+		m.HTTPRequestsTotal.WithLabelValues(method, path, status).Inc()
+		m.HTTPRequestDuration.WithLabelValues(method, path).Observe(duration)
 	}
 }
--- a/pkg/picod/metrics.go
+++ b/pkg/picod/metrics.go
@@ -1,0 +1,114 @@
+/*
+Copyright The Volcano Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package picod
+
+import (
+	"strconv"
+	"time"
+
+	"github.com/gin-gonic/gin"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+// Metrics holds the Prometheus collectors for PicoD.
+type Metrics struct {
+	Registry             *prometheus.Registry
+	ActiveExecutions     prometheus.Gauge
+	ExecuteRequestsTotal *prometheus.CounterVec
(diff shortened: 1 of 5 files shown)


ranxi2001:
I did a local pass over the current head (`3e4ac99`) and the PicoD package tests pass for me:

- `go test ./pkg/picod -count=1`
- `go test -race ./pkg/picod -count=1`
- `go test ./pkg/picod -run TestMetrics_Exposition -count=20`

One issue still looks worth fixing before merge:

`s.metrics.Middleware()` is currently registered after `maxBodySizeMiddleware()` in `pkg/picod/server.go`. Requests rejected by the body-size middleware return `413` and abort before the metrics middleware runs, so they are missing from `picod_http_requests_total` and `picod_http_request_duration_seconds`.

I verified this locally with a temporary test that sends an oversized `POST /api/execute` request. The response is `413`, but there is no `picod_http_requests_total{method="POST", path="/api/execute", status_code="413"}` sample. Moving the metrics middleware before `maxBodySizeMiddleware()` makes the same test pass.

Suggested focused test coverage:

1. Send an oversized `POST /api/execute`.
2. Assert the response status is `413`.
3. Assert `picod_http_requests_total` includes `method="POST", path="/api/execute", status_code="413"`.

Two smaller follow-ups:

- `picod_active_executions` currently counts the whole execute handler, including JSON parsing and validation. If the intended metric is active command executions, the gauge should wrap only the `cmd.Run()` window; otherwise the name/help text should make it clear that this is active execute handler requests.
- It would be useful to add a regression test for unmatched routes to ensure the path label stays fixed as `unmatched` and never uses the raw URL path.

vanshika2720:
@ranxi2001 Thanks for the detailed review! I addressed the suggested changes:

Moved s.metrics.Middleware() before maxBodySizeMiddleware() so 413 responses are also captured by the HTTP metrics.
Added a regression test for oversized POST /api/execute requests to verify the 413 metrics are recorded.
Added a regression test to ensure unmatched routes always use the path="unmatched" label.

vanshika2720:
@hzxuzhonghu  @RainbowMango  Please review it.

pkg/picod/server.go
```diff
@@ -79,9 +81,10 @@ func NewServer(config Config) *Server {
 	// Global middleware
 	engine.Use(gin.Logger())   // Request logging
 	engine.Use(gin.Recovery()) // Crash recovery
+	engine.Use(s.metrics.Middleware())
```
acsoto, line 83: Metrics should wrap `gin.Recovery()` as well. With the current order, a panic recovered by `Recovery()` unwinds past this middleware before the post-`c.Next()` recording runs, so recovered 500s are not counted.

acsoto:
The PR body says `Fixes #386`, but #386 is the v0.2.0 umbrella tracking issue. Please change this to `Part of #386` or link a dedicated metrics issue so merging this PR does not auto-close the umbrella.

pkg/picod/server.go
vanshika2720: @acsoto resolved!

vanshika2720:
> The PR body says `Fixes #386`, but #386 is the v0.2.0 umbrella tracking issue. Please change this to `Part of #386` or link a dedicated metrics issue so merging this PR does not auto-close the umbrella.

@acsoto Done!

ranxi2001 (commented):
I re-reviewed the current head (`0d4576f`). The earlier 413, unmatched-path, and recovered-panic issues are fixed. I found two remaining observability concerns inline. Codex assisted the source audit; I verified the behaviors with focused local tests.

pkg/picod/metrics.go
```diff
@@ -0,0 +1,114 @@
...
+			return
+		}
+
+		start := time.Now()
+		c.Next()
+		duration := time.Since(start).Seconds()
+
+		status := strconv.Itoa(c.Writer.Status())
+		m.HTTPRequestsTotal.WithLabelValues(c.Request.Method, path, status).Inc()
+		m.HTTPRequestDuration.WithLabelValues(c.Request.Method, path).Observe(duration)
```
ranxi2001: Although `path` is now bounded, `c.Request.Method` is still passed to both metric vectors verbatim. HTTP methods are arbitrary valid tokens, and unmatched routes reach this global middleware without `/api` authentication. On this head, sending 32 distinct custom methods produced 32 counter label sets and 32 histogram label sets; those children remain registered for the process lifetime, so a caller can grow PicoD's metrics memory without a bound.

Please normalize methods through a finite allowlist and collapse everything else to `unknown` (matching `promhttp`'s instrumentation pattern), then add a regression test showing that many custom methods create one fallback series.

pkg/picod/metrics.go
```diff
@@ -0,0 +1,114 @@
...
+			Help: "Total number of HTTP requests processed, partitioned by method, path, and status code.",
+		},
+		[]string{"method", "path", "status_code"},
+	)
+
+	httpRequestDuration := prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "picod_http_request_duration_seconds",
+			Help:    "Latency of HTTP requests, partitioned by method and path.",
+			Buckets: prometheus.DefBuckets,
```
ranxi2001: `prometheus.DefBuckets` has no finite boundary above 10 seconds, but `/api/execute` defaults to a 60-second timeout and accepts longer durations. Consequently, all executions above 10 seconds are indistinguishable in the bucket distribution. [Prometheus documents](https://prometheus.io/docs/prometheus/3.5/querying/functions/#histogram_quantile) that `histogram_quantile()` returns the second-highest boundary when a classic-histogram quantile falls in the `+Inf` bucket, so a 30-60 second tail can be reported as 10 seconds.

Could we choose buckets from PicoD's execution SLO/timeout range (or expose a separate command-execution duration histogram) and add a test for the intended long-running boundary?

vanshika2720:
@ranxi2001  I've addressed both concerns:

Bounded HTTP method cardinality: Added a normalizeMethod() helper that allowlists the standard HTTP methods and maps everything else to unknown. The middleware now records the normalized method instead of c.Request.Method. I also added regression tests to verify that arbitrary custom methods collapse into a single method="unknown" series and do not create unbounded label values.

Histogram bucket range: Replaced prometheus.DefBuckets with a bucket set extending to 60 seconds, matching PicoD's default execution timeout. I considered introducing a separate execution-duration histogram, but since /api/execute is synchronous, picod_http_request_duration_seconds{path="/api/execute"} already provides the execution latency without adding another metric series. I also updated the tests to verify that the configured buckets are applied correctly.

ranxi2001:
Thanks for addressing the review feedback so quickly. I re-reviewed `b8c4ed5`: the HTTP method labels are now bounded, the histogram buckets cover PicoD's default execution window, and the focused package and race tests pass. My review is complete.

/lgtm

vanshika2720:
@RainbowMango
@JesseStutler
@hajnalmt

PR #407, opened 2026-06-27
fix(workloadmanager): refresh user client when token changes

### LINKED ISSUE #406: workloadmanager reuses cached user k8s client after service account token changes
WorkloadManager caches user Kubernetes clients by namespace + service account name, but the cache lookup ignores the current bearer token.

So if the same service account comes in with a different token, GetOrCreateUserK8sClient still returns the old cached client. That client was built with the old BearerToken.

This is bad for token rotation/replacement. The request has already authenticated with the new token, but the actual Kubernetes operation may still use the old one.

Steps to reproduce:
1. Create a K8sClient with a ClientCache and base rest.Config.
2. Call GetOrCreateUserK8sClient("first-token", "default", "runner").
3. Call GetOrCreateUserK8sClient("second-token", "default", "runner").
4. Compare the returned clients.

Actual:
Both calls return the same cached UserK8sClient.

Expected:
A token change should create/update the cached client, or the cache should include/check the token before reuse.

Files:
- pkg/workloadmanager/client_cache.go
- pkg/workloadmanager/k8s_client.go

Probably fix is to store the token or token hash in ClientCache entries and make Get check it before returning the cached client. Add tests for same SA + changed token.

avinxshKD:
**What type of PR is this?**

/kind bug

WorkloadManager cached user Kubernetes clients by namespace and service account only. If the same service account came in with a different bearer token, it could still reuse the client built with the old token.

This makes the cache check the token too, so token changes create a fresh user client. Also adds tests for same-token reuse and changed-token refresh.

Fixes #406

NONE

```release-note
NONE

### DIFF
--- a/pkg/workloadmanager/client_cache.go
+++ b/pkg/workloadmanager/client_cache.go
@@ -30,6 +30,7 @@ import (
 type clientCacheEntry struct {
 	key         string
 	client      *UserK8sClient
+	token       string
 	tokenExpiry time.Time // Token expiration time parsed from JWT
 	element     *list.Element
 }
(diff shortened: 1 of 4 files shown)


avinxshKD:
@LiZhenCheng9527 @YaoZengzeng Pls take a look when get chance

avinxshKD:
@acsoto ptal when get chance

avinxshKD:
@RainbowMango PTAL. its been marked as lgtm. can we merge this??

PR #410, opened 2026-06-28
Fix : Remove default maxSessionDuration for AgentRuntime

### LINKED ISSUE #303: Remove AgentRuntime default max ttl

safiya2610:
**What type of PR is this?**

This PR resolves Issue #303  by making AgentRuntime.spec.maxSessionDuration optional instead of defaulting to 8 hours.

## Changes
Removed the default maxSessionDuration value for AgentRuntime.
Updated the workload builder to set Sandbox.Spec.Lifecycle.ShutdownTime only when maxSessionDuration is explicitly configured.
Preserved the existing default maximum session duration behavior for CodeInterpreter.
Updated the AgentRuntime CRD to reflect the optional field.
Added and updated unit tests to cover the new behavior.
Why
Previously, every AgentRuntime was assigned a default maximum session duration of 8 hours, causing long-running agents to be terminated even when they were still active. With this change, long-running AgentRuntime instances are no longer force-terminated unless a maximum session duration is explicitly specified, while CodeInterpreter continues to retain its existing behavior.

Testing
go test ./pkg/workloadmanager/...
go test ./pkg/apis/runtime/v1alpha1/...
go build ./...

Fixes #303

### DIFF
--- a/pkg/apis/runtime/v1alpha1/agent_type.go
+++ b/pkg/apis/runtime/v1alpha1/agent_type.go
@@ -53,8 +53,8 @@ type AgentRuntimeSpec struct {
 
 	// MaxSessionDuration describes the maximum duration for a session.
 	// After this duration, the session will be terminated no matter active or inactive.
-	// +kubebuilder:validation:Required
-	// +kubebuilder:default="8h"
+	// This field is optional for AgentRuntime; if omitted, sessions are not capped by a hard lifetime.
+	// Non-positive values (e.g., 0s, -1h) are also treated as omitted and will not apply a hard lifetime cap.
 	MaxSessionDuration *metav1.Duration `json:"maxSessionDuration,omitempty" protobuf:"bytes,3,opt,name=maxSessionDuration"`
 }
 
--- a/pkg/workloadmanager/sandbox_helper.go
+++ b/pkg/workloadmanager/sandbox_helper.go
@@ -38,6 +38,11 @@ const (
 	sandboxStatusNotReady = "not-ready"
 )
 
+// noExpiryDeadline represents a sandbox with no hard expiry deadline
+// (used for AgentRuntime without maxSessionDuration).
+// Setting ExpiresAt to this value ensures the sandbox won't be selected by ListExpiredSandboxes queries.
+var noExpiryDeadline = time.Date(9999, 12, 31, 23, 59, 59, 0, time.UTC)
+
 var sandboxEntrypointDial = func(ctx context.Context, endpoint string, timeout time.Duration) error {
 	dialer := &net.Dialer{Timeout: timeout}
 	conn, err := dialer.DialContext(ctx, "tcp", endpoint)
@@ -52,7 +57,9 @@ func buildSandboxPlaceHolder(sandboxCR *sandboxv1alpha1.Sandbox, entry *sandboxE
 	if sandboxCR.Spec.Lifecycle.ShutdownTime != nil {
 		expiresAt = sandboxCR.Spec.Lifecycle.ShutdownTime.Time
 	} else {
-		expiresAt = time.Now().Add(DefaultSandboxTTL)
+		// If no ShutdownTime is set (AgentRuntime without maxSessionDuration),
+		// use noExpiryDeadline to indicate no hard expiry deadline.
+		expiresAt = noExpiryDeadline
 	}
 	idleTimeout := entry.IdleTimeout
 	if idleTimeout == 0 {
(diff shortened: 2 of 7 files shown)


pkg/workloadmanager/workload_builder.go
```diff
@@ -154,14 +154,15 @@ type buildSandboxClaimParams struct {
...
-	if params.ttl == 0 {
-		params.ttl = DefaultSandboxTTL
-	}
 	if params.idleTimeout == 0 {
 		params.idleTimeout = DefaultSandboxIdleTimeout
 	}
 
-	shutdownTime := metav1.NewTime(time.Now().Add(params.ttl))
+	var shutdownTime *metav1.Time
+	if params.ttl > 0 {
```
acsoto: This should probably treat any non-zero TTL as configured. With `ttl < 0`, the old behavior produced an already-expired shutdown time, but this now silently removes the hard deadline.

acsoto:
This matches the direction discussed in #303: AgentRuntime should not get an implicit 8h hard lifetime, while CodeInterpreter can keep the existing default. Please run `make gen`/`make gen-check` and include the generated changes; the current Codegen Check is failing because the generated AgentRuntime CRD is missing the new `maxSessionDuration` description line, and `go mod tidy` also changes `go.sum`.

pkg/workloadmanager/workload_builder.go
safiya2610: sure, I will make changes.

safiya2610: solve

safiya2610:
> This matches the direction discussed in #303: AgentRuntime should not get an implicit 8h hard lifetime, while CodeInterpreter can keep the existing default. Please run `make gen`/`make gen-check` and include the generated changes; the current Codegen Check is failing because the generated AgentRuntime CRD is missing the new `maxSessionDuration` description line, and `go mod tidy` also changes `go.sum`.

@acsoto Hello, Please review the PR, everything is done.

acsoto:
One small follow-up: `NoExpiryDeadline` looks like an internal sentinel and is currently an exported mutable variable. Please make it unexported (for example `noExpiryDeadline`) so other packages cannot accidentally change the GC/store expiry behavior.

safiya2610:
> One small follow-up: `NoExpiryDeadline` looks like an internal sentinel and is currently an exported mutable variable. Please make it unexported (for example `noExpiryDeadline`) so other packages cannot accidentally change the GC/store expiry behavior.

done, Please check.

pkg/workloadmanager/sandbox_helper.go
```diff
@@ -72,7 +79,7 @@ func buildSandboxPlaceHolder(sandboxCR *sandboxv1alpha1.Sandbox, entry *sandboxE
 
 func buildSandboxInfo(sandbox *sandboxv1alpha1.Sandbox, podIP string, entry *sandboxEntry) *types.SandboxInfo {
 	createdAt := sandbox.GetCreationTimestamp().Time
-	expiresAt := createdAt.Add(DefaultSandboxTTL)
+	expiresAt := noExpiryDeadline
 	if sandbox.Spec.Lifecycle.ShutdownTime != nil {
 		expiresAt = sandbox.Spec.Lifecycle.ShutdownTime.Time
 	}
```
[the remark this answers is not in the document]
safiya2610: I believe the nil ShutdownTime path is already exercised by the existing tests, so I'm keeping the current test coverage for this PR.

safiya2610:
@acsoto ig, everything is done, Please review.

safiya2610:
@acsoto  Sorry for the ping! I accidentally pushed some extra commits earlier, but I have now reset the branch back to the exact code you approved. Whenever you get a chance, could you please add / lgtm again? Thank you!

safiya2610:
@RainbowMango @JesseStutler @hajnalmt

safiya2610:
> @RainbowMango @JesseStutler @hajnalmt

@RainbowMango Please take a look on this PR, its is already taged lgtm

RainbowMango:
OK. 
Putting it into my queue. Sorry for letting this sit.
/assign

pkg/apis/runtime/v1alpha1/agent_type.go
```diff
@@ -53,8 +53,8 @@ type AgentRuntimeSpec struct {
 
 	// MaxSessionDuration describes the maximum duration for a session.
 	// After this duration, the session will be terminated no matter active or inactive.
-	// +kubebuilder:validation:Required
-	// +kubebuilder:default="8h"
+	// This field is optional for AgentRuntime; if omitted, sessions are not capped by a hard lifetime.
+	// Non-positive values (e.g., 0s, -1h) are also treated as omitted and will not apply a hard lifetime cap.
```
RainbowMango, line 57:
```suggestion
	// The value must be a positive duration (e.g., "30m", "8h").
	// If not specified, sessions are not capped by a maximum lifetime.
```
1. It's redundant to mention that this is for `AgentRuntime`.
2. Non-positive values are not allowed, otherwise brings misleading.

pkg/workloadmanager/workload_builder.go
```diff
@@ -154,14 +154,15 @@ type buildSandboxClaimParams struct {
...
 	if params.idleTimeout == 0 {
 		params.idleTimeout = DefaultSandboxIdleTimeout
 	}
 
-	shutdownTime := metav1.NewTime(time.Now().Add(params.ttl))
+	var shutdownTime *metav1.Time
+	if params.ttl != 0 {
+		st := metav1.NewTime(time.Now().Add(params.ttl))
+		shutdownTime = &st
+	}
```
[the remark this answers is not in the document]
RainbowMango: +1

RainbowMango (changes requested):

PR #412, opened 2026-06-29
feat: add soft node-affinity stickiness for AgentRuntime and CodeInte…

Turbo-Jiaxxx789:
…rpreter

When enabled, a new session prefers (soft / PreferredDuringScheduling) the node the previous session for the same workload landed on. The scheduled node is recorded back onto the workload via a last-node annotation after the pod is ready. Gated by a CRD-level nodeStickiness.enabled switch that defaults to off, so the feature is fully inert unless explicitly turned on. Has no effect for CodeInterpreter warm pools, since those sandboxes are pre-scheduled.

What type of PR is this?
/kind feature

What this PR does / why we need it:
Adds opt-in soft node-affinity stickiness for AgentRuntime and CodeInterpreter. When spec.nodeStickiness.enabled is set, a new session prefers to schedule onto the node where the previous session for the same workload ran, improving warm-cache / local-data reuse across sessions.
**In our RL experiments, enable node-affinity stickiness for AgentRuntime reduced startup time by 70%+ across ~20k containers + ~20k images.**
New optional field spec.nodeStickiness.enabled (bool, default false) on both CRDs.
When enabled and a last node is recorded, a PreferredDuringSchedulingIgnoredDuringExecution node-affinity term (weight 100, key kubernetes.io/hostname) is appended to the sandbox pod spec — user-defined affinity is preserved, never overwritten.
After the session pod becomes ready, the scheduled node is written back onto the owning workload via the runtime.agentcube.io/last-node annotation.
It is a soft preference only, so scheduling never fails if the preferred node is cordoned, deleted, or full.

Which issue(s) this PR fixes: Fixes #

Special notes for your reviewer:
Backward compatible: a nil/omitted nodeStickiness is treated as disabled, so existing CRs need no migration.
Has no effect for CodeInterpreter with warmPoolSize > 0, since warm-pool sandboxes are pre-scheduled before a session claims them.
The last-node write-back is a no-op when stickiness is disabled.
Does this PR introduce a user-facing change?:

Add opt-in soft node-affinity stickiness for AgentRuntime and CodeInterpreter via `spec.nodeStickiness.enabled` (default false). When enabled, sessions prefer the node used by the previous session for the same workload.


pkg/workloadmanager/handlers.go
```diff
@@ -287,6 +287,8 @@ func (s *Server) createSandbox(ctx context.Context, dynamicClient dynamic.Interf
 		return nil, api.NewInternalError(fmt.Errorf("failed to verify sandbox %s/%s entrypoints: %w", sandbox.Namespace, sandbox.Name, err))
 	}
 
+	s.recordStickyNode(sandbox.Namespace, sandboxEntry, nodeName)
```
acsoto: Please move this write-back until after `UpdateSandbox` succeeds. If the store update fails, the rollback path can delete the sandbox while this goroutine has already recorded a node for a session that was not actually established.

pkg/workloadmanager/k8s_client.go
```diff
@@ -310,20 +320,20 @@ func (u *UserK8sClient) DeleteSandboxClaim(ctx context.Context, namespace, sandb
 	return deleteSandboxClaim(ctx, u.dynamicClient, namespace, sandboxClaimName)
 }
 
-// GetSandboxPodIP gets the IP address of the pod corresponding to the Sandbox
-func (c *K8sClient) GetSandboxPodIP(_ context.Context, namespace, sandboxName, podName string) (string, error) {
+// GetSandboxPodIP gets the IP address and scheduled node of the pod corresponding to the Sandbox
+func (c *K8sClient) GetSandboxPodIP(_ context.Context, namespace, sandboxName, podName string) (podIP string, nodeName string, err error) {
```
acsoto: Nit: this now returns both the pod IP and scheduled node, so `GetSandboxPodIP` is a bit misleading. `GetSandboxPodInfo` would better match the return values.

pkg/workloadmanager/handlers.go
Turbo-Jiaxxx789: good suggestion, fixed now.

pkg/workloadmanager/k8s_client.go
Turbo-Jiaxxx789: fixed

Turbo-Jiaxxx789:
closed by mistake operation

Turbo-Jiaxxx789:
> /lgtm

rebase main

Turbo-Jiaxxx789:
> /lgtm

@acsoto review again please，just rebase on the newest main branch after lgtm label

pkg/workloadmanager/workload_builder.go
```diff
@@ -305,9 +313,78 @@ func buildSandboxByAgentRuntime(namespace string, name string, ownerID string, i
...
+// already defined on the pod spec. It is a no-op when nodeName is empty or an
+// equivalent hostname preference already exists, to avoid duplicate terms that
+// would amplify scheduler weighting and bloat the pod spec.
+func injectPreferredNodeAffinity(podSpec *corev1.PodSpec, nodeName string) {
+	if nodeName == "" {
+		return
+	}
+	if podSpec.Affinity != nil && podSpec.Affinity.NodeAffinity != nil {
+		for _, existing := range podSpec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution {
+			if isEquivalentHostnamePreference(existing.Preference, nodeName) {
```
acsoto: This de-duplication ignores the existing term's weight. A user-defined hostname preference with weight 1 suppresses the intended weight-100 stickiness, so enabling this feature may have almost no effect.

pkg/workloadmanager/handlers.go
```diff
@@ -464,11 +464,50 @@ func (s *Server) createSandbox(ctx context.Context, dynamicClient dynamic.Interf
...
+		return
+	}
+	s.wg.Add(1)
+	s.shutdownMu.RUnlock()
+
+	go func() {
+		defer s.wg.Done()
+		patchCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		if err := s.k8sClient.PatchWorkloadLastNode(patchCtx, namespace, kind, name, nodeName); err != nil {
```
acsoto, line 511: Concurrent sessions can complete these detached patches out of order, so an older session may overwrite the node recorded by a newer one. Is last-writer-by-patch-completion the intended stickiness semantics?

Turbo-Jiaxxx789: > Concurrent sessions can complete these detached patches out of order, so an older session may overwrite the node recorded by a newer one. Is last-writer-by-patch-completion the intended stickiness semantics?

Stickiness is a soft hint: concurrent sessions may complete their patches in any order (last writer wins), and the annotation simply records whichever node the most recently completed session landed on. This is intentional — strict ordering would add complexity disproportionate to the value of a soft affinity preference that the scheduler may ignore anyway.

pkg/workloadmanager/workload_builder.go
Turbo-Jiaxxx789: > This de-duplication ignores the existing term's weight. A user-defined hostname preference with weight 1 suppresses the intended weight-100 stickiness, so enabling this feature may have almost no effect.

Good catch — fixed. injectPreferredNodeAffinity now upgrades the weight of an existing equivalent hostname preference to at least 100 instead of silently returning. If the user already has hostname In [node-d] with weight 1, the weight is bumped to 100 so stickiness still takes effect. If the existing weight is already ≥ 100, it is preserved as-is.

acsoto: This still rewrites the user-defined weight instead of preserving it. If that term has weight 50 and another preference has weight 100, upgrading it to 100 leaves a tie, while appending stickiness would score the sticky node at 150.

pkg/workloadmanager/handlers.go
acsoto: ok

pkg/workloadmanager/workload_builder.go
Turbo-Jiaxxx789: > This still rewrites the user-defined weight instead of preserving it. If that term has weight 50 and another preference has weight 100, upgrading it to 100 leaves a tie, while appending stickiness would score the sticky node at 150.

Good point, fixed it — you're right that upgrading puts us at a tie (50→100 vs 100) while appending gives us a clear win (50+100=150). Fixed: the term is now always appended instead of upgrading the existing weight.

The scheduler sums the weights of all matching PreferredDuringScheduling terms, so appending weight-100 on top of a user-defined weight-n term gives the sticky node a total score of 100+n, making it a stronger preference than either term alone. The isEquivalentHostnamePreference helper and its de-duplication logic have been removed entirely since they're no longer needed — the stickiness term is always appended unconditionally.

pkg/workloadmanager/workload_builder.go
```diff
@@ -305,9 +313,83 @@ func buildSandboxByAgentRuntime(namespace string, name string, ownerID string, i
...
+// scoring semantics.
+func injectPreferredNodeAffinity(podSpec *corev1.PodSpec, nodeName string) {
+	if nodeName == "" {
+		return
+	}
+	if podSpec.Affinity != nil && podSpec.Affinity.NodeAffinity != nil {
+		for i := range podSpec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution {
+			existing := &podSpec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution[i]
+			if isEquivalentHostnamePreference(existing.Preference, nodeName) {
+				existing.Weight += 100
```
acsoto: `PreferredSchedulingTerm.Weight` must remain in the 1–100 range, so every equivalent existing term becomes invalid here (for example, 1 becomes 101) and the sandbox pod will be rejected. Keep the user term unchanged and append a separate weight-100 stickiness term as described in the thread.

pkg/workloadmanager/workload_builder.go
acsoto: 🤔

pkg/workloadmanager/workload_builder.go
Turbo-Jiaxxx789: > `PreferredSchedulingTerm.Weight` must remain in the 1–100 range, so every equivalent existing term becomes invalid here (for example, 1 becomes 101) and the sandbox pod will be rejected. Keep the user term unchanged and append a separate weight-100 stickiness term as described in the thread.

thanks for the thorough review. After several iterations, the final approach is: always append a separate weight-100 term to void the 1–100 constraint of per-term (API validation),which I missed before.

acsoto:
overall nice feature. opt-in scheduling improvement for RL workloads
cc @RainbowMango

PR #413, opened 2026-06-29
Use sandbox pod annotation for GetSandboxPodIP

### LINKED ISSUE #277: Simplify sandbox pod match

safiya2610:
## Summary

This change makes GetSandboxPodIP prefer an explicit pod name. When no pod name is provided, it falls back to reading the SandboxPodNameAnnotation from the Sandbox resource instead of discovering the pod through labels or OwnerReferences.

## Changes
- Prefer explicit pod name when available
- Read SandboxPodNameAnnotation from the Sandbox resource when pod name is not provided
- Remove label/OwnerReference-based pod lookup
- Update unit tests to use explicit pod names
- Update tests to expect NotFound when no dynamic client/annotation is available

Fixes #277


safiya2610:
@hzxuzhonghu Hello, Please review this PR, this PR Simplify sandbox pod match https://github.com/volcano-sh/agentcube/issues/277 .

acsoto:
The `agents.x-k8s.io/pod-name` annotation is being removed upstream ([kubernetes-sigs/agent-sandbox#1297](https://github.com/kubernetes-sigs/agent-sandbox/issues/1297)). Please don’t make AgentCube depend on it. We can simply use the Sandbox name to look up the Pod.

safiya2610:
> The `agents.x-k8s.io/pod-name` annotation is being removed upstream ([kubernetes-sigs/agent-sandbox#1297](https://github.com/kubernetes-sigs/agent-sandbox/issues/1297)). Please don’t make AgentCube depend on it. We can simply use the Sandbox name to look up the Pod.

sure.

Issue #430, opened 2026-07-08
[Discussion] Evolving architecture: decouple agent provisioning from the Kubernetes control plane


RainbowMango:
## Background

Today, AgentCube is built entirely around the Kubernetes control plane: every agent / code-interpreter session is backed by Kubernetes resources (sandbox CRs and Pods) that are created on demand by the Workload Manager. This gives us good governance, RBAC, and ecosystem integration — but it also means every session creation must go through the Kubernetes API server and scheduler.

As we move toward large-scale, parallel agent workloads (hundreds or thousands of agents starting in a short burst), this design shows some real limits.

## Problems with the current architecture

1. **Session creation is slow because it goes through Kubernetes scheduling.**
   Creating a sandbox involves many steps: CR creation → reconcile → Pod creation → scheduling → kubelet sync → image pull → microVM boot. Each step adds delay. Taking a couple of seconds to start a Pod is fine for long-running services, but for an interactive agent session that may only live for seconds, this delay is the main part of the user-visible latency. Warm pools help, but do not solve the problem: refilling the pool still goes through the full Kubernetes provisioning path, so a large burst that empties the pool falls back to cold-start latency.

2. **The API server is not designed for a huge number of short-lived objects.**
   Agent sessions are bursty and short-lived. If every session maps to Kubernetes objects, we generate a large volume of create/update/delete writes against etcd. At the scale of massive parallel agent creation, this becomes a bottleneck and can even hurt the stability of the whole cluster.

3. **Idle sandboxes waste resources.**
   Agents spend most of their time idle, waiting for input. Keeping one Pod per idle session wastes CPU and memory and limits how many agents a cluster can hold. Kubernetes has no native suspend/resume for Pods, and even if it did, going through the API server for every wake-up would be too slow.

4. **Capacity planning is mixed up with session lifecycle.**
   Today, "how much capacity is reserved for agents" and "how many agent sessions exist right now" are the same thing. Administrators have no way to plan and reserve capacity for agents ahead of time, independently of live sessions.

## Proposed direction

Split the responsibilities into two layers:

- **Kubernetes manages resources, not sessions.**
  Administrators plan one or more **resource pools** ahead of time (e.g., a new CRD) dedicated to running agents. Kubernetes does what it is good at: provisioning nodes, scheduling long-running worker Pods (sandbox hosts), scaling the pool, and applying the usual RBAC / quota / policy to the pool definition. 

[...]

PR #431, opened 2026-07-09
[Proposal] add sandbox-pool management proposal

lichuqiang:
**What type of PR is this?**

/kind feature

As discussed in #430, the current agentcube architecture has limitations in certain scenarios. To break through the
bottlenecks of the Kubernetes architectural paradigm in workload startup speed and managed resource scale, we envision an            architectural evolution of agentcube that establishes a dual fast/slow resource track. The slow track builds sandbox resource pool
management capabilities on top of the Kubernetes CRD framework, supporting the partitioning of resources on K8s nodes as resource pools for sandbox execution. In parallel, a lightweight, fast, Kubernetes-independent sandbox lifecycle management platform is
established to achieve more extreme performance and a larger management scale.

This proposal focuses on the slow track. The system implements declarative management of sandbox resources through a two-layer architecture:

- Global Policy Layer (SandboxPool Controller): Responsible for Class → Pool mapping, node selection, policy snapshot
synchronization, and Phase aggregation.
- Node Execution Layer (placeholder-agent): Acts as the node-local agent, responsible for resource policy enforcement, watermark
checking, and status reporting.

Core capabilities include:

- Providing declarative CRD APIs (SandboxPoolClass + SandboxPool) to manage node-level sandbox resource pools
- Reliably locking scheduled resources through the Static Pod model, with automatic mirror pod rebuild in < 5s upon accidental
deletion
- Supporting  resource specification adjustment by recreating placeholder pod

Integrated with the fast-track data plane, this ultimately enables co-location of regular Pods and sandboxes on the same node.

Refs #430

Architecture evolution

### DIFF
--- a/docs/proposals/sandbox-pool-management/README.md
+++ b/docs/proposals/sandbox-pool-management/README.md
@@ -41,7 +41,7 @@ Core designs include:
 
 - Using the **Static Pod model** to lock scheduling resources; mirror pods are automatically rebuilt after accidental deletion.
 - Declarative state synchronization based on **CRD Watch/List**.
-- Implements vertical resource scaling via **[containerd runtime shim (v2 Task API)](https://github.com/containerd/containerd/blob/main/api/runtime/task/v2/README.md)** (VPA analogy), independent of InPlacePodVerticalScaling Feature Gate.
+- Implements vertical resource scaling via **[containerd runtime shim (v2 Task API)](https://github.com/containerd/containerd/blob/main/docs/runtime-v2.md)** (VPA analogy), independent of InPlacePodVerticalScaling Feature Gate.
 - Coordinating multi-component status writes via the **SSA (server-side apply) field manager mechanism** to avoid field-level conflicts.
 
 ## Motivation
@@ -126,7 +126,7 @@ The system adopts a two-layer architecture model:
 | Storage backend | Pure etcd (K8s API Server) | Resource pool control plane only needs CRD state; no external storage needed |
 | CRD Scope | Cluster-scoped | placeholder-agent is a node-local singleton (one per node) bound to one Pool; no namespace isolation needed |
 | Phase calculator | Controller aggregation | When placeholder-agent is unreachable (node deleted or agent crash), it cannot update Phase; Controller aggregation + stale detection ensures Phase can still be updated in extreme scenarios |
-| Status write coordination | SSA Apply (fieldManager) | placeholder-agent SSA-Applies status every 30s; Controller occasionally SSA-Applies; field-level ownership avoids conflicts |
+| Status write coordination | SSA Apply (fieldManager) | placeholder-agent SSA-Applies status on event (only on status field changes); heartbeat via Lease every 10s; Controller occasionally SSA-Applies; field-level ownership avoids conflicts |
 | Resource adjustment method | containerd runtime shim (VPA analogy) | As a containerd shim v2 plugin, intercepts container creation calls to adjust resources online; watermark protection during downscale prevents running sandbox OOM |
 
 > **Manifest Self-Healing**: The Static Pod model protects against accidental **Pod** deletion (kubelet rebuilds from manifest). But if the **manifest file itself** is accidentally deleted, kubelet will stop the static pod. To cover this case, the placeholder-agent daemon's reconcile loop periodically verifies that the manifest exists for every active Pool on its node; if the manifest is missing while the Pool CRD still exists, the daemon rewrites it. This makes manifest recovery as automatic as mirror-pod recovery — the agent is the authoritative writer of the manifest, and the CRD (not the manifest) is the source of truth for desired state.
@@ -166,12 +166,12 @@ SandboxPool status is managed by two components via the SSA (Server-Side Apply) 
 
 | Field Manager | Component | Managed Fields | Write Timing |
 |-----------|-----------|---------------|--------------|
-| `placeholder-agent` | placeholder-agent | placeholderPod, placeholderAgent, override, resize, nodeCtl, poolInfo, conditions (non-NodeNotFound/PlaceholderAgentHealthy), lastAppliedGeneration | Periodic SSA Apply every 30s |
+| `placeholder-agent` | placeholder-agent | placeholderPod, placeholderAgent, override, resize, nodeCtl, poolInfo, conditions (non-NodeNotFound/PlaceholderAgentHealthy), lastAppliedGeneration | Event-driven SSA Apply (only on status field changes); heartbeat via Lease every 10s |
 | `sandboxpool-controller` | SandboxPool Controller | **phase**, conditions (NodeNotFound, PlaceholderAgentHealthy) | SSA Apply during Pool Reconcile |
 
-This "data source and aggregator separated" design ensures Phase can still be updated in extreme scenarios — when a node is deleted, the Controller sets Phase=Unready based on the NodeNotFound Condition; when the agent crashes but the Node still exists, the Controller detects stale `PlaceholderAgent.LastHeartbeat` (> 2min) and sets `PlaceholderAgentHealthy=False`, downgrading Phase to Degraded/Unready.
+This "data source and aggregator separated" design ensures Phase can still be updated in extreme scenarios — when a node is deleted, the Controller sets Phase=Unready based on the NodeNotFound Condition; when the agent crashes but the Node still exists, the Controller detects the expired Lease (TTL > 40s with no renewal) and sets `PlaceholderAgentHealthy=False`, downgrading Phase to Degraded/Unready.
 
-> **Heartbeat Timeout Scheduling**: After the agent crashes, no further Pool/Node events are generated, so the Controller's Watch would not trigger a new Reconcile. To cover this scenario, the Controller computes `RequeueAfter = lastHeartbeat + 2min - now` (if the heartbeat has not yet expired) or requeues immediately (if already expired) at the end of each Reconcile, ensuring that even without external events, the heartbeat timeout still triggers a Reconcile to update `PlaceholderAgentHealthy`. Additionally, the Controller performs a full Pool sweep every 30s as a fallback, preventing individual Pool requeues from being lost due to leader election switches or restarts.
+> **Lease-Based Heartbeat**: Each placeholder-agent maintains a `coordination.k8s.io/Lease` object (name = Pool name, cluster-scoped) and renews it every 10s. The Controller Watches Leases; when a Lease's `renewTime` exceeds its TTL (default 40s = 4× renewal interval), the Controller sets `PlaceholderAgentHealthy=False` and downgrades Phase. This mirrors the kubelet heartbeat pattern: the Lease is a high-frequency small object (just renewTime + holderIdentity), while the status SSA Apply is event-driven — only triggered when a status field actually changes. In a 10,000-node cluster, steady-state load is ~1,000 Lease renewals/s (lightweight) with near-zero status Apply requests, versus ~333 status Apply/s if status were used for both heartbeat and state. The Lease renewal interval and TTL are configurable via `--lease-renew-interval` (default 10s) and `--lease-ttl` (default 40s).
 
 > **Conditions SSA note**: The `conditions` field must be declared as a map list in the CRD schema (`x-kubernetes-list-type: map`, `x-kubernetes-list-map-keys: [type]`) so that different field managers can own and update different condition entries without replacing the entire list.
 
@@ -187,30 +187,35 @@ This "data source and aggregator separated" design ensures Phase can still be up
 
 ```go
 type SandboxPoolClassSpec struct {
-    // Target node selector; immutable after creation
+    // Target node selector; immutable after creation.
+    //
+    // Why immutable: the Selector determines the set of nodes a Class applies to.
+    // Changing it would cause large-scale churn — nodes that no longer match would need
+    // their Pools garbage-collected, and newly-matching nodes would need new Pools
+    // created, resulting in a mass rebuild behavior. This mirrors how
+    // Deployment.spec.selector is immutable in Kubernetes for the same reason.
+    // No actual use case requiring mutation has been identified; the restriction can
+    // be lifted in the future if a concrete requirement emerges.
     Selector metav1.LabelSelector `json:"selector"`
 
-    // Additional node selection criteria
-    NodeSelector map[string]string `json:"nodeSelector,omitempty"`
-
-    // Per-node resource quota
+    // Per-node resource quota. Required; resources must be non-empty and include
+    // at least `cpu` and `memory` keys. Other resource keys (e.g. hugepages,
+    // extended resources) are optional.
     ResourcePolicy ResourcePolicy `json:"resourcePolicy"`
 
-    // Placeholder Pod template
-    PlaceholderPodTemplate *PlaceholderPodTemplateSpec `json:"placeholderPodTemplate,omitempty"`
-
-    // node-ctl endpoint (informational only; placeholder-agent obtains the actual address
-    // via --node-ctl-socket startup parameter, not from this field)
-    // Reserved for future declarative reconciliation; currently only unix:// supported
-    NodeCtlEndpoint string `json:"nodeCtlEndpoint,omitempty"`
+    // Placeholder Pod configuration
+    PlaceholderPod *PlaceholderPodSpec `json:"placeholderPod,omitempty"`
 }
 
+// ResourcePolicy wraps a corev1.ResourceList for per-node resource quota definition.
+// Using ResourceList (map[corev1.ResourceName]resource.Quantity) instead of hardcoded
+// CPU/Memory fields preserves room for future extension — e.g. percentage-based specs
+// or additional resource types — without API changes.
 type ResourcePolicy struct {
-    CPU    resource.Quantity `json:"cpu"`
-    Memory resource.Quantity `json:"memory"`
+    Resources corev1.ResourceList `json:"resources,omitempty"`
 }
 
-type PlaceholderPodTemplateSpec struct {
+type PlaceholderPodSpec struct {
     Annotations     map[string]string `json:"annotations,omitempty"`
     Tolerations     []corev1.Toleration `json:"tolerations,omitempty"`
     RuntimeClassName *string           `json:"runtimeClassName,omitempty"`
@@ -236,11 +241,6 @@ type SandboxPoolClassStatus struct {
 }
 ```
 
-**Validation Rules**:
-- `spec.selector` is immutable after creation (Webhook validation)
-- `spec.resourcePolicy.cpu` / `spec.resourcePolicy.memory` are required
-- `spec.nodeCtlEndpoint` (if non-empty) must start with `unix://`. `http://` / `https://` are reserved for future extension; the current version webhook rejects non-`unix://` values
-
 #### SandboxPool
 
 **Group / Version / Kind / Scope**: `sandboxpool.agentcube.volcano.sh` / `v1alpha1` / `SandboxPool` / `Cluster`
@@ -249,31 +249,55 @@ type SandboxPoolClassStatus struct {
 
 ```go
 type SandboxPoolSpec struct {
+    // Reference to the parent SandboxPoolClass. Set by the Controller during Pool creation.
+    // Immutable after creation — changing it would detach the Pool from its Class.
     ClassRef ClassRef `json:"classRef"`
 
+    // Name of the node this Pool is bound to. Set by the Controller during Pool creation.
+    // Immutable after creation — changing it would break the 1-Pool-per-node binding.
+    // A single node cannot have Pools of different Classes — enforced by the Validating
+    // Webhook and four-layer protection (see Validating Webhook section).
     NodeName string `json:"nodeName"`
 
-    // Declarative placeholder; placeholder-agent obtains the actual address via --node-ctl-socket
+    // Node-ctl configuration snapshot. Set by the Controller from Class spec.
+    // placeholder-agent obtains the actual socket address via --node-ctl-socket
+    // startup parameter; this field is informational and for future declarative use.
+    // Optional; when nil, the agent uses its startup flag value.
     NodeCtl *NodeCtlConfig `json:"nodeCtl,omitempty"`
 
-    // Policy snapshot: synced by Controller from Class.spec.resourcePolicy
-    // When Override is enabled, placeholder-agent prioritizes spec.override.resourcePolicy
+    // Resource policy snapshot synced by the Controller from Class.spec.resourcePolicy.
+    // When Override.enabled=true, placeholder-agent prioritizes spec.override.resourcePolicy.
+    // Optional in spec but always set by the Controller during creation.
     ResourcePolicy *ResourcePolicy `json:"resourcePolicy,omitempty"`
 
+    // Operator override for node-specific adjustments. Optional.
+    // When enabled, the agent uses the override's resourcePolicy instead of the Class snapshot.
     Override *OverrideSpec `json:"override,omitempty"`
 }
 
 type OverrideSpec struct {
-    Enabled        bool            `json:"enabled,omitempty"`
-    Reason         string          `json:"reason,omitempty"`
+    // Whether the override is active. When true, the agent uses the override's
+    // resourcePolicy instead of the Class snapshot. Set by operators.
+    Enabled bool `json:"enabled,omitempty"`
+
+    // Human-readable reason for the override. Required when Enabled=true.
+    // Provides audit traceability for why the override was applied.
+    Reason string `json:"reason,omitempty"`
+
+    // The override resource policy. Required when Enabled=true.
+    // When nil and Enabled=true, the webhook rejects the Pool.
     ResourcePolicy *ResourcePolicy `json:"resourcePolicy,omitempty"`
 }
 
 type ClassRef struct {
+    // Name of the SandboxPoolClass this Pool belongs to. Set by the Controller.
     Name string `json:"name"`
 }
 
 type NodeCtlConfig struct {
+    // node-ctl socket endpoint. Currently only unix:// is supported.
+    // placeholder-agent obtains the actual address via --node-ctl-socket;
+    // this field is informational and reserved for future declarative reconciliation.
     Endpoint string `json:"endpoint,omitempty"`
 }
 ```
@@ -318,7 +342,7 @@ type PlaceholderPodInfo struct {
 
 type PlaceholderAgentInfo struct {
     Version        string       `json:"version,omitempty"`
-    LastHeartbeat  *metav1.Time `json:"lastHeartbeat,omitempty"` // Updated by placeholder-agent on each status Patch; used by Controller for staleness detection
+    LastHeartbeat  *metav1.Time `json:"lastHeartbeat,omitempty"` // Updated by placeholder-agent on each status Apply; informational — Controller uses Lease TTL for liveness detection
     NodeCtlStarted bool         `json:"nodeCtlStarted,omitempty"`
     NodeCtlStartAt *metav1.Time `json:"nodeCtlStartAt,omitempty"`
 }
@@ -366,13 +390,6 @@ type PoolInfo struct {
 **Naming Rules**: SandboxPool naming format is `placeholder-{node-name}`, e.g. `placeholder-node-1`. Naming is deterministic per node (one Pool name per node), so Kubernetes API create conflict provides atomicity — when two Classes concurrently create a Pool for the same node, they produce the same name; the API Server accepts only the first, returning AlreadyExists for the second. Class ownership is identified via `spec.classRef.name` and the `sandbox-pool.io/class` Label; do not reverse-derive from the Pool name.
 
 > **Long Node Name Truncation**: Kubernetes object names are limited to 253 characters. The `placeholder-` prefix is 12 characters, so the maximum node name length is 241 characters. If the node name exceeds 241 characters, truncate it to 232 characters and append `-` + the first 8 hex characters of SHA256(node-name) as a suffix, yielding a total length of `placeholder-` (12) + 232 + 1 + 8 = 253 ≤ 253. The truncated name remains deterministic per node, preserving create conflict atomicity.
-
-**Validation Rules**:
-- When `override.enabled=true`, `override.resourcePolicy` must be non-nil
-- When `override.enabled=true`, `override.reason` must not be empty
-- `spec.nodeName` is immutable after creation (Webhook validation) — changing it would detach the Pool from its node and break the 1-Pool-per-node binding
-- `spec.classRef.name` is immutable after creation (Webhook validation) — changing it would detach the Pool from its Class
-- A single node cannot have Pools of different Classes (four-layer protection, see Validating Webhook section below)
 
 **PrintColumn Configuration**:
 
@@ -654,7 +671,7 @@ Node startup
 | `NodeCtlHealthy` | placeholder-agent | node-ctl reachable and responding normally | node-ctl unreachable or abnormal response |
 | `ResizeInProgress` | placeholder-agent | Scale-up or scale-down in progress | No resize in progress |
 | `ResizeDeferred` | placeholder-agent | Downscale deferred due to high watermark | No deferred downscale operation |
-| `PlaceholderAgentHealthy` | sandboxpool-controller | Agent heartbeat within 2min (`status.placeholderAgent.lastHeartbeat` fresh) | Agent never reported / heartbeat timeout > 2min (agent crashed but Node exists) |
+| `PlaceholderAgentHealthy` | sandboxpool-controller | Agent Lease active (renewed within TTL, default 40s) | Agent never reported / Lease expired (TTL > 40s with no renewal, agent crashed but Node exists) |
 | `NodeNotFound` | sandboxpool-controller | Node does not exist; Pool is orphaned | Node exists |
 
 ### Label and Annotation System
@@ -673,7 +690,7 @@ Node startup
 | Role | Permission Scope | Description |
 |------|-----------------|-------------|
 | `sandbox-pool-controller` | ClusterRole | sandboxpoolclasses (get/list/watch), sandboxpools (CRUD+status), nodes (get/list/watch), pods (get/list), events (create/patch), leases (full) |
-| `sandbox-pool-placeholder-agent` | ClusterRole | sandboxpools (get/list/watch), sandboxpools/status (patch), endpoints (full), events (create/patch) |
+| `sandbox-pool-placeholder-agent` | ClusterRole | sandboxpools (get/list/watch), sandboxpools/status (patch), endpoints (full), events (create/patch), leases (create/get/update) |
 | `sandbox-pool-admin` | ClusterRole | CRUD SandboxPoolClass + SandboxPool |
 | `sandbox-pool-viewer` | ClusterRole | Get/List/Watch |
 | `sandbox-pool-operator` | ClusterRole | Update SandboxPool (override operations) |
@@ -682,14 +699,7 @@ Node startup
 
 ### Validating Webhook
 
-| Validation Target | Rule | Description |
-|------------------|------|-------------|
-| SandboxPoolClass UPDATE | `spec.selector` immutable | Prevents mutation of running node set |
-| SandboxPool UPDATE | `spec.nodeName` immutable | Changing it would detach the Pool from its node and break the 1-Pool-per-node binding |
-| SandboxPool UPDATE | `spec.classRef.name` immutable | Changing it would detach the Pool from its Class |
-| SandboxPool CREATE/UPDATE | When override.enabled=true, override.resourcePolicy must be non-nil | Prevents inability to derive valid resource quota |
-| SandboxPool CREATE/UPDATE | When override.enabled=true, override.reason must not be empty | Audit traceability |
-| SandboxPool CREATE | A single node cannot have Pools of different Classes | Second layer of four-layer protection (defense in depth, see below) |
+The ValidatingWebhook enforces the field-level validation rules documented in the API type definitions above (see `SandboxPoolClassSpec` and `SandboxPoolSpec` field comments). Key invariants: `spec.selector` / `spec.nodeName` / `spec.classRef.name` immutability; `override.resourcePolicy` and `override.reason` required when `override.enabled=true`; single-Class-per-node enforced via four-layer protection below.
 
 > **Four-Layer Protection (Single-Class-per-Node Invariant)**:
 >
@@ -717,7 +727,7 @@ Node startup
 | Risk | Impact | Mitigation |
 |------|--------|-----------|
 | Phase update delay (≤30s) | Delay between Conditions change and Phase update | Alert rule sets `for ≥ 1m` buffer to avoid flapping |
-| Conditions stuck at stale values when placeholder-agent unreachable | Phase may not reflect latest state | Controller detects agent heartbeat staleness via `PlaceholderAgentHealthy` Condition (`status.placeholderAgent.lastHeartbeat` > 2min), downgrading Phase to Degraded/Unready; NodeNotFound covers node deletion scenario |
+| Conditions stuck at stale values when placeholder-agent unreachable | Phase may not reflect latest state | Controller detects agent heartbeat staleness via `PlaceholderAgentHealthy` Condition (expired Lease (TTL > 40s)), downgrading Phase to Degraded/Unready; NodeNotFound covers node deletion scenario |
 | SSA Apply concurrent conflict | Both sides SSA Apply status simultaneously | SSA field manager field-level non-overwrite; conflict auto-retried on next Reconcile/reportStatus |
 | selector immutability causes operational inconvenience | Adjusting node range requires deleting and recreating Class | Alternative approaches provided: taint+toleration, remove node Label |
 | EverReady Condition precise tracking | Already implemented via sticky `ConditionEverReady` (set True on first Ready, never reset) | Implemented, no additional field needed |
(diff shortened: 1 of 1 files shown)


ranxi2001:
Could this PR use `Refs #430` instead of `Fixes #430`?

My understanding is that #430 is a broader architecture discussion, while this proposal focuses on the slow resource track / SandboxPool management. Using `Fixes #430` would close the discussion when this PR is merged, which may be too early if the fast-track lifecycle pieces remain out of scope.

vivek41-glitch:
@lichuqiang please commit signoff statement too , dco also failling

vivek41-glitch:
> Could this PR use `Refs #430` instead of `Fixes #430`?
> 
> My understanding is that #430 is a broader architecture discussion, while this proposal focuses on the slow resource track / SandboxPool management. Using `Fixes #430` would close the discussion when this PR is merged, which may be too early if the fast-track lifecycle pieces remain out of scope.

abso , this is discussion issue not a bug or need to fix , u should first purpose and discuss then open pr @lichuqiang

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,643 @@
...
+│  └──────────────────────┘                                            │
+│       uses carved-out resources to create sandboxes                  │
+└──────────────────────────────────────────────────────────────────────┘
+```
+
+### Core Design Decisions
+
+| Decision | Choice | Reason |
+|----------|--------|--------|
+| Placeholder Pod model | Static Pod | mirror pod rebuilt within < 5s after deletion; naturally immune to eviction; no PDB needed |
```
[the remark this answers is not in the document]
lichuqiang: The 5 seconds here is just a rough estimate; actually deleting the mirror pod won't affect the resource usage of the placeholder pod.

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,643 @@
...
+| Placeholder Pod execution mode | No actual process, skip cgroup | placeholder-agent is a host-level service, runs before the Pod |
```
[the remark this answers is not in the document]
lichuqiang: The placeholder-agent is a host process and doesn't involve any actual containers. The image-related configurations are just there to formally meet k8s configuration requirements (there are related notes later in the document).

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,643 @@
...
+| Resource adjustment method | VPA InPlaceResize | Adjust resources online without rebuilding Pods; watermark protection during downscale prevents running sandbox OOM |
+
+### Component/Object Responsibility Matrix
+
+| Layer | Component/Object | Responsibilities | Interaction with node-ctl | Deployment Form |
+|-------|-----------------|------------------|--------------------------|-----------------|
+| Global policy layer | SandboxPool Controller | Class→Pool mapping + node selection + policy snapshot sync + Phase aggregation + Finalizer + node check | ❌ None | Deployment + leader election |
+| Node execution layer | placeholder-agent | CRI handler + manifest management + node-ctl full proxy + CRD Watch + Conditions reporting | ✅ Sole interaction party | host-level systemd service |
+| Node execution layer | Static Pod (mirror pod) | Placeholder locking of scheduling resources (resources.requests/limits) + RuntimeClass routing CRI calls to placeholder-agent + no actual process/no cgroup created | Indirect: node-ctl uses its carved-out resources to create sandboxes | kubelet auto-maintained from manifest |
```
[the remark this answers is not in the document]
lichuqiang: same as above

acsoto:
I want to better understand the relationship with the existing agent-sandbox WarmPool path. Today `CodeInterpreter.warmPoolSize` uses SandboxTemplate/SandboxWarmPool/SandboxClaim to pre-create session-level sandboxes. This proposal seems to introduce a node-level resource pool for the new decoupled path. Are these intended to coexist as separate modes, or is SandboxPool expected to become the lower layer that eventually replaces the current WarmPool path?

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,643 @@
...
+| CRD v1 | 1.16 GA | None |
+| SSA FieldOwner | 1.18 GA | None |
+| VPA InPlaceResize | 1.27 Alpha / 1.31 GA | FeatureGate: InPlacePodVerticalScaling |
+
+## Risks and Mitigations
+
+| Risk | Impact | Mitigation |
+|------|--------|-----------|
+| Phase update delay (≤30s) | Delay between Conditions change and Phase update | Alert rule sets `for ≥ 1m` buffer to avoid flapping |
+| Conditions stuck at stale values when placeholder-agent unreachable | Phase may not reflect latest state | NodeNotFound covers node deletion; long unreachability causes Node to become NotReady, indirectly covering |
```
ranxi2001: I have one question about this stale-status mitigation.

My reading is that `placeholder-agent` owns the node-local conditions and patches status every 30s, while the controller owns `phase` and `NodeNotFound`. If the Kubernetes Node still exists but `placeholder-agent` has stopped or can no longer patch status, the last successful `PlaceholderPodReady=True` / `NodeCtlHealthy=True` values may stay in status.

Could the proposal make the stale-status rule explicit for that case? For example, should the controller derive `Degraded` / `Unready` from a `lastHeartbeat` timeout, Node readiness, condition timestamps, or a controller-owned `PlaceholderAgentHealthy` condition?

This would make the phase aggregation and fault-injection tests clearer.

lichuqiang:
> I want to better understand the relationship with the existing agent-sandbox WarmPool path. Today `CodeInterpreter.warmPoolSize` uses SandboxTemplate/SandboxWarmPool/SandboxClaim to pre-create session-level sandboxes. This proposal seems to introduce a node-level resource pool for the new decoupled path. Are these intended to coexist as separate modes, or is SandboxPool expected to become the lower layer that eventually replaces the current WarmPool path?

@acsoto The topic discussed in issue 430 and the current agentcube implementation represent more of a two-generation architecture relationship. 
In the new architecture, to significantly improve sandbox startup efficiency and management scale, we will build an independent, lighter-weight, and   higher-performance sandbox control plane (the so-called "fast resource"). 
Sandboxes will no longer be deployed as Pods, and therefore Kubernetes nodes cannot be directly used to run sandboxes. This requires a upfront sandbox resource pool management system (the so-called"slow resource") — which is what is being designed here. It can be thought of as a prerequisite resource preparation step for sandbox deployment. 
What SandboxTemplate/SandboxClaim resource models correspond to is the related content of the fast resource system.

docs/proposals/sandbox-pool-management/README.md
lichuqiang: Good suggestion. Updated related content

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,644 @@
...
+### K8s Version Compatibility
+
+| Feature | Minimum K8s Version | Dependencies |
+|---------|-------------------|--------------|
+| Static Pod | 1.0 | None |
+| mirror Pod auto-rebuild | 1.0 | None |
+| ValidatingWebhook | 1.16 GA | None |
+| CRD v1 | 1.16 GA | None |
+| SSA FieldOwner | 1.18 GA | None |
+| VPA InPlaceResize | 1.27 Alpha / 1.31 GA | FeatureGate: InPlacePodVerticalScaling |
```
ranxi2001: I think the compatibility assumption here needs clarification because the proposal combines two behaviors that the native Kubernetes resize path does not currently support together.

The proposal makes both Static Pod resource locking and no-rebuild `VPA InPlaceResize` part of the v1alpha1 core scope. However, Kubernetes [KEP-1287](https://github.com/kubernetes/enhancements/blob/481e3710995d9996e27c7364b14d9ab870b65e74/keps/sig-node/1287-in-place-update-pod-resources/README.md#L332-L336) explicitly lists "The pod is a static pod" as an `Infeasible` in-place resize case. Its [metadata](https://github.com/kubernetes/enhancements/blob/481e3710995d9996e27c7364b14d9ab870b65e74/keps/sig-node/1287-in-place-update-pod-resources/kep.yaml#L35-L42) records v1.27 alpha, v1.33 beta, and v1.35 stable, rather than v1.31 GA.

Could the proposal clarify which mechanism Phase 3 intends to use?

- Native Kubernetes `/resize`: Static Pods are unsupported.
- Local Static Pod manifest update: if kubelet rebuilds the Pod, this is not a no-rebuild resize, and the resource-locking behavior during replacement needs to be defined.
- A custom `placeholder-agent` / runtime path: could the proposal describe how Kubernetes-visible requests, mirror Pod state, and node-ctl's actual limits remain consistent, instead of treating it as the native `InPlacePodVerticalScaling` path?

This distinction changes the compatibility table, implementation contract, and e2e acceptance criteria.

lichuqiang: You're right — "VPA" here is more of an analogy, and should not depend on the community's pod VPA feature. Static pod manifest resource spec updates actually go through a delete-and-recreate process, but during the window kubelet can correctly account for the static pod's resource usage, ensuring that the SandboxPool's resources are not consumed by mis-scheduled pods during that period (this is what we care about most).   I've added the relevant explanation in the "Update Flow" section.

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,659 @@
...
+metadata:
+  name: placeholder
+handler: placeholder
+overhead:
+  podFixed:
+    cpu: 10m
+    memory: 16Mi
+```
+
+kubelet routes CRI calls to placeholder-agent's Socket (`/run/sandbox-pool/cri.sock`) based on the RuntimeClass handler name. The overhead values remain unchanged during VPA resize.
```
ranxi2001: I have one question about the node-side RuntimeClass / CRI integration contract.

The updated proposal makes custom CRI interception part of the resource-adjustment path. It also says that kubelet routes CRI calls to `/run/sandbox-pool/cri.sock` based on the `placeholder` RuntimeClass handler, with `placeholder-agent` implementing the CRI server.

In [kubelet v1.36.2](https://github.com/kubernetes/kubernetes/blob/v1.36.2/pkg/kubelet/kuberuntime/kuberuntime_sandbox.go#L56-L68), RuntimeClass is resolved to a `runtimeHandler` and passed to the already configured `runtimeService.RunPodSandbox(...)`; it does not select a second CRI endpoint. With containerd, the handler selects a configured runtime/shim (or sandboxer integration) behind the same CRI service.

Could the proposal specify the missing integration layer between the `placeholder` handler and `placeholder-agent`?

- Is `placeholder-agent` intended to integrate as a containerd runtime v2 shim / sandboxer?
- Is there a CRI proxy in front of containerd that dispatches by `runtime_handler`?
- Or is the node-wide kubelet CRI endpoint expected to point directly to `placeholder-agent`, with normal workloads forwarded elsewhere?

This determines which API `placeholder-agent` must implement, how normal Pods continue to use the default runtime, and whether the proposed per-Pod routing can work without replacing the node's CRI path.

lichuqiang: Good question. The expectation here is to serve as the runtime implementation behind containerd, registered as a plugin into containerd. Therefore placeholder-agent implements the v2 shim interface. 
I have already corrected the relevant content in the documentation.

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,682 @@
...
+| Resource adjustment method | containerd runtime shim (VPA analogy) | As a containerd shim v2 plugin, intercepts container creation calls to adjust resources online; watermark protection during downscale prevents running sandbox OOM |
+
+### Component/Object Responsibility Matrix
+
+| Layer | Component/Object | Responsibilities | Interaction with node-ctl | Deployment Form |
+|-------|-----------------|------------------|--------------------------|-----------------|
+| Global policy layer | SandboxPool Controller | Class→Pool mapping + node selection + policy snapshot sync + Phase aggregation + Finalizer + node check | ❌ None | Deployment + leader election |
+| Node execution layer | placeholder-agent daemon | manifest management + node-ctl full proxy + CRD Watch + Conditions reporting | ✅ Sole interaction party | host-level systemd service |
+| Node execution layer | placeholder-agent shim | containerd runtime shim v2 (Task API), stateless TTRPC proxy; business logic moved to daemon, resource adjustment in daemon.Create (shim forwards) | ✅ via daemon (shim forwards via TTRPC, daemon calls node-ctl) | containerd runtime plugin (config.toml registration) |
```
ranxi2001, line 140: One implementation-contract question about the no-workload-process shim:

The proposal describes the shim as a stateless TTRPC proxy and says `Start` only marks the container as Running. However, containerd's [Task v2 service](https://github.com/containerd/containerd/blob/ba015360dd6857dbd0f0b36912d89a5946220057/api/runtime/task/v2/shim.proto#L29-L50) expects the shim to own the task lifecycle and exposes `Create`, `Start`, `State`, `Kill`, `Wait`, `Delete`, `Connect`, and `Shutdown`; the responses also carry PID and exit status/time.

What task or process represents this placeholder Pod, and how will PID, `State`, `Wait`, and exit events be satisfied when there is no workload process? Could the Phase 2 plan include a minimal shim spike/e2e covering create/start, CRI readiness, stop/remove, wait/exit, and reconnect behavior before `PlaceholderPodReady` is derived from “shim task created”?

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,682 @@
...
+### Status Writers and SSA Coordination
+
+SandboxPool status is managed by two components via the SSA FieldOwner mechanism, ensuring non-overlapping writes:
+
+| FieldOwner | Component | Managed Fields | Write Timing |
+|-----------|-----------|---------------|--------------|
+| `placeholder-agent` | placeholder-agent | placeholderPod, placeholderAgent, override, resize, nodeCtl, poolInfo, conditions (non-NodeNotFound/PlaceholderAgentHealthy), lastAppliedGeneration | Periodic Patch every 30s |
+| `sandboxpool-controller` | SandboxPool Controller | **phase**, conditions (NodeNotFound, PlaceholderAgentHealthy) | During Pool Reconcile |
+
+This "data source and aggregator separated" design ensures Phase can still be updated in extreme scenarios — when a node is deleted, the Controller sets Phase=Unready based on the NodeNotFound Condition; when the agent crashes but the Node still exists, the Controller detects stale `PlaceholderAgent.LastHeartbeat` (> 2min) and sets `PlaceholderAgentHealthy=False`, downgrading Phase to Degraded/Unready.
```
ranxi2001: Thanks for adding the separate `status.placeholderAgent.lastHeartbeat`; that makes the health signal clear.

One remaining timing question: the controller derives `PlaceholderAgentHealthy` during Pool Reconcile, but if the agent stops after a fresh status patch, there may be no later object update when that timestamp crosses the two-minute threshold.

Could the proposal specify how the timeout schedules another reconciliation, for example `RequeueAfter` based on the heartbeat expiry or a periodic controller sweep? It would also help to include a fault-injection case where the agent stops and no further Pool/Node events occur, verifying that the Phase still leaves `Ready` after the timeout.

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,682 @@
...
+        ▼
+kubelet detects manifest removal → automatically stop Static Pod
+  → containerd calls shim.Shutdown → shim cleanup; daemon stops node-ctl when Pool is deleted
+  → containerd RemovePodSandbox → clean up shim state
+        │
+        ▼
+SandboxPool Controller checks Finalizer
+  → isPlaceholderAgentCleanupDone: query whether mirror Pod has been removed
+  → mirror Pod removed → remove Finalizer → Pool deletion complete
+  → Timeout (> 10min) → force remove Finalizer + Warning Event
```
ranxi2001, line 589: I have one failure-recovery question about the forced finalizer timeout.

The deletion flow says that `placeholder-agent` removes the node-local manifest, while the controller force-removes the Pool finalizer after 10 minutes if cleanup has not completed. If the agent is unreachable for that whole window, the `SandboxPool` can disappear from the API while its local Static Pod manifest still exists on the node.

When the agent later recovers, how is that orphaned manifest detected and removed? A watch of current Pool objects would not replay the already-completed deletion. Could the proposal add a startup reconciliation rule that scans agent-managed manifests and removes any whose Pool UID no longer exists, or describe another durable tombstone/cleanup-ack mechanism?

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,682 @@
...
+### Validating Webhook
+
+| Validation Target | Rule | Description |
+|------------------|------|-------------|
+| SandboxPoolClass UPDATE | `spec.selector` immutable | Prevents mutation of running node set |
+| SandboxPool UPDATE | `spec.nodeName` immutable | Changing it would detach the Pool from its node and break the 1-Pool-per-node binding |
+| SandboxPool UPDATE | `spec.classRef.name` immutable | Changing it would detach the Pool from its Class and invalidate the naming convention |
+| SandboxPool CREATE/UPDATE | When override.enabled=true, override.resourcePolicy must be non-nil | Prevents inability to derive valid resource quota |
+| SandboxPool CREATE/UPDATE | When override.enabled=true, override.reason must not be empty | Audit traceability |
+| SandboxPool CREATE | A single node cannot have Pools of different Classes | First layer of four-layer protection (Webhook is the second atomic layer) |
```
ranxi2001: I have one concurrency question about the single-Class-per-node invariant.

The proposal says a node cannot have Pools from different Classes, but Pool names include both the Class and node names. If two Classes select the same node, concurrent reconciles can therefore create two different Pool objects. A validating webhook that lists existing Pools cannot make that cross-object check atomic: concurrent create requests could both observe no conflicting Pool and both be admitted.

Could the proposal define an atomic ownership mechanism for each node, such as a stable per-node claim/name that uses Kubernetes create conflict, or a deterministic Class precedence rule with reconciliation of the loser? A concurrent-create integration test would make this invariant verifiable.

ranxi2001 (commented):
I reviewed the latest force-pushed proposal. The recent updates resolve the earlier heartbeat-source, RuntimeClass routing, optional-field, and SSA condition ownership questions. I left four focused questions about time-driven reconciliation, per-node ownership atomicity, deletion recovery, and the containerd Task contract.

docs/proposals/sandbox-pool-management/README.md
lichuqiang: Thanks for your feedback. Most of these issues were actually considered during the local detailed design phase (exceptbfor the single-Class-per-node one). 
To keep the proposal concise, I didn't write them out in detail. To avoid misunderstanding, I'll now add the relevant content back in.

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,743 @@
+---
+title: SandboxPool Resource Pool Management
```
RainbowMango: The proposal name, which will be used as the feature name, feels redundant with the double `pool`. 

Here are some alternatives:
- Sandbox Resource Reservation
- Sandbox Resource Pool Management

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,743 @@
...
+  - TBD
+approvers:
+  - TBD
+creation-date: 2026-07-06
+last-updated: 2026-07-06
+status: draft
+tracking-issue: "#430"
+---
+
+# SandboxPool Resource Pool Management
```
RainbowMango: ditto.

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,743 @@
...
+2. **Fast resources**: An independent, high-performance, lightweight control plane supporting sandbox ecosystem management.
+
+The end goal is co-location of regular Pods and sandboxes on the same node.
+
+### Scope of This Proposal
+
+This proposal focuses on the **slow resource** part of the next-generation architecture — the design of SandboxPool resource pool management. The system implements declarative management of sandbox resources through a two-layer architecture:
+
+- **Global policy layer** (SandboxPool Controller): Responsible for Class → Pool mapping, node selection, policy snapshot synchronization, and Phase aggregation.
+- **Node execution layer** (placeholder-agent): Acts as the node-local agent, responsible for resource policy enforcement, watermark checks, and status reporting.
```
RainbowMango, line 38: What do you think of renaming the `placeholder-agent` to `agentcube-pool-agent`?

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,743 @@
...
+
+Core designs include:
+
+- Using the **Static Pod model** to lock scheduling resources; mirror pods are automatically rebuilt after accidental deletion.
+- Declarative state synchronization based on **CRD Watch/List**.
+- Implements vertical resource scaling via **containerd runtime shim (v2 Task API)** (VPA analogy), independent of InPlacePodVerticalScaling Feature Gate.
```
RainbowMango: Can you help add a link to the `containerd runtime shim (v2 Task API)` as a reference?

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,743 @@
...
+- Coordinating multi-component status writes via the **SSA (server-side apply) field manager mechanism** to avoid field-level conflicts.
+
+## Motivation
+
+In co-location scenarios, regular container workloads and sandbox workloads coexist on the same node. There is currently no Kubernetes-native way to:
```
RainbowMango: Could you please elaborate on why the cluster administrator wants or needs to build `sandbox pool` in the co-location scenarios?

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,743 @@
...
+- **Elastically scale sandbox resources**: When sandbox load changes dynamically, resource quotas need to be adjusted online without interrupting running sandboxes.
+- **Guarantee resource locking reliability**: Ensure that scheduling resources occupied by sandboxes are not preempted by other workloads, even in abnormal scenarios such as accidental deletion or node failure.
+- **Provide global observability**: Platform operators need a global view of the health status of sandbox resource pools across all nodes.
+
+### Goals
+
+- Provide declarative CRD APIs (SandboxPoolClass + SandboxPool) to manage node-level sandbox resource pools.
+- Reliably lock scheduling resources through the Static Pod model; mirror pods are automatically rebuilt after accidental deletion (typically within seconds, depending on kubelet sync interval).
+- Support adjusting placeholder Pod resource quotas via containerd runtime shim. This is NOT Kubernetes in-place Pod resize (InPlacePodVerticalScaling) — the placeholder Pod is rebuilt through Static Pod manifest update (delete/recreate), but running sandboxes managed by node-ctl are not interrupted during the rebuild.
```
RainbowMango, line 60: > node-ctl

Note: In the other discussion, the component name `node-ctl` has been deprecated.

RainbowMango (commented):
Review is still ongoing.

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,743 @@
...
+| **node-ctl stability** | Scale-driven rebuild does NOT stop node-ctl; only genuine deletion (manifest removal) stops it | The daemon checks if the manifest file still exists when handling Pool deletion, distinguishing scale-driven rebuild from real teardown |
+| **Node-level cgroup** | node-ctl keeps running during rebuild, cgroup limits uninterrupted | sandbox resources managed by node-ctl at node-level cgroup, decoupled from Pod `resources.requests` |
+| **Scheduler view** | mirror pod is deleted+recreated by kubelet on manifest hash change (NOT patched), brief window in API server | Safety does NOT rely on scheduler view consistency — kubelet admit check uses local podManager (always contains static pod), which is the real safety gate; conflicting new Pods are admit-failed |
+| **New static Pod startup** | Created locally by kubelet, does not go through scheduler | Regardless of mirror pod state, the new static Pod starts normally |
+| **Downscale watermark protection** | Downscale checks node-ctl watermark; if above risk threshold, deferred to Deferred status | Prevents downscale from causing OOM in running sandboxes; auto-retries when watermark is safe |
```
ranxi2001, line 554: Could the ordering be made explicit before calling the Deferred path safe? The update flow at lines 510-519 rewrites the Static Pod manifest before the rebuilt shim calls `daemon.Create` and applies the node-ctl policy. On an unsafe downscale, the lower Pod request can therefore become the kubelet/scheduler reservation while node-ctl still retains the old allocation and reports `Deferred`. At that point the local podManager also sees the lower request, so it no longer protects the gap. Please define staged desired/applied ordering: reserve first, then expand node-ctl for scale-up; shrink node-ctl first, then lower the manifest for scale-down; and retain the last applied request while Deferred. The resize test should assert that scheduler-visible requests never fall below the applied sandbox allocation.

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,743 @@
...
+type ResourcePolicy struct {
+    CPU    resource.Quantity `json:"cpu"`
+    Memory resource.Quantity `json:"memory"`
+}
+
+type PlaceholderPodTemplateSpec struct {
+    Annotations     map[string]string `json:"annotations,omitempty"`
+    Tolerations     []corev1.Toleration `json:"tolerations,omitempty"`
+    RuntimeClassName *string           `json:"runtimeClassName,omitempty"`
+    PriorityClassName string           `json:"priorityClassName,omitempty"`
```
ranxi2001: Kubernetes documents that Static Pods do not support `priorityClassName`; avoiding node-pressure eviction requires setting the numeric `priority` field directly. This field therefore cannot provide the protection described later in the proposal. Could the API model `Priority *int32` (or define a fixed node-critical numeric priority in the generated manifest) instead? https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/#self-healing-behavior

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,743 @@
...
+                                         + NodeCtlHealthy=True → Ready
+```
+
+| Phase | Entry Condition | Exit Condition |
+|-------|----------------|----------------|
+| `Pending` | PlaceholderPodReady=False (never Ready) | PlaceholderPodReady=True + NodeCtlHealthy=True → Ready |
+| `Ready` | PlaceholderPodReady=True + NodeCtlHealthy=True + ResourceSynced=True (or ResizeDeferred=True) | NodeCtlHealthy=False → Degraded; NodeNotFound=True → Unready; PlaceholderAgentHealthy=False → Degraded |
+| `Degraded` | NodeCtlHealthy=False (< 5min), or ResourceSynced=False (non-Deferred), or PlaceholderAgentHealthy=False (agent unreachable but Pod still shows Ready) | NodeCtlHealthy=True + ResourceSynced=True → Ready; NodeCtlHealthy=False ≥5min → Unready; PlaceholderAgentHealthy=True → Ready |
+| `Unready` | NodeCtlHealthy=False (≥ 5min), or PlaceholderPodReady=False (abnormal after being Ready), or NodeNotFound=True, or PlaceholderAgentHealthy=False + PlaceholderPodReady=False (was Ready) | PlaceholderPodReady=True + NodeCtlHealthy=True → Ready; NodeNotFound=False → re-evaluate all conditions; PlaceholderAgentHealthy=True → Ready |
```
ranxi2001: The `PlaceholderAgentHealthy=True -> Ready` shortcut here, and in the Degraded row above, bypasses the Ready invariant. Agent recovery does not imply that `NodeCtlHealthy`, `PlaceholderPodReady`, and `ResourceSynced` are also true. Could both shortcuts say “re-evaluate all conditions,” with only the aggregate `computePhase` Ready predicate allowed to enter `Ready`?

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,743 @@
...
+| `sandbox-pool.io/pool-uid` | Pod | Associated Pool UID (Annotation only, written into manifest); used by agent at startup for orphaned manifest detection — manifests with non-matching UID are removed |
+| `sandbox-pool.io/skip-cgroup` | Pod | Hint to skip cgroup creation (written to both Label and Annotation); requires placeholder shim support, not default K8s behavior |
+| `sandbox-pool.io/manifest-hash` | Pod | manifest SHA256 hash (Annotation only) |
+
+### RBAC Permission Model
+
+| Role | Permission Scope | Description |
+|------|-----------------|-------------|
+| `sandbox-pool-controller` | ClusterRole | sandboxpoolclasses (get/list/watch), sandboxpools (CRUD+status), nodes (get/list/watch), pods (get/list), events (create/patch), leases (full) |
+| `sandbox-pool-placeholder-agent` | ClusterRole | sandboxpools (get/list/watch), sandboxpools/status (patch), endpoints (full), events (create/patch) |
```
ranxi2001: This ClusterRole lets every node-local agent patch every SandboxPool status. Filtering the watch by `spec.nodeName` is only a client-side convention, so a compromised node could forge another node's pool health. Could the proposal define a server-enforced write boundary, for example admission that binds the authenticated node identity to `spec.nodeName`, or a controller-owned status aggregation path? This security boundary affects status trust and should be established before treating RBAC as Phase 5 wiring.

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,743 @@
...
+   ├── 3. GET SandboxPool CRD (by UID or by name)
+   │      → Pool exists and UID matches → keep manifest (normal path)
+   │      → Pool does not exist or UID mismatch → delete manifest file
+   │
+   ├── 4. kubelet detects manifest removal → auto-stops residual static Pod
+   │
+   └── 5. After cleanup completes, enter normal Watch loop
+```
+
+> **Design Constraint**: Cleanup runs only at daemon startup (not continuous polling), avoiding runtime overhead. The `sandbox-pool.io/pool-uid` annotation in the manifest is the authoritative orphan criterion — only UID mismatch triggers deletion, preventing accidental removal of active manifests during temporary API Server unavailability.
```
ranxi2001: This constraint conflicts with line 579: that line deletes on “Pool does not exist or UID mismatch,” while this line says only UID mismatch triggers deletion. Kubernetes also cannot GET an object by UID. Could the startup algorithm specify GET-by-name followed by UID comparison: delete on NotFound or UID mismatch, preserve and retry on transient/Forbidden/API-unavailable errors, and enter the normal watch loop only after startup reconciliation reaches a defined outcome?

ranxi2001 (commented):
Follow-up review on current head `c2f2502` after reading the full proposal and existing threads. These inline comments focus on design contracts that are new or remain unresolved after the latest edits. I used AI-assisted tooling to help trace the proposal and verified each concern against the current text and upstream contracts.

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,743 @@
...
+│  └────────────────────────────────────────────────────────────────┘  │
+│              uses carved-out resources to create sandboxes            │
+└──────────────────────────────────────────────────────────────────────┘
+```
+
+### Core Design Decisions
+
+| Decision | Choice | Reason |
+|----------|--------|--------|
+| Placeholder Pod model | Static Pod | mirror pod auto-recreated by kubelet from manifest after deletion/eviction (typically within seconds, depending on kubelet sync interval); continuously recreated as long as the manifest exists; no PDB needed. During rebuild, kubelet's local podManager still holds the static pod, so resource accounting (admit checks, node allocatable) is unaffected — resource locking does not depend on the mirror pod's existence in the API |
```
RainbowMango, line 124: > continuously recreated as long as the manifest exists

The reason for using static pods is the concern that pods might be accidentally deleted. But what if the manifest file is accidentally deleted?

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,743 @@
...
+| Placeholder Pod execution mode | No workload process | placeholder-agent runs as a host-level systemd service; the manifest uses pause:3.9 only to satisfy K8s Pod schema requirements; cgroup creation is skipped via the placeholder shim, not by default K8s behavior |
+| Storage backend | Pure etcd (K8s API Server) | Resource pool control plane only needs CRD state; no external storage needed |
+| CRD Scope | Cluster-scoped | placeholder-agent is a node-local singleton (one per node) bound to one Pool; no namespace isolation needed |
+| Phase calculator | Controller aggregation | When placeholder-agent is unreachable (node deleted or agent crash), it cannot update Phase; Controller aggregation + stale detection ensures Phase can still be updated in extreme scenarios |
```
RainbowMango, line 128: What is the `Phase` thing? How does the `Controller aggregation` work?

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,743 @@
...
+| Status write coordination | SSA Apply (fieldManager) | placeholder-agent SSA-Applies status every 30s; Controller occasionally SSA-Applies; field-level ownership avoids conflicts |
```
RainbowMango: It is concerning that each `placeholder-agent` updates the status against the CR so frequently that the kube-apiserver cannot handle the request in a large cluster with thousands of nodes.

docs/proposals/sandbox-pool-management/README.md
lichuqiang: addressed in lastest commit

docs/proposals/sandbox-pool-management/README.md
lichuqiang: done

docs/proposals/sandbox-pool-management/README.md
lichuqiang: Let's keep it as is for now; whether to change it will be decided after a broader discussion.

docs/proposals/sandbox-pool-management/README.md
lichuqiang: Basically similar to kubelet's behavior. We can reduce the update frequency somewhat, but the mechanism itself is hard to change.

docs/proposals/sandbox-pool-management/README.md
lichuqiang: See phase State Machine section for detail

docs/proposals/sandbox-pool-management/README.md
lichuqiang: add 'Manifest self-healing' section

lichuqiang:
Given that quite a few commits were added to address review comments, I squashed them to make it easier to read.

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,761 @@
...
+#### SandboxPoolClass
+
+**Group / Version / Kind / Scope**: `sandboxpool.agentcube.volcano.sh` / `v1alpha1` / `SandboxPoolClass` / `Cluster`
+
+**Spec Definition**:
+
+```go
+type SandboxPoolClassSpec struct {
+    // Target node selector; immutable after creation
+    Selector metav1.LabelSelector `json:"selector"`
```
RainbowMango: Would you mind explaining why the `Selector` is immutable? What's your concern here? 
It would be great to explain it in comments.

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,761 @@
...
+
+    // Additional node selection criteria
+    NodeSelector map[string]string `json:"nodeSelector,omitempty"`
```
RainbowMango: Can you elaborate on why the `NodeSelector` is needed in addition to the `Selector` above? What's the relationship between them?

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,761 @@
...
+
+    // Per-node resource quota
+    ResourcePolicy ResourcePolicy `json:"resourcePolicy"`
```
RainbowMango: Listing each resource as a field doesn't scale. With hardcoded `CPU` and `Memory`, we need to update the API when we try to support a new resource type, like `GPU, hugepages`, or vendor-specific resources like `example.com/cpu`. We should use the same structure as Pod, i.e. `corev1.ResourceList`:
```golang
Resources corev1.ResourceList `json:"resources"`
````

By the way, the name `ResourcePolicy` is not accurate; there is no policy here, just a declaration of resource amounts.

File: docs/proposals/sandbox-pool-management/README.md [hunk 28]
```diff
@@ -0,0 +1,761 @@
...
+
+    // Placeholder Pod template
+    PlaceholderPodTemplate *PlaceholderPodTemplateSpec `json:"placeholderPodTemplate,omitempty"`
```
RainbowMango:
```suggestion
    // Placeholder Pod template
    PlaceholderPod *PlaceholderPodTemplate `json:"placeholderPod,omitempty"`
```

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,761 @@
...
+
+    // node-ctl endpoint (informational only; placeholder-agent obtains the actual address
+    // via --node-ctl-socket startup parameter, not from this field)
+    // Reserved for future declarative reconciliation; currently only unix:// supported
+    NodeCtlEndpoint string `json:"nodeCtlEndpoint,omitempty"`
```
RainbowMango: Do we really expect the user to provide the information? I bet not, and suggest removing it from the API.

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,761 @@
...
+- `spec.nodeCtlEndpoint` (if non-empty) must start with `unix://`. `http://` / `https://` are reserved for future extension; the current version webhook rejects non-`unix://` values
+
+#### SandboxPool
+
+**Group / Version / Kind / Scope**: `sandboxpool.agentcube.volcano.sh` / `v1alpha1` / `SandboxPool` / `Cluster`
+
+**Spec Definition**:
+
+```go
+type SandboxPoolSpec struct {
```
RainbowMango, line 251: The comments in  SandboxPoolSpec  (and its sub-types like  OverrideSpec ,  ClassRef ,  NodeCtlConfig ) are too brief to understand the design intent. 

Could you please add complete comments for each field? Following the Kubernetes API convention: what the field means, who sets it, whether it is optional/immutable, and the default behavior when unset.

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,761 @@
...
+  1. Selector matches target nodes
+  2. Create SandboxPool CRD per node (with resourcePolicy snapshot)
+  3. Aggregate and update Class Status
+        │ Pool CRD created
+        ▼
+placeholder-agent daemon (systemd service)
+  1. Watch detects new SandboxPool CRD (filtered by node label)
+  2. Generate and write manifest file → /etc/kubernetes/manifests/sandbox-pool-{pool}.yaml
+  3. kubelet detects new manifest → creates Static Pod
+  4. kubelet → containerd (CRI) → starts placeholder-agent shim by runtimeClassName
```
RainbowMango, line 513: Do we need to change kubelet's configuration to let it handle the customized runtime class? 
If yes, we need to document it, as it is a serious constraint.

docs/proposals/sandbox-pool-management/README.md
```diff
@@ -0,0 +1,761 @@
...
+
+> **Security Boundary — Per-Node Status Write Scoping**: The `sandbox-pool-placeholder-agent` ClusterRole grants `sandboxpools/status (patch)` cluster-wide, which technically allows any node's agent to patch any Pool's status. To prevent a compromised node from forging another node's pool health, the ValidatingWebhook enforces a server-side write boundary: on a SandboxPool STATUS UPDATE, the webhook inspects the authenticated request identity (`userInfo.username` / `userInfo.groups` from the admission review) and rejects the request with `Forbidden` unless the caller is bound to `spec.nodeName`. Concretely, each placeholder-agent authenticates as a per-node identity — either a node-scoped client cert yielding `system:node:<nodeName>` in group `system:nodes`, or a dedicated per-node ServiceAccount whose name encodes the node name. The webhook extracts `<nodeName>` from the authenticated identity and compares it against `spec.nodeName`; only the agent running on that node may patch that Pool's status. This makes the client-side `spec.nodeName` watch filter a server-enforced invariant rather than a convention.
+
+### Validating Webhook
```
RainbowMango, line 700: We don't need to specify the validation items separately. They should be included in each field's comments.

docs/proposals/sandbox-pool-management/README.md
lichuqiang: Actually, this isn't necessary — it's a redundant design. Let me revise these API design suggestions

docs/proposals/sandbox-pool-management/README.md
RainbowMango: Nice catch! @ranxi2001

docs/proposals/sandbox-pool-management/README.md
lichuqiang: I prefer the current naming convention to preserve room for future extension — for example, someone has already mentioned the possibility of specifying resource specs by percentage later on.

docs/proposals/sandbox-pool-management/README.md
lichuqiang: No, as described here, it actually registers as a runtime shim implementation under containerd, so kubelet doesn't need to be directly aware of it.

docs/proposals/sandbox-pool-management/README.md
RainbowMango: Fair enougf.

docs/proposals/sandbox-pool-management/README.md
RainbowMango: I see, but that's a dead link: https://github.com/containerd/containerd/blob/main/api/runtime/task/v2/README.md

docs/proposals/sandbox-pool-management/README.md
RainbowMango: > **Manifest Self-Healing**: The Static Pod model protects against accidental **Pod** deletion (kubelet rebuilds from manifest). But if the **manifest file itself** is accidentally deleted, kubelet will stop the static pod. To cover this case, the placeholder-agent daemon's reconcile loop periodically verifies that the manifest exists for every active Pool on its node; if the manifest is missing while the Pool CRD still exists, the daemon rewrites it. This makes manifest recovery as automatic as mirror-pod recovery — the agent is the authoritative writer of the manifest, and the CRD (not the manifest) is the source of truth for desired state.

Yeah, I see, but this is still concerning because the kubelet would stop the static pod `immediately` once the manifest is deleted, then the resources are freed from the scheduler's view. Which means the scheduler may place regular pods onto the node, and when the `placeholder-agent` rewrites the manifest, the static pod may fail admission.

docs/proposals/sandbox-pool-management/README.md
RainbowMango: We already have lessons learned from `kubelet` heartbeat mechanism. Should we introduce `lease` for this case?

docs/proposals/sandbox-pool-management/README.md
lichuqiang: I think this should be classified as a constraint rather than a scenario that needs to be guarded against. No system mechanism can guarantee the prevention of any destructive operation — even cgroup configurations can be manually modified.

docs/proposals/sandbox-pool-management/README.md
lichuqiang: good suggestion

PR #437, opened 2026-07-13
examples: fix agent runtime and pcap configurations

acsoto:
**What type of PR is this?**

/kind documentation

Make the AgentRuntime SDK example runnable with the repository manifest by adding a minimal echo HTTP server and health probes. Document the deployment and invocation flow, and fix the PCAP analyzer's Router service address.

N/A

This change was prepared with AI assistance and manually verified.

Tested with:
- `python -m pytest sdk-python/tests/test_agent_runtime.py -q`
- `ruff check sdk-python/examples/agent_runtime_usage.py`
- Kubernetes API server dry-runs for both changed manifests
- Direct health, echo POST, and 404 checks against the embedded HTTP server

```release-note
Fix the AgentRuntime and PCAP analyzer examples so they use runnable and correctly addressed services.
```

### DIFF
--- a/example/agent-runtime/agent-runtime.yaml
+++ b/example/agent-runtime/agent-runtime.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: default
 spec:
   targetPort:
-    - pathPrefix: "/"
+    - pathPrefix: "/echo"
       port: 8080
       protocol: "HTTP"
   podTemplate:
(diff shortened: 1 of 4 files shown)


example/agent-runtime/agent-runtime.yaml
```diff
@@ -16,7 +16,68 @@ spec:
...
+                      if not isinstance(request, dict):
+                          self.send_json(400, {"error": "request body must be a JSON object"})
+                          return
+                      input_text = request.get("input", "")
+                      self.send_json(200, {"output": f"echo: {input_text}"})
+
+                  def log_message(self, format, *args):
+                      pass
+
+              http.server.ThreadingHTTPServer(("0.0.0.0", 8080), EchoHandler).serve_forever()
```
ranxi2001, line 84: The container starts `python3 -c` directly, so the embedded server runs as PID 1 and installs no SIGTERM handler. In a local reproduction using this image and command, `docker stop -t 2` waited 2.21s before exit 137 (`OOMKilled=false`); a version with a TERM exit handler stopped in 0.20s with exit 0. Since Kubernetes also terminates containers with TERM followed by KILL, and this Pod does not set `terminationGracePeriodSeconds`, session expiry or deletion can leave the container running until the default 30-second grace period ends, delaying Pod teardown and resource release. Could we add a SIGTERM exit path and a smoke test that verifies prompt, clean exit?

acsoto: good catch. fixed

PR #450, opened 2026-08-02
fix(workloadmanager): check CodeInterpreter child ownership

### LINKED ISSUE #449: CodeInterpreter controller can delete unowned SandboxTemplate resources
**What happened**:

The CodeInterpreter controller assumes that any SandboxTemplate or SandboxWarmPool with the same namespace/name belongs to the CodeInterpreter.

Existing resources are updated without checking their owner reference. When `warmPoolSize` is zero, the same-named resources are deleted unconditionally. This can overwrite or delete resources created by another user or controller.

The controller should only update or delete resources controlled by that CodeInterpreter. A name collision with an unowned resource should fail reconciliation and report a condition/event.

1. Create a standalone `SandboxTemplate` named `test`.
2. Create a `CodeInterpreter` named `test` in the same namespace with a valid template and `warmPoolSize: 0`.
3. Wait for the CodeInterpreter controller to reconcile.
4. Run:

   `kubectl get sandboxtemplate test`

The original template is deleted.

The update path can also be reproduced with `warmPoolSize: 1`; the existing template spec is overwritten while its ownership is never verified.

`SetControllerReference` is only called when creating a new child. The existing-resource update and delete paths do not check `metav1.IsControlledBy` or the controller UID.

Relevant code:
https://github.com/volcano-sh/agentcube/blob/0704bb96502af32f2bd90d47f1e11b4c8099959e/pkg/workloadmanager/codeinterpreter_controller.go#L143-L280

**Environment**:

- agentcube version: main (`0704bb9`)
- Kubernetes version: not version-specific
- Others: agent-sandbox v0.4.6

avinxshKD:
***What type of PR is this?**

/kind bug

Prevents the CodeInterpreter controller from updating or deleting same-named `SandboxTemplate` and `SandboxWarmPool` resources that it does not own

Ownership conflicts now fail reconciliation and set the CodeInterpreter Ready condition to `False` with reason `OwnershipConflict`.

Fixes #449

`AlreadyExists` from child creation is now returned so a concurrent name collision is retried and checked instead of being treated as success.

```release-note
CodeInterpreter no longer modifies or deletes unowned SandboxTemplate and SandboxWarmPool resources.
```

### DIFF
--- a/pkg/workloadmanager/codeinterpreter_controller.go
+++ b/pkg/workloadmanager/codeinterpreter_controller.go
@@ -292,7 +292,10 @@ func (r *CodeInterpreterReconciler) deleteSandboxWarmPool(ctx context.Context, c
 		return err
 	}
 
-	if err := r.Delete(ctx, warmPool); err != nil {
+	if err := r.Delete(ctx, warmPool, client.Preconditions{
+		UID:             &warmPool.UID,
+		ResourceVersion: &warmPool.ResourceVersion,
+	}); err != nil {
 		if !errors.IsNotFound(err) {
 			return fmt.Errorf("failed to delete SandboxWarmPool: %w", err)
 		}
@@ -315,7 +318,10 @@ func (r *CodeInterpreterReconciler) deleteSandboxTemplate(ctx context.Context, c
 		return err
 	}
 
-	if err := r.Delete(ctx, sandboxTemplate); err != nil {
+	if err := r.Delete(ctx, sandboxTemplate, client.Preconditions{
+		UID:             &sandboxTemplate.UID,
+		ResourceVersion: &sandboxTemplate.ResourceVersion,
+	}); err != nil {
 		if !errors.IsNotFound(err) {
 			return fmt.Errorf("failed to delete SandboxTemplate: %w", err)
 		}
--- a/pkg/workloadmanager/codeinterpreter_controller.go
+++ b/pkg/workloadmanager/codeinterpreter_controller.go
@@ -91,7 +91,7 @@ func (r *CodeInterpreterReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	}
 
 	// Update status with ready condition
-	if err := r.updateStatus(ctx, codeInterpreter); err != nil {
+	if err := r.updateStatus(ctx, codeInterpreter, true, "Reconciled", "CodeInterpreter is ready"); err != nil {
 		logger.Error(err, "failed to update status")
 		return ctrl.Result{}, err
 	}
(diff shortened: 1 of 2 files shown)


avinxshKD:
@ranxi2001 @acsoto ptal when u get chance, this prevents the CodeInterpreter controller from modifying or deleting same-named resources it doesn’t own.

pkg/workloadmanager/codeinterpreter_controller.go
```diff
@@ -287,6 +311,9 @@ func (r *CodeInterpreterReconciler) deleteSandboxTemplate(ctx context.Context, c
 	} else if err != nil {
 		return fmt.Errorf("failed to get SandboxTemplate: %w", err)
 	}
+	if err := r.validateChildOwnership(ctx, ci, sandboxTemplate, "SandboxTemplate"); err != nil {
+		return err
+	}
 
 	if err := r.Delete(ctx, sandboxTemplate); err != nil {
```
ranxi2001: **[Medium] Bind the ownership check to the object being deleted**

Both delete helpers first GET the child and confirm `metav1.IsControlledBy`, but `r.Delete(ctx, child)` sends no UID or resourceVersion preconditions. Passing the fetched object to `Delete` does not automatically add those fields to `DeleteOptions`, so with another authorized writer the checked object and the deleted object can differ:

```mermaid
sequenceDiagram
    participant WM as CodeInterpreter controller
    participant API as Kubernetes API
    participant Other as Concurrent authorized writer

    WM->>API: GET child "test"
    API-->>WM: A (UID 111, RV 10, owned by this CodeInterpreter)
    Note over WM: IsControlledBy(A) is true
    Other->>API: Delete A
    Other->>API: Create B with the same name (UID 222, not owned)
    alt Current: DELETE has no preconditions
        WM->>API: DELETE "test" by namespace/name
        API-->>WM: Delete the current object B
    else Requested: bind DELETE to observed identity/version
        WM->>API: DELETE "test" with UID 111 and RV 10
        API-->>WM: Reject the stale request, then reconcile re-reads
    end
```

This is a source-proven latent race, not an observed failure: it requires a concurrent authorized writer to replace or mutate the child after the GET. Even so, the final DELETE can violate this PR's ownership invariant.

Could both delete helpers pass the observed UID and resourceVersion through `client.Preconditions` and return a precondition conflict so reconciliation re-reads the child? Please also add focused tests for both `SandboxTemplate` and `SandboxWarmPool` that replace or mutate the object between GET and DELETE.

avinxshKD: good catch, both delete paths now use UID and resourceVersion preconditions, with replacement-race tests for SandboxTemplate and SandboxWarmPool. PTAL. @ranxi2001

avinxshKD:
ptal @ranxi2001

ranxi2001:
Thanks, this addresses my concern. I verified that the ownership and replacement tests fail without the fix and pass on the current head. My review is complete.

/lgtm

avinxshKD:
@RainbowMango @hzxuzhonghu  ptal

PR #453, opened 2026-08-05
docs: align agent-sandbox version with go.mod

FengyuanYin:
**What type of PR is this?**

/kind documentation

Updates the agent-sandbox version in `docs/getting-started.md` from v0.1.1 to v0.4.6 for both installation and cleanup.

AgentCube currently depends on agent-sandbox v0.4.6, while the getting-started guide still installed v0.1.1. This version mismatch could cause sandbox creation to time out, especially when using the warm pool example, even though the sandbox Pods were running.

None.

The version now matches the `sigs.k8s.io/agent-sandbox v0.4.6` dependency declared in `go.mod`.

Yes.

```release-note
Updated the getting-started guide to install agent-sandbox v0.4.6, matching AgentCube's dependency and preventing sandbox creation timeouts caused by incompatible versions.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -28,7 +28,7 @@ AgentCube relies on the [kubernetes-sigs/agent-sandbox](https://github.com/kuber
 
 ```bash
 # Install agent-sandbox CRDs and controller
-AGENT_SANDBOX_VERSION=v0.1.1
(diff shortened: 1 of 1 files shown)


acsoto:
Thanks for contributing but we do not need this for now because #446 is working on upgrading

PR #460, opened 2026-08-21
fix: align session expiration behavior across API and SDK

### LINKED ISSUE #459: Clarify session expiration behavior in API and Python SDK

Clarify and align session expiration behavior across the Router, Workload Manager API, CRDs, and Python SDK.

A client may continue using `x-agentcube-session-id` after the corresponding sandbox has been reclaimed. The Router then returns HTTP 404, but the response does not provide a stable machine-readable error code. The Python SDK also propagates it as a generic HTTP exception without invalidating the stale session ID.

As a result, clients may repeatedly send an expired session ID and remain in a permanent 404 loop.

Session lifetime configuration is also inconsistent:

- `sessionTimeout` controls idle expiration.
- `maxSessionDuration` controls the hard lifetime limit.
- The Python SDK exposes `ttl`, but the current create-session API does not apply it.
- SDK `timeout` controls HTTP request timeout and is unrelated to session lifetime.

The expected behavior should be:

1. Return a stable error such as `SESSION_NOT_FOUND` when the Router cannot resolve a supplied session ID.
2. Keep this distinguishable from a 404 returned by the application inside the sandbox.
3. Let the Python SDK expose a specific session-not-found error and stop treating the stale session ID as valid.
4. Define whether request-level `ttl` is supported. If supported, enforce it under the CRD maximum lifetime policy; otherwise remove or reject it explicitly instead of silently ignoring it.
5. Document that clients must discard an expired session ID and explicitly create a new session if their application wants to continue.

**Why is this needed**:

Session reuse is part of the public AgentCube API, but expiration behavior is not currently defined well enough for official or third-party SDKs.

A consistent error and lifetime contract prevents clients from repeatedly using stale session IDs and makes session behavior predictable.

acsoto:
**What type of PR is this?**

/kind bug

Aligns the session expiration contract across the Router, Workload Manager, and Python SDK:

- returns `SESSION_NOT_FOUND` only when the stored session is missing;
- treats request `ttl` as a maximum lifetime capped by `spec.maxSessionDuration`, independent of the idle `spec.sessionTimeout`;
- raises `SessionNotFoundError` and clears stale session IDs without implicit recreation or retry.

Application-level 404 responses remain ordinary HTTP errors.

Fixes #459

`ttl` is expressed in seconds. Omitting it preserves the workload CR/default lifetime.

Validated with Go unit tests and vet for the changed packages, 26 Python SDK tests, Ruff, and diff checks. Prepared with AI assistance and manually reviewed and validated.

```release-note
Session expiration now returns a structured error, Python SDKs invalidate stale session IDs, and request-level TTL is enforced within workload limits.
```


acsoto:
cc @RainbowMango

acsoto:
Local E2E verification completed on the current PR head:

- `E2E_CLEAN_CLUSTER=false make e2e`: passed with mTLS enabled.
- `E2E_CLEAN_CLUSTER=false MTLS_ENABLED=false make e2e`: passed with the direct WorkloadManager paths enabled, including CodeInterpreter, Python SDK, LangChain, local/stdio/in-cluster MCP, and load tests (100/100 and 20/20 requests successful).
- Targeted SDK TTL check against the Kind cluster:
  - default `ttl=None` omitted `ttl` from the create request; the Sandbox lifetime was `28800s`, matching `spec.maxSessionDuration: 8h`;
  - explicit `ttl=600` produced a Sandbox lifetime of exactly `600s`.

The targeted sessions were deleted after verification.

RainbowMango (approved):
/lgtm
/approve

